PR #449, opened 2015-05-20
Dev dockerfile

benjamin-guillet:


ejholmes:
This is pretty old. Gonna close for now.

PR #514, opened 2015-06-16
Bumpversion

benjamin-guillet:
This sets the version in the CLI `empire --version`, our logs (version label), tags our docker image and push it to the registry.

It uses [bumpversion](https://github.com/peritus/bumpversion) to increment the versions in `cmd/empire/version.go`, the Makefile and the Dockerfile, as well as tag and commit the changes. The default is `patch`, but you can specify it via the `TYPE` variable:

```console
make release TYPE=patch # Results in 0.8.0 -> 0.8.1
make release TYPE=minor # Results in 0.8.0 -> 0.9.0
make release TYPE=major # Results in 0.8.0 -> 1.0.0
```

### DIFF
--- a/empire/.bumpversion.cfg
+++ b/empire/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version=0.8.0
+current_version = 0.9.0
 commit = True
 tag = True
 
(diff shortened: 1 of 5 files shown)


ejholmes:
Boom :+1:

PR #526, opened 2015-06-17
Add notes about deleting an Empire stack

bmarini:


ejholmes:
:+1:

benjamin-guillet:
Nice, I feel like I should put https://github.com/remind101/empire/wiki/Disable-TLS-on-boot2docker in there

bmarini:
Good idea @benjamin-guillet

docs/troubleshooting.md
```diff
@@ -32,3 +32,11 @@ you need to disable TLS on boot2docker. Here is how to do so:
   ```console
   eval "$(boot2docker shellinit)"
   ```
+
+### Deleting an Empire CloudFormation stack
+
+If you've created an Empire CloudFormation stack and deployed an app to it, you have created an ECS Service with an attached ELB inside the VPC of your Empire stack. Before you can delete the stack, you must no longer have any services or ELBs running inside of it. You can do this by running `emp destroy <app>` for each app in your Empire cluster.
+
+Alternatively, you can [manually delete them][deletingaservice] in the Amazon console. After that, make sure you've deleted the ELB associated with the service as well. Once all dependent ECS Services and ELBs have removed from the VPC of Empire stack, you should be able to delete the stack from the CloudFormation console.
```
phobologic: Does this cause any issues if you delete a service that was maintained by CloudFormation and not by Empire?

phobologic:
Ok, I verified that the only way to ensure that this doesn't happen is to manually delete *all* services in the cluster, including the Empire services.  If you do that, then the CF stack terminates cleanly - and really fast actually.

bmarini:
@phobologic did you also have to delete associated ELBs?

phobologic:
So if they do an `emp destroy` on all their apps, they shouldn't have to do this.  Honestly the only service that you need to delete manually, provided you've destroyed all your apps, is the Empire one, because CF doesn't manage that correctly.

ejholmes:
:+1:

docs/troubleshooting.md
```diff
@@ -32,3 +32,11 @@ you need to disable TLS on boot2docker. Here is how to do so:
   ```console
   eval "$(boot2docker shellinit)"
   ```
+
+### Deleting an Empire CloudFormation stack
+
+If you've created an Empire CloudFormation stack and deployed an app to it, you have created an ECS Service with an attached ELB inside the VPC of your Empire stack. Before you can delete the stack, you must no longer have any services or ELBs running inside of it. You can do this by running `emp destroy <app>` for each app in your Empire cluster.
+
+Additionally, you must [manually delete][deletingaservice] the Empire ECS Service in the Amazon console. After that, make sure you've deleted the ELB associated with the service as well. Once all dependent ECS Services and ELBs have removed from the VPC of Empire stack, you should be able to delete the stack from the CloudFormation console.
```
phobologic: Probably worth mentioning that this is required because of some weirdness with CloudFormation & ECS Services right now, and that we hope this won't be required later.

phobologic:
:+1:

Issue #525, opened 2015-06-17
Support any private registry


desmondmorris:
Currently there is the ability to auth a Docker Hub account.  it would be nice to do the same for any registry.

ejholmes:
Hey @desmondmorris. We haven't tried against another registry like Quay in a while, but I don't think we're doing anything right now that should prevent this. You should be able to deploy from a private registry with:

```
$ emp deploy quay.io/<org>/<repo>:<tag>
```

We should add some tests and update the docs to make this support more explicit.

desmondmorris:
@ejholmes gotcha thanks!

I have already started a patch to add this support.  Basically, it just adds an additional parameter `DockerRegistry` that defaults to the docker hub uri.  I will open a PR.

phobologic:
This was fixed by #528 right?

ejholmes:
Might be some other areas that don't work well with registries other than the official one. Might be good to keep this open until we explicitly support it with tests/docs.

bronislav:
I'm trying to deploy public image from quay.io and it seems that empire looking for it in the wrong place. Below is excerpts from `dockerd` log:
```console
Jul 28 01:53:43 ip-10-5-4-12.empire dockerd: time="2015-07-27T22:53:43.129463639Z" level=error msg="HTTP Error" err="No such image: quay.io/bronislav/clsi-sharelatex:latest" statusCode=404
Jul 28 01:54:43 ip-10-5-4-12.empire dockerd: time="2015-07-27T22:54:43.894958097Z" level=info msg="POST /images/create?fromImage=bronislav%2Fclsi-sharelatex&registry=quay.io&tag=latest"
Jul 28 01:54:43 ip-10-5-4-12.empire dockerd: time="2015-07-27T22:54:43.926265117Z" level=error msg="Error from V2 registry: token auth attempt for registry https://registry-1.docker.io/v2/: https://auth.docker.io/token?account=bronislav%2Bempire&scope=repository%3Abronislav%2Fclsi-sharelatex%3Apull&service=registry.docker.io request failed with status: 401 Unauthorized"
Jul 28 01:54:44 ip-10-5-4-12.empire dockerd: time="2015-07-27T22:54:44.165942820Z" level=info msg="GET /images/quay.io/bronislav/clsi-sharelatex:latest/json"
[...]

PR #528, opened 2015-06-17
Adds ability to add any private docker registry on launch

desmondmorris:
* Adds new input variable to cli to capture docker registry url
* Updates cloudformation template replacing hardcoded docker registry url with the new parameter

Related to: #525


phobologic:
Awesome - thanks for this @desmondmorris - and thanks for being the first non-remind person to submit a pull request :) :+1:

PR #529, opened 2015-06-17
Add dependencies so stack deletes properly

phobologic:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-dependson.html#gatewayattachment

We don't see this in other places because of work stacker does for us.
Also I threw some extra dependencies on the ECS stuff, just to make sure
things delete in the proper order.


ejholmes:
:+1:

Issue #534, opened 2015-06-17
Issue deploying application after running bootstrap script


zachlatta:
I went ahead and followed the instructions in https://github.com/remind101/empire/blob/master/docs/installing.md and everything worked flawlessly. Then I tried to deploy the sample app, as shown below, and got the below error.

```
[zrl@hermes empire]$ emp deploy remind101/acme-inc:latest
Pulling repository remind101/acme-inc
345c7524bc96: Download complete 
a1dd7097a8e8: Download complete 
23debee88b99: Download complete 
31862d352883: Download complete 
c7388ff7ab91: Download complete 
78fb106ed050: Download complete 
133fcef559c4: Download complete 
Status: Downloaded newer image for remind101/acme-inc:latest
error: InvalidParameterException: Unable to assume role and validate the listeners configured on your load balancer.  Please verify the role being passed has the proper permissions.
	status code: 400, request id: []
```

ejholmes:
hey @zachlatta. Woops. Looks like we're not setting the ECS_SERVICE_ROLE environment variable in the task definition created in the demo cloudformation. You should be able to fix this easily in the AWS console:

1. Go to the ECS cluster created by cloudformation
2. Find the empire ECS service
3. Click on the task definition it's currently using.
4. Create a new version of the task definition and add the `ECS_SERVICE_ROLE` environment variable to the `empire` container. The value should be the value of the `ServiceRole` output from the cloudformation stack.
5. Update the empire ECS service to use the new task definition and wait for the new containers to register with the ELB.

Let me know if that helps! I'll update the cloudformation template to include this.

ejholmes:
Made a typo in the above. Environment variable is `EMPIRE_ECS_SERVICE_ROLE`.

phobologic:
I think this should fix the problem: #536

phobologic:
Let us know if this fixes your issue @zachlatta. Thanks again for the issue!

zachlatta:
Things seem to be working! Thanks so much.

phobologic:
Awesome!

PR #536, opened 2015-06-18
Use created serviceRole rather than default

phobologic:
Fixes GH-534

This was being caused when someone launched the Empire cloudformation in
their account with never having used ECS before.  Since we were using
the default ServiceRole (ecsServiceRole) it would fail on those
accounts.

Now we use the ServiceRole that is created with the CF template.

Verified by deleting my account default ecsServiceRole and InstanceRole,
launched a Empire cluster, then made sure I could do things like create
load balancers, etc.


ejholmes:
Works for me too. :+1:

PR #539, opened 2015-06-18
Add Github quiz to bootstrap

ferrix:
This will ask for Github application credentials. The only thing
special about it is that the environment variables have a `Condition`
so that they will not be defined even if they are empty.

### DIFF
--- a/bin/bootstrap
+++ b/bin/bootstrap
@@ -110,6 +110,7 @@ then
   echo "$ export EMPIRE_API_URL=http://$(output ELBDNSName)/"
   echo "$ emp login"
 else
+  echo "==> Your load balancer is at http://$(output ELBDNSName)/"
   echo "==> Dumping stack info into .env for devmode."
   echo "AWS_REGION=$AWS_DEFAULT_REGION" > .env
   echo "AWS_ACCESS_KEY_ID=$(output AccessKeyId)" >> .env
--- a/bin/bootstrap
+++ b/bin/bootstrap
@@ -46,8 +46,10 @@ SAFEZONES=$(echo $ZONES | sed -e 's/ /\\,/g')
 
 read -p "AWS SSH KeyName: " keyname
 read -p "Do you have a docker account & want to use it for private repo access? [y/N] " private_docker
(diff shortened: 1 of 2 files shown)


bin/bootstrap
```diff
@@ -46,8 +46,10 @@ SAFEZONES=$(echo $ZONES | sed -e 's/ /\\,/g')
 
 read -p "AWS SSH KeyName: " keyname
 read -p "Do you have a docker account & want to use it for private repo access? [y/N] " private_docker
+read -p "Do you want to use github for authentication [y/N] " github_auth
```
bmarini: Minor,but missing a `?`

phobologic:
So, before I think we'd merge this I really think we'd need to include some work for enabling SSL.  Honestly I worry that we'd be moving past a simple demo/dev env and into something more full featured, and maybe the bootstrap script isn't the right place for that.

ferrix:
How about a big fat warning? Honestly, this puts the passwords in danger, not the system. I could have added a quiz for a cert arn as well, but I think auth and TLS are important even on a test system.

ejholmes:
Yeah, I definitely agree that we should put a warning in `bin/bootstrap` to make it clear that this is _not_ secure and is only meant to try out Empire. We try to make that [clear](https://github.com/remind101/empire/blob/master/docs/installing.md#empire--installing) in the installation guide, but putting a message in `bin/bootstrap` will make it more obvious.

Regarding GitHub auth in `bin/bootstrap`, I agree with mike that this makes it overly complicated for a demo. We'll be working on docs with guides and best practices for setting up and configuring a production grade Empire installation, like isolating Empire from the ECS cluster minions and putting the Empire API behind a VPN.

phobologic:
I'm going to go ahead and close this.  We aren't comfortable asking people for github credentials, since it might give the idea that those credentials would be secure.  Even with a warning it seems too risky.  I'm going to try to put some time in soon to make either our stacker blueprints available for a more production friendly environment, or even make an attempt at a cloudformation template that would do so.

Thanks for the PR though @ferrix - it pointed out that we need to be more explicit with our security warnings.

PR #540, opened 2015-06-18
move references to readthedocs

phobologic:
Also remove ToC on each page, since readthedocs handles that in a much
better way.


bmarini:
LGTM

Issue #537, opened 2015-06-18
Funny behavior when not logged to docker


ferrix:
As the final step to my crusade to install this thing with `bin/bootstrap` I found this thing. I had not logged in to docker locally, so my `~/.dockercfg` was missing. I would get `server1` die on `/root/.dockercfg` being a directory. It actually gets created as a directory on my host home directory. A simple `touch` on the file gives me the same thing and now with something about EOF.

Unless this is something you can fix, document the use of `docker login` before attempting this.

ejholmes:
I think I remember running into this too. `touch`ing `~/.dockercfg` then restarting boot2docker should fix it. I'll see if I can repro again and get the docs cleaned up. Maybe touch `~/.dockercfg` in `./bin/bootstrap` for good measure?

ferrix:
I had to:

    rmdir ~/.dockercfg
    docker login

A simple `touch` would not do. I think `echo '{}' > ~/.dockercfg` could do it because everything needs to be valid JSON these days.

ejholmes:
Documenting the repro for this for anyone that gets into this state:

**To reproduce**

1. Move your ~/.dockercfg somewhere
2. `docker-compose up`
3. You'll see the error message `read /root/.dockercfg: is a directory`
4. `rm -rf ~/.dockercfg` then move your original `.dockercfg` back to `~/.dockerfg`
5. `docker-compose up`
6. You'll see the error message `Cannot start container f9a2cdf5f7fcf6634cc7f8a2d3b46b860df47ad22c95a77b5b982c7e041c49c3: [8] System error: not a directory`

**Fix**

1. Restart boot2docker
2. `docker-compose up` should work now.

I think it would be nice if we just used the `~/.dockercfg` file instead asking for username and password. If they say they want to use a private registry and they don't have a `~/.dockercfg`, do a `docker login`.

PR #542, opened 2015-06-18
Create an empty ~/.dockercfg if one isn't present.

ejholmes:
I originally wanted to change `bin/bootstrap` to be more like [this](https://github.com/remind101/empire/compare/docker-auth) and just use ~/.dockercfg if provided, but the AWS CLI [doesn't really like JSON](https://github.com/aws/aws-cli/issues/1400). Maybe I'm doing something blatantly wrong.

Closes https://github.com/remind101/empire/issues/537

### DIFF
--- a/bin/bootstrap
+++ b/bin/bootstrap
@@ -98,6 +98,10 @@ then
   echo "$ export EMPIRE_API_URL=http://$(output ELBDNSName)/"
   echo "$ emp login"
 else
+  if [ ! -e ~/.dockercfg ]; then
+    echo '{}' > ~/.dockercfg
+  fi
+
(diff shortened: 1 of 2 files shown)


bin/bootstrap
```diff
@@ -98,6 +98,10 @@ then
   echo "$ export EMPIRE_API_URL=http://$(output ELBDNSName)/"
   echo "$ emp login"
 else
+  if [ ! -e ~/.dockercfg ]; then
```
phobologic: I'd put some sort of output here to let people know what is happening.

phobologic:
:+1:

Issue #512, opened 2015-06-13
Restart is potentially dangerous


ejholmes:
Currently, restarting relies on stopping ECS tasks. If you blindly restart an entire application, all of the current ECS tasks will be stopped. Ideally, ECS would provide a method to do a rolling restart of some tasks, but that's not a feature at the moment. A potential work around would be to re-submit the service with a new task definition to trigger a rolling restart of all of the tasks.

bgentry:
Was this resolved by #543?

ejholmes:
Yep. Closing.

PR #543, opened 2015-06-19
Re-submit last release when restarting entire app.

ejholmes:
This is to address https://github.com/remind101/empire/issues/512. With this change, restarting an app will re-submit the release to ECS, triggering a rolling restart of all tasks.

### DIFF
--- a/empire/apps.go
+++ b/empire/apps.go
@@ -253,36 +253,14 @@ func (s *scaler) Scale(ctx context.Context, app *App, t ProcessType, quantity in
 
 // restarter is a small service for restarting an apps processes.
 type restarter struct {
-	manager service.Manager
+	manager  service.Manager
+	releaser *releaser
 }
 
-func (s *restarter) Restart(ctx context.Context, app *App, t ProcessType, id string) error {
-	instances, err := s.manager.Instances(ctx, app.ID)
-	if err != nil {
-		return err
-	}
-
-	var selected []*service.Instance
-
+func (s *restarter) Restart(ctx context.Context, app *App, id string) error {
 	if id != "" {
-		for _, i := range instances {
-			if i.ID == id {
-				selected = []*service.Instance{i}
-			}
-		}
-	} else if t != "" {
-		for _, i := range instances {
-			if i.Process.Type == string(t) {
-				selected = append(selected, i)
-			}
-		}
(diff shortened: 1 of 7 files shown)


empire/server/heroku/processes.go
```diff
@@ -100,15 +100,18 @@ type DeleteProcesses struct {
 
 func (h *DeleteProcesses) ServeHTTPContext(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
 	vars := httpx.Vars(ctx)
-	ptype := empire.ProcessType(vars["ptype"])
 	pid := vars["pid"]
 
+	if vars["ptype"] != "" {
+		return errNotImplemented("Restarting a process type is currently not implemented.")
```
ejholmes: So, technically we could support this but just submitting a single process instead of all of the processes (just updating a single ECS service).

empire/apps.go
```diff
@@ -253,36 +253,14 @@ func (s *scaler) Scale(ctx context.Context, app *App, t ProcessType, quantity in
...
-				selected = append(selected, i)
-			}
-		}
-	}
-
-	for _, i := range selected {
-		if err := s.manager.Stop(ctx, i.ID); err != nil {
-			return err
-		}
+		return s.manager.Stop(ctx, id)
```
phobologic: So why do we still issue a stop?  Won't this cause the same issues we're seeing right now?  I'm sure I'm confusing something here :)

phobologic:
:+1: looks good, glad to have this

empire/apps.go
ejholmes: Already talked to you about this in person, but just commenting here for future reference. `id` here will be an ECS task uuid and will be unique across all tasks. I'm making the assumption that it's ok to stop 1 task since more often than not, you're gonna have more than 1 of that process type running.

ejholmes:
Actually, now that I think about it. I don't really like this implementation haha. This should really live as a `Restart(app, process string)` method on the `Manager` interface. The ECS implementation can simply pull the Task Definition for the service, re-submit it and update the service. That'll also solve restarting specific process types better, and we don't have to worry about eager loading all the associations for a release.

If ECS ever supports a restart, then the implementation can simply be swapped out with that.

PR #544, opened 2015-06-19
No need to disable TLS in boot2docker anymore.

### LINKED ISSUE #505: Investigate why we need to disable SSL for docker-compose
Since empire needs to interact with the docker daemon API, we need to expose all of the `DOCKER_*` environment variables within the docker container. boot2docker by default enables SSL on the api and it was initially easier to just disable this, but this will be a barrier for contributions and we should fix it.

benjamin-guillet:
This fixes #505 . Before this, we would use the docker.NewClient, instead of docker.NewTLSClient,
since no certPath was provided. Also we now provides the path to the `ca.pem`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,7 @@ server:
     - "8080:8080"
   volumes:
     - ~/.dockercfg:/root/.dockercfg
-    - /var/lib/boot2docker/tls:/root/certs
+    - "$DOCKER_CERT_PATH:/root/certs:ro"
   env_file: .env
   user: root
   environment:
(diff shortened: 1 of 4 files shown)


ejholmes:
Nice. This works well for me :+1:.

docker-compose.yml
```diff
@@ -7,11 +7,13 @@ server:
     - "8080:8080"
   volumes:
     - ~/.dockercfg:/root/.dockercfg
+    - /var/lib/boot2docker/tls:/root/certs
```
ejholmes: This seems to also work for me:

```
- "$DOCKER_CERT_PATH:/root/certs:ro"
```

Which means it'll work without the boot2docker dependency.

benjamin-guillet: Nice! Updated

bmarini:
Nice!

Issue #550, opened 2015-06-23
Proposal: Implement `emp run` within Empire


ejholmes:
Right now, there's a couple of disadvantages with the separate [relay](https://github.com/remind101/empire/tree/master/relay):

1. Since it's separate, it's another dependency you have to run alongside Empire, and you need to configure Empire to know where to find the relay.
2. The relay is fairly complicated to setup, since it requires listening on both HTTP and TCP with TLS.
3. It's difficult to horizontally scale it in it's current implementation (you can only have 1 process running atm).

Now that we've separated out the `emp` CLI, I think it might be worthwhile to break Heroku API compatibility here with something that more closely matches the [attach](https://docs.docker.com/reference/api/docker_remote_api_v1.19/#attach-to-a-container) endpoint of the Docker daemon API. The primary advantages would be:

1. Nothing extra to run. Spin up Empire and `emp run`'s will just work, keeping with our goal of making Empire simple to run.
2. It has no impact on horizontally scaling Empire, since you'll simply attach to the stdout of the container from the HTTP request instead of having to re-connect to a TCP endpoint.

I think the only real disadvantage here is that we'll break `hk run`, but I think it would be worth it.

/cc @phobologic @bmarini

phobologic:
:+1: I'm all for simplifying at the cost of breaking heroku compatibility.

zachlatta:
:+1: agreed with @phobologic

bmarini:
:+1: This will be a much better experience out of the box too.

ejholmes:
Some initial work on this in https://github.com/remind101/empire/pull/553

pikeas:
Might as well keep the current `hk run` around as...`hk-run`?

ejholmes:
Hey @pikeas. Would you want to keep the `hk-run` to use `emp` with Heroku? We're going to be removing the Heroku compatibility for running attached processes within Empire since it can be handled much better by hijacking the connection instead of requiring a separate TCP service.

pikeas:
Not a huge thing for me personally. It doesn't seem like anyone using Empire needs Heroku compatibility, so it's probably fine to diverge as needed.

zachlatta:
Agreed on diverging from Heroku compatibility :thumbsup:.

rgabo:
[...]

PR #552, opened 2015-06-25
Refactor out Image type

ejholmes:
This is some work related to https://github.com/remind101/empire/issues/550 and eventually should help with https://github.com/remind101/empire/issues/525. The primary change is refactoring out the `Image` type into a package that can encode and decode the docker image format (`NAME[:TAG|@DIGEST]`).

I was originally hoping that docker/docker would contain something re-usable for this, but the parsing logic is contained in 3-4 different packages and there doesn't seem to be a single package related to encoding and decoding this format, nor any structure that represents the parsed image.

1. There's [parsers.ParseRepositoryTag](https://github.com/docker/docker/blob/50a1d0f0ef83a9ed55ea2caaa79539ec835877a3/pkg/parsers/parsers.go#L71-L89) to split the remote and tag.
2. [utils.ImageReference](https://github.com/docker/docker/blob/50a1d0f0ef83a9ed55ea2caaa79539ec835877a3/utils/utils.go#L273-L282) to combine a remote and tag/digest into a string.
3. [registry.ParseRepositoryInfo](https://github.com/docker/docker/blob/206acbcf89d2f499b4c8423fb121516d88226a24/registry/config.go#L348-L352) for splitting a remote into a `registry.RepositoryInfo` struct.

Importing any of the above imports a ton of extra packages.

package `image` currently doesn't support private registries or digests. Thinking I'll follow up later by adding that support. **edit** it does now. Boom.

I also refactored out the authenticated pulls logic that the resolver was using, since it will be re-used in the runner.


empire/pkg/image/image_test.go
```diff
@@ -0,0 +1,39 @@
...
+var images = []struct {
+	s     string
+	image Image
+}{
+	{"ubuntu:14.04", Image{Repository: "ubuntu", Tag: "14.04"}},
+	{"remind101/acme-inc:latest", Image{Repository: "remind101/acme-inc", Tag: "latest"}},
+	{"remind101/acme-inc:foo", Image{Repository: "remind101/acme-inc", Tag: "foo"}},
+
+	// TODO
+	//{"quay.io/remind101/acme-inc:foo", Image{Registry: "quay.io", Repository: "remind101/acme-inc", Tag: "foo"}},
```
ejholmes: I'll eventually extend this with support for private registries and digests.

PR #553, opened 2015-06-26
Runner

ejholmes:
This is WIP for https://github.com/remind101/empire/issues/550. Opening it up now for a round of feedback.

There are lot of little details that I want to refactor and cleanup, but for the most part, this actually works pretty nicely:

**Non-Interactive**
https://asciinema.org/a/d38flqwpw36aqxep4ltollc8s

**Interactive**
https://asciinema.org/a/5amhjeznqg1fxhl85hliuf6of

* [x] Refactor to set `EMPIRE_*` environment variables inside the run.
* [x] Return error if `-d` flag is provided (we should consider these as potentially long running and hand them off to ECS to run. Eventually, we should add a Runner implementation for detached runs).
* [x] Handle signals sent by the client.
* [x] Fix occasional bug where stdout doesn't appear.
* [x] Replace "\n" with "\r\n" in pull output.


empire/empire.go
```diff
@@ -315,15 +314,9 @@ func (e *Empire) ProcessesRestart(ctx context.Context, app *App, id string) erro
...
-type ProcessesRunOpts struct {
-	Attach bool
-	Env    map[string]string
-	Size   string
-}
-
 // ProcessesRun runs a one-off process for a given App and command.
-func (e *Empire) ProcessesRun(ctx context.Context, app *App, command string, opts ProcessesRunOpts) (*ContainerRelay, error) {
-	return e.runner.Run(ctx, app, command, opts)
+func (e *Empire) ProcessesRun(ctx context.Context, app *App, command string, in io.Reader, out io.Writer) error {
```
ejholmes: Should probably move all of this into an options struct.

empire/empire.go
```diff
@@ -416,19 +409,23 @@ func newCertManager(config *aws.Config) sslcert.Manager {
...
 }
 
-func newRunner(options RunnerOptions, s *store) *runner {
-	var r containerRelayer
-	if options.API == "fake" {
-		log.Println("warn: runner not configured, command runner disabled.")
-		r = &fakeRelayer{}
-	} else {
-		r = &relayer{API: options.API}
+func newRunner(o DockerOptions, s *store) (*runnerService, error) {
```
ejholmes: We're initializing a lot of docker clients now. It could be better to initialize one and pass it down here.

empire/pkg/runner/runner.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	c, err := r.createContainer(opts)
+	if err != nil {
+		return fmt.Errorf("runner: create container: %v", err)
+	}
+	defer r.removeContainer(c.ID)
+
+	if err := r.startContainer(c.ID); err != nil {
+		return fmt.Errorf("runner: start containeer: %v", err)
```
ejholmes: If the command you provide doesn't exist, you get an ugly error message. I think a new version of the go-dockerclient fixes this. I'll save that update for later though.

empire/server/heroku/processes.go
```diff
@@ -74,24 +76,20 @@ func (h *PostProcess) ServeHTTPContext(ctx context.Context, w http.ResponseWrite
...
-		Type:      relay.Type,
-		Size:      relay.Size,
-		CreatedAt: relay.CreatedAt,
+	inStream, outStream, err := hijackServer(w)
+	if err != nil {
+		return err
+	}
+	defer closeStreams(inStream, outStream)
+
+	fmt.Fprintf(outStream, "HTTP/1.1 200 OK\r\nContent-Type: application/vnd.docker.raw-stream\r\n\r\n")
```
ejholmes: Remove `docker` here.

benjamin-guillet:
Bottom!

empire/pkg/runner/runner.go
```diff
@@ -0,0 +1,152 @@
...
+
+	return nil
+}
+
+func (r *Runner) pull(img image.Image, out io.Writer) error {
+	return r.client.PullImage(docker.PullImageOptions{
+		Registry:     img.Registry,
+		Repository:   img.Repository,
+		Tag:          img.Tag,
+		OutputStream: out,
```
ejholmes: We probably shouldn't actually be writing this to the connection. This could prevent usage of `emp run` for things like this:

```console
$ emp run cat file -a app > file
```

It would be ideal if we could separate stdout and stderr and write this to stderr.

ejholmes: Maybe use the [stdcopy](https://github.com/docker/docker/blob/master/pkg/stdcopy/stdcopy.go#L76) package from docker.

empire/runner.go
```diff
@@ -81,48 +83,42 @@ func (r *relayer) Relay(ctx context.Context, f *postContainersForm) (*ContainerR
...
 	if err != nil {
-		return nil, err
+		return err
 	}
 
-	return r.relayer.Relay(ctx, c)
+	opts.Input = in
+	opts.Output = out
+
+	return r.runner.Run(ctx, opts)
```
ejholmes: It might be good to actually put the `Run` method behind the `service.Manager` interface, since ideally anything that needs to be run should go through the manager.

empire/pkg/runner/runner.go
```diff
@@ -0,0 +1,152 @@
...
+
+	if err := r.attach(c.ID, opts.Input, opts.Output); err != nil {
+		return fmt.Errorf("runner: attach: %v", err)
+	}
+
+	if err := r.wait(c.ID); err != nil {
+		return fmt.Errorf("runner: wait: %v", err)
+	}
+
+	if err := r.stop(c.ID); err != nil {
```
ejholmes: Going to need to wrap the io.Reader with a reader implementation that proxies signals to the container.

ejholmes:
Alright. Pretty happy with this. Tests are gonna fail until the CLI changes are merged.

* Signals now get properly passed through via ANSI escape sequences (ctrl+c kills `sleep x` properly).
* Container gets killed when `ctrl+c` or `ctrl+d` is used.
* `Run` was moved to the `service.Manager` interface. Next step in this would be to implement `Run` method for the ECS implementation for detached runs.

empire/pkg/runner/runner.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// replaceNL returns an io.Writer that will replace "\n" with "\r\n" in the
+// stream.
+var replaceNL = func(w io.Writer) io.Writer {
+	o, n := []byte("\n"), []byte("\r\n")
+	return writerFunc(func(p []byte) (int, error) {
+		return w.Write(bytes.Replace(p, o, n, -1))
+	})
+}
```
rgabo: @ejholmes why is this necessary? I would think we're in an LF-only environment (Linux/Mac). Just curious :blush:

empire/pkg/runner/runner.go
```diff
@@ -0,0 +1,178 @@
...
+
+	"code.google.com/p/go-uuid/uuid"
+
+	"github.com/fsouza/go-dockerclient"
+	"github.com/remind101/empire/empire/pkg/dockerutil"
+	"github.com/remind101/empire/empire/pkg/image"
+	"golang.org/x/net/context"
+)
+
+// DefaultStopTimeout is the number of seconds to wait when stopping a
```
phobologic: # of seconds to wait when stopping a container before...?  Killing it?

empire/pkg/runner/runner.go
```diff
@@ -0,0 +1,178 @@
+// package runner provides a simple interface for running docker containers.
```
phobologic: Is this all docker containers?  Only docker containers for specific commands?  I'm sure I'll figure it out from the code, but might be worth giving a little more description here anyway.

phobologic:
Would be good to add some logging whenever someone launches a one-off container.  Would probably be useful for admins of Empire.

empire/pkg/runner/runner.go
ejholmes: The `emp` cli [sets the terminal into raw mode](https://github.com/remind101/emp/pull/9/files#diff-12f3273da1625fbe0b8ae7f4eefe2c3cR143) so "\n" just puts the output on a newline and doesn't reset it to column 0. "\r\n" does the proper thing.

I believe this is controlled by the `-icrnl` stty flag? Could be wrong.

empire/pkg/service/ecs.go
```diff
@@ -287,6 +288,10 @@ func (m *ecsProcessManager) CreateProcess(ctx context.Context, app *App, p *Proc
 	return err
 }
 
+func (m *ecsProcessManager) Run(ctx context.Context, app *App, process *Process, in io.Reader, out io.Writer) error {
```
phobologic: Might be worth a comment here describing why this is for now - might help a couple of people (like it did me, when we discussed it :))

empire/pkg/runner/runner.go
ejholmes: Already talked to you in personal about this so just putting it here to document.

`package runner` just provides an easy interface for running "attached" containers; where you want to run some command in an image with attached stdin, stdout, stderr.

This package just gets consumed as a primitive by `package service` for running attached processes. It would be nice if we can eventually implement this in such a way that even attached containers get run by ECS...

phobologic:
:+1:

empire/pkg/runner/runner.go
```diff
@@ -0,0 +1,178 @@
...
+	// Image is the image to run.
+	Image image.Image
+
+	// Command is the command to run.
+	Command string
+
+	// Environment variables to set.
+	Env map[string]string
+
+	// Streams fo Stdout, Stderr and Stdin.
```
bmarini: *for

bmarini:
Nice :+1:

docs/cloudformation.json
```diff
@@ -239,7 +239,7 @@
           {
             "LoadBalancerPort": "80",
             "InstancePort": "8080",
-            "Protocol": "HTTP"
+            "Protocol": "TCP"
```
ejholmes: One downside with this is that it seems we need to put the ELB in `TCP` mode to be able to use Hijack. If the ELB is in `HTTP` mode, hijacking the connection and attempting to write to it results in a `broken pipe` error.

In prod, this would be `SSL` here. This isn't a big deal for the controllers, since we just lose the proxy headers.

ejholmes:
Closing this and re-opening against master.

PR #554, opened 2015-06-26
GitHub Enterprise support.

### LINKED ISSUE #547: GitHub Enterprise Support
Currently the GitHub auth support is limited to github.com. We are about to start moving to GHE. Supporting GHE would mean we could reduce the attack surface further and would make administering user access simpler.

public:
Closes #547.

Adds EMPIRE_GITHUB_API_URL option. Should be set to something like

https://your-github-enterprise.com/api/v3

### DIFF
--- a/empire/server/authorization/github/github.go
+++ b/empire/server/authorization/github/github.go
@@ -22,6 +22,9 @@ type Authorizer struct {
 	// organization.
 	Organization string
 
+	// The oauth application URL.
+	ApiURL string
+
 	client interface {
 		CreateAuthorization(CreateAuthorizationOpts) (*Authorization, error)
 		GetUser(token string) (*User, error)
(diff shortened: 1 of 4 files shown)


iserko:
:+1:

empire/server/authorization/github/client.go
```diff
@@ -42,6 +42,7 @@ const (
 // authorization.
 type CreateAuthorizationOpts struct {
 	Scopes       []string
+	ApiUrl       string
```
ejholmes: Looks like this doesn't get used.

empire/server/authorization/github/github.go
```diff
@@ -22,6 +22,9 @@ type Authorizer struct {
 	// organization.
 	Organization string
 
+	// The oauth application URL.
+	ApiUrl string
```
ejholmes: Minor: can we capitalize `URL` here? We try to follow [this convention](https://github.com/golang/go/wiki/CodeReviewComments#initialisms).

ejholmes:
Other than those two points, this looks great. Thanks for this!

phobologic:
:+1: Awesome, thanks for this @public - I don't have anything to add other than what @ejholmes already has.

public:
Updated. Thanks for the pointers.

public:
That's more like it. I wonder why this didn't fail locally :-/

Issue #545, opened 2015-06-19
RFC: Documentation Organization


phobologic:
This is meant to open up our discussions about the organization of the current documentation to the community, as well as gather ideas on what kind of documentation would be useful to everyone.  Please let us know:

- does this look good?
- is there anything missing?
- is there a particular site/project that you think has amazing documentation?

Thanks!

---

**Overview**
- Features & Design GOALS
- Known Limitations
- Roadmap
- FAQ

**Getting Started**
- Prerequisites
- Quickstart
- Empire in Production - Best Practices
  - Deployment
  - Routing
  - Logging
  - Security

**Users Guide**
- Deploying an application
- Process Management
- Exposing an app publicly
- SSL Certs

**Hacking on Empire**
- Running the tests
- Running Empire Locally
- Contributing Guidelines
- Reporting Bugs
- Reporting Security Vulnerabilities

**Reference**
- Configuration Options
- COMMAND LINE
- Troubleshooting

public:
Sounds like a great start.

ekosz:
@phobologic I think would would be a great addition is recommendation on how to run your distributed application locally. For example Empire does its service discovery via DNS, a section how how to replicate that locally would be awesome.

phobologic:
@ekosz When you say 'run your application locally' do you mean running Empire without ECS?

Replicating something like the service discovery without the benefit of things like ECS/ELB/Route53 would be pretty hard.  We use a bit of docker-compose internally, but it's super basic, and not meant to replicate a full production environment.

ekosz:
@phobologic Yeah thats what I meant. How do developers develop without being able to run their applications locally? Can you only test integrations of separate services at the staging level?

What I'm thinking about is that you have a web front-end that serves up a list of products. It may be fetching the products from an inventory service, the signed in user information from another service, and finally product recommendations from the recommendation service.

When working on that web frontend, if you're unable to run those services locally do you just rely on heavy use of mocking those API calls?

zachlatta:
Looks good! Any plans on the timeline of having these docs created? We'd love to experiment with moving some (non critical) applications to Empire in production.

ejholmes:
Hey @ekosz I can explain our development environments a little bit.

Because we have a lot of services, and each one of those services has it's own set of dependencies, we do use a lot of mocking. If you were to setup one of our core services with all of it's dependencies being concrete implementations, it would be a real headache.

[...]

PR #567, opened 2015-06-30
Update doc structure

phobologic:
http://empire.readthedocs.org/en/updating_doc_structure/

New structure from #545


phobologic:
I wanted to get this out there so we could hopefully divide & conquer. It's kind of messy, because of all the stuff that is 'TODO'. If you guys think this isn't worth pushing until we have more content, np.

benjamin-guillet:
:+1: fine by me

ejholmes:
Cool! :+1:

PR #569, opened 2015-07-01
Remove old relay.

ejholmes:
This removes the old relay service, since this is now built into Empire.


benjamin-guillet:
:+1:

phobologic:
:+1:

PR #571, opened 2015-07-01
More empire/empire to just empire.

benjamin-guillet:
Now that /relay is gone,  we can more empire/ to the root of the repository.


ejholmes:
:+1:

bmarini:
Hooray

Issue #570, opened 2015-07-01
Add ECS implementation of `Run` method.


ejholmes:
When running detached processes, we should run them with the [RunTask](http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_RunTask.html) API call. We'd basically want to:

1. Create a Task Definition for the run process.
2. RunTask using the created TD.

https://github.com/remind101/empire/blob/f70a5e2f1b10ec8ee7710847c22befa7efdf5bd4/empire/pkg/service/ecs.go#L291-L302

ejholmes:
One thing to consider here is that in the above scenario, the running task would need by visible with `emp ps` since ps only looks at ECS services. If the `Instances` method was changed to look through Task Definitions instead, that might work better. Ideally, ECS would just support tags on tasks so that we could query for tasks with `app=<app>` tag.

crohr:
Would be really nice to have this feature. Right now it's a bit of a pain to not being able to automatically run migration tasks from our CI for our staging platform. Is the `expect` workaround described in https://github.com/remind101/emp/issues/28 the only way to do it for now?

ejholmes:
Hey @crohr.

Unfortunately that workaround is probably the best way to run one off tasks in an automated way right now. We definitely want to support detached runs, but we're a little tight on resources dedicated to Empire so no eta on this.

PR #573, opened 2015-07-01
Implement Run for ECS.

ejholmes:
Closes https://github.com/remind101/empire/issues/570

This implements detached runs using the ECS `RunTask` api. I probably won't merge this until we have a solution for showing one off tasks like this within `emp ps`.


ejholmes:
Going to hold on this until we have a better way of showing one-off's in `emp ps`.

PR #576, opened 2015-07-02
Fix CFN template for ECS Service to depend on policy and role.

aaithal:
Not having the ECS::Service resource depend on Roles and Permissions required
would result in not being able to delete the stack as the Role gets deleted
before the Service and you end up with the error "Service did not stabilize".


ejholmes:
This is great! Thanks!

PR #580, opened 2015-07-09
Faster docker build.

ejholmes:
Installing godep is the most time consuming part of building. Moving it up makes it cacheable and makes `docker-compose build` without the `--no-cache` flag work as expected.


bmarini:
:+1:

PR #581, opened 2015-07-11
Add a button to create an empire stack.

ejholmes:
Adds a clouformation stack installation button to the readme.

We should followup with a script to automatically upload the cloudformation.json to the s3 bucket for the master branch. The parameters could probably also use some better descriptions for this mode to be more friendly.


morenoh149:
got this after finishing the wizard. Didn't read the instructions too carefully so I may have missed something obvious.
screenshot 2015-07-11 11 33 31

ejholmes:
@morenoh149 you'll need to select an ssh key pair from the dropdown. If you don't have one, you'll need to [create](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-key-pairs.html) one first.

phobologic:
:+1:

Issue #566, opened 2015-06-30
Split commands properly


ejholmes:
There's a couple of places where we take a string command and split it on spaces. This isn't necessarily the correct thing to do. For example, this works ok:

```go
strings.Split("sleep 60", " ")
# => ["sleep", "60"]
```

But if the command has quotes, it's not going to be split properly:

```go
strings.Split(`/bin/sh -c "sleep 60"`, " ")
# => ["/bin/sh", "-c", "\"sleep", "60\""]
```

PR #582, opened 2015-07-11
Default to port 80.

ejholmes:
So, the reason I'm thinking this would be a good idea is that it makes deploying containers that don't respect $PORT easier. This would mostly be an optimization for new comers.

For example, deploying the official `nginx` container would be possible after this change, which would be a pretty slick experience imo:

```
$ emp deploy nginx
$ open http://nginx-elb # default "nginx is installed page"
```

If we did merge this, we'd have to check a couple of our own applications. I know we have a few that currently make the assumption to bind to 8080.

### DIFF
--- a/docs/cloudformation.json
+++ b/docs/cloudformation.json
@@ -16,7 +16,7 @@
     },
     "KeyName": {
       "Type": "AWS::EC2::KeyPair::KeyName",
-      "Description": "The name of the key pair to use to allow SSH access."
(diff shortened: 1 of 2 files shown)


ejholmes:
https://github.com/remind101/empire/issues/566 is also a blocker for deploying the official nginx image.

ejholmes:
Someday! Gonna close for now.

PR #585, opened 2015-07-13
Properly split commands.

### LINKED ISSUE #566: Split commands properly

benjamin-guillet:
With this change we now handle well quoted strings in commands. Fix #566 .

Ex:
```go
args, err := shellwords.Parse("/bin/bash -c 'sleep 60'")
// => ["./bin/bash", "-c", "sleep 60"]
```

Same with backticks:
```go
args, err := shellwords.Parse("/bin/bash -c `sleep 60`")
// => ["./bin/bash", "-c", "`sleep 60`"]
```


ejholmes:
Cool! This looks pretty good to me. I'll play with it tomorrow. I'm curious though, why does go-sqlite3 get pulled in? Last time I pulled it in the tests started running like 10x slower.

pkg/runner/runner.go
```diff
@@ -5,11 +5,11 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"strings"
 
 	"code.google.com/p/go-uuid/uuid"
 
 	"github.com/fsouza/go-dockerclient"
+	"github.com/mattn/go-shellwords"
```
ejholmes: Can we import this as `shellwords "github.com/mattn/go-shellwords"`. Without it, goimports will remove it if you haven't `godep restore`'d.

ejholmes:
Turns out there's a related bug in https://github.com/remind101/empire/blob/7b06a33be2c25c32e6a6aa5eeefeb0ccf1b9192d/extractor.go#L55, where we re-join the command without wrapping it in quotes.

I think the following would fix for now, but kinda sucks:

```go
var parts []string
for _, p := range i.Config.Cmd {
        parts = append(parts, fmt.Sprintf("'%s'", p))
}
pm[ProcessType("web")] = Command(strings.Join(parts, " "))
```

Don't need to fix it in this PR though.

ejholmes:
:+1: once the go-sqlite3 is removed if we can.

benjamin-guillet:
@ejholmes I'm not sure why `go-sqlite3` got pulled in, but it seems to be a dependency of `gorm`, and got pulled in when I `godep save`'d.

On CircleCI, it doesn't seem to slow down the tests: https://circleci.com/gh/remind101/empire/1145 (this branch, 2:22), https://circleci.com/gh/remind101/empire/1143 (master, 5:11).

Locally (time godep go test ./... && time godep go vet ./...)
- this branch: 
```console
godep go test ./...  20.61s user 4.37s system 247% cpu 10.112 total
godep go vet ./...  0.64s user 0.15s system 98% cpu 0.797 total
```
- master
```console
godep go test ./...  20.66s user 4.40s system 251% cpu 9.958 total
godep go vet ./...  0.70s user 0.17s system 98% cpu 0.886 total
```

benjamin-guillet:
About https://github.com/remind101/empire/blob/7b06a33be2c25c32e6a6aa5eeefeb0ccf1b9192d/extractor.go#L55, I'm not sure I understand, would'nt `strings.Join(cmd, " ")` do the right thing if cmd is `cmd := []string{"/bin/sh", "-c", "'sleep 60'"}`? I.e properly formatted with the quotes?

Do we lose the quote from Dockerfile `["/bin/sh", "-c", "'sleep 60'"]` to `cmd := []string{"/bin/sh", "-c", "'sleep 60'"}`?

ejholmes:
Yeah, if `cmd := []string{"/bin/sh", "-c", "'sleep 60'"}` then it would do the right thing, but since it's coming from docker, it's actually `cmd := []string{"/bin/sh", "-c", "sleep 60"}`.

This looks good! :shipit:

PR #586, opened 2015-07-14
Upload cloudformation to s3

ejholmes:
This will upload the cloudformation to the s3 bucket when we commit to master. It uses a locked down user that can only upload this file.


phobologic:
:+1:

Issue #587, opened 2015-07-16
One off tasks don't get stopped


ejholmes:
Been noticing that tasks started with `emp run` aren't getting stopped.

ejholmes:
Ah, this makes sense now. Here's a repro:

```
$ emp run bash -a app
```

Then in another window, kill that process. The container keeps running because we [wait](https://github.com/remind101/empire/blob/master/pkg/runner/runner.go#L69) for the container to exit.

I think the immediate solution would be to kill the container on a client disconnect.

PR #588, opened 2015-07-16
Wrap RemoveContainer with tracing.

ejholmes:
I haven't been able to reproduce https://github.com/remind101/empire/issues/587 locally, so my suspicion is that this call is failing. This will give us errors in HB if it is.


bmarini:
:+1:

PR #591, opened 2015-07-17
Paginate releases.

benjamin-guillet:


pkg/headerutil/headerutil.go
```diff
@@ -0,0 +1,21 @@
+package headerutil
+
+import "strings"
+
+type Range map[string]string
```
ejholmes: What do you think about making this into a struct instead?

```go
type Order int

const (
        Desc Order = iota
        Asc
)

type Range struct {
        // If provided, Sort specifies field to sort by.
        Sort *string

        // If provided, limits the results to the provided value.
        Max *int

        // The order that results are returned in. Zero value is Desc order.
        Order Order
}
```

releases.go
```diff
@@ -49,11 +49,14 @@ func (r *Release) BeforeCreate() error {
...
 type ReleasesQuery struct {
-	// If Provided, an app to filter by.
+	// If provided, an app to filter by.
 	App *App
 
 	// If provided, a version to filter by.
 	Version *int
+
+	// If provided, paginate the results.
+	Limit *int
```
ejholmes: If we specify a Range type as a struct, might be nice to just pass that directly here:

```go
// If provided, uses the limit and sorting parameters specified in the range.
Range *headerutil.Range
```

empire.go
```diff
@@ -323,8 +323,8 @@ func (e *Empire) ProcessesRun(ctx context.Context, app *App, opts ProcessRunOpts
 }
 
 // ReleasesFindByApp returns all Releases for a given App.
-func (e *Empire) ReleasesFindByApp(app *App) ([]*Release, error) {
-	return e.store.Releases(ReleasesQuery{App: app})
+func (e *Empire) ReleasesFindByApp(app *App, limit *int) ([]*Release, error) {
```
ejholmes: I wanted to do this previously, but can we just replace this method with a:

```go
ReleasesFirst(q ReleasesQuery) ([]*Release, error)
```

And consume that from the http api.

ejholmes:
Aside from those comments, and the failing tests, this is great!

pkg/headerutil/headerutil.go
```diff
@@ -0,0 +1,46 @@
...
+	Sort *string
+
+	// If provided, limits the results to the provided value.
+	Max *int
+
+	// The order the results are returned in.
+	Order *string
+}
+
+func ParseRange(header string) (*Range, error) {
```
ejholmes: Could we add some unit tests for this. Probably just some table driven tests with a bunch of various inputs.

releases.go
```diff
@@ -68,9 +72,19 @@ func (q ReleasesQuery) Scope(db *gorm.DB) *gorm.DB {
 		scope = append(scope, FieldEquals("version", *version))
 	}
 
+	if rangeHeader := q.Range; rangeHeader != nil {
```
ejholmes: This looks pretty reusable. We could add a factory method that takes a Range and returns a new Scope:

```go
func Range(r *headeruitl.Range) Scope {
        // Set limit
        // Set sort
        // Set offset
}
```

And then use it like:

```go
if q.Range != nil {
        scope = append(scope, Range(q.Range))
}
```

ejholmes:
Amazing. :shipit:!

PR #593, opened 2015-07-17
Change data type of `memory` column to bigint.

ejholmes:
Turns out we can't scale past 1GB! I guess we could store in mb since we don't need that much resolution, but this should do.


bmarini:
Ha whoops :+1:

phobologic:
:+1: hah :)

PR #597, opened 2015-07-20
Remove test/cli/.netrc when tests run. Fix #592.

### LINKED ISSUE #592: Remove test/cli/.netrc when tests run
I've seen tests fail occasionally because this file was malformed from previous runs. It should be removed before tests start so we always start with a clean slate.

benjamin-guillet:


tests/cli/cli_test.go
```diff
@@ -127,6 +127,14 @@ func writeNetrc(token, uri string) (*os.File, error) {
 	return f, nil
 }
 
+func deleteNetrc() error {
```
ejholmes: Now that I'm looking at this, I'm actually kinda confused why a `.netrc` ever gets written to this directory, since we write the `.netrc` to a tempfile: https://github.com/remind101/empire/blob/remove-netrc/tests/cli/cli_test.go#L109-L128

Maybe it's the login tests?

benjamin-guillet: Yup, it's https://github.com/remind101/empire/blob/master/tests/cli/login_test.go#L10 and https://github.com/remind101/empire/blob/master/tests/cli/login_test.go#L50

ejholmes: Ah, that makes sense. I wonder if it would be more straight forward to just set `NETRC_PATH` in NewCmd. Then have the `Authorize` method just write to that file.

ejholmes:
lgtm :+1:

PR #602, opened 2015-07-22
GitHub deployments support.

ejholmes:
WIP

This adds back in support for GitHub deployments, with the option of using [tugboat](https://github.com/remind101/tugboat/pull/37) as a deployment logs and status backend. If you're not using tugboat, but still using GitHub deployments then this will still work well for you. In the future, we can provide other `deployer` implementations that shuttle logs and status updates elsewhere.

Like the [current provider]() within tugboat, it (by default) assumes that you have a matching docker repo with an image tagged with the git commit sha, but this is easy to change with the `EMPIRE_GITHUB_DEPLOYMENTS_IMAGE_TEMPLATE` environment variable. e.g.:

```
quay.io/{{ .Repository.FullName }}:{{ .Deployment.Ref }}
```

This also has some unique advantages over what we're doing right now:

1. It'll show who deployed in `emp releases` (**TODO** just need to add a column to releases)
2. Instead of pointing webhooks at some tugboat instance, you just point it at Empire now. Eventually, it could be nice to have a CLI command to setup the webhook for you.
3. It formats the json message stream so we get pretty output:

Tests are going to fail because I'm leaving out godeps to make it easier to review.

**TODO**

* [x] Docs
* [ ] Add user column to releases. (I'll save this for a followup PR).
* [x] Use httpx.Handler in `server/github`.
* [x] Make webhook perform deployment in a goroutine.
* [ ] Commit godeps


bmarini:
This looks great :+1: 

It would be nice to expose deployment logs and statuses to webhooks or services, similar to github webhooks and services. Tugboat could become a "service" in that scenario.

ejholmes:
@bmarini totally. Having a webhook system would be really nice for extending some things.

PR #605, opened 2015-07-23
Replace events chan with an io.Writer

ejholmes:
I think we're kinda sold into the jsonmessage package (CLI uses it to decode deploy output) so this just let's us remove a lot of code.


bmarini:
Sweet :+1:

PR #607, opened 2015-07-23
Store release creator.

ejholmes:
With this change, the user that created the release will be displayed in the CLI output for `emp releases`:

```console
$ emp releases -a nginx                                                                           v1        Jul 23 11:15  Deploy nginx:latest
v1  ejholmes  Jul 23 11:44  Deploy nginx:latest
```

This is kinda a shitty, mvp version. Really, we should have a `user_id` column on releases with a `users` table.

After https://github.com/remind101/tugboat/pull/38, I'll `godep update` and the deployer will be tracked in tugboat as well.

**TODO**

* [ ] Pass the use down when updating config.


phobologic:
:-1:

PR #608, opened 2015-07-23
Update tugboat client.

ejholmes:
This will send the user that deployed to tugboat. See https://github.com/remind101/tugboat/pull/38


benjamin-guillet:
Awwww yeahh :+1:

Issue #599, opened 2015-07-21
`emp run` timeouts


ejholmes:
If there's no output for 60 seconds (default ELB idle connection timeout), the connection will close and the container will be killed. We should send periodic heartbeats, probably with a null character.

PR #611, opened 2015-07-23
Send a periodic heartbeat to keep the connection alive.

benjamin-guillet:
Fix for #599 

Maybe we want to reduce the frequency to something < 60s? We can do this:

```golang
ticker := time.NewTicker(30 * time.Second)
go func() {
	for {
		select {
		case <-ticker.C:
			fmt.Fprintf(outStream, "\x00")
			runtime.Gosched()
		}
	}
}()
```

Can also put that into its own function.

### DIFF
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -58,7 +58,7 @@ func (r *Runner) Run(ctx context.Context, opts RunOpts) error {
 	defer r.remove(ctx, c.ID)
 
 	if err := r.start(ctx, c.ID); err != nil {
-		return fmt.Errorf("runner: start containeer: %v", err)
+		return fmt.Errorf("runner: start container: %v", err)
 	}
 	defer tryClose(opts.Output)
 
--- a/server/heroku/processes.go
+++ b/server/heroku/processes.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"runtime"
 
 	"github.com/bgentry/heroku-go"
 	"github.com/remind101/empire"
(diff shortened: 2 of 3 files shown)


pkg/runner/runner.go
```diff
@@ -58,7 +58,7 @@ func (r *Runner) Run(ctx context.Context, opts RunOpts) error {
 	defer r.remove(ctx, c.ID)
 
 	if err := r.start(ctx, c.ID); err != nil {
-		return fmt.Errorf("runner: start containeer: %v", err)
```
ejholmes: embarrassing

benjamin-guillet: ahah

server/heroku/processes.go
```diff
@@ -91,6 +92,15 @@ func (h *PostProcess) ServeHTTPContext(ctx context.Context, w http.ResponseWrite
 
 		fmt.Fprintf(outStream, "HTTP/1.1 200 OK\r\nContent-Type: application/vnd.empire.raw-stream\r\n\r\n")
 
+		// We send the null character periodically, to keep the connection alive.
+		// Otherwise the connection would close after the ELB idle connection timeout.
+		go func() {
+			for {
+				fmt.Fprintf(outStream, "\x00")
```
ejholmes: We only really need to do this every 10-30 seconds. I think we'll also want to make sure that this goroutine stops when the request has stopped. We can select on a couple channels like this:

```go
stop := make(chan struct{})
go heartbeat(w)

// Do stuff

close(stop)

func heartbeat(stop chan struct{}, w io.Writer) {
        // Send heartbeats every 10 seconds.
        t := time.NewTicker(10*time.Second)

        go func() {
                for {
                        select {
                        case <-t: // Tick event
                                fmt.Fprintf(outStream, "\x00")
                                continue
                        case <-stop:
                                break
                        }
                        break
                }
        }()
}
```

benjamin-guillet: Yeah, that's what I put on the PR preview (minus the stop which is a good pt). I'll update.

ejholmes: I added a package under pkg/stream/http. Might be nice to throw that heartbeat function in there :).

Maybe it would be better to return a stop channel, instead of passing it in.

```go
stop := streamhttp.Heartbeat(w, 10*time.Second)
defer close(stop)
```

benjamin-guillet: :boom:

ejholmes: Even nicer, just 1 line it:

```go
defer close(streamhttp.Heartbeat(w, 10*time.Second))
```

ejholmes:
Looks great :+1:

PR #612, opened 2015-07-24
Fix scope of ReleasesQuery when the Range header is not present.

benjamin-guillet:
A lot of things depend of ReleasesQuery returning the last release (order by `version desc`) by default. Including `ReleasesFirst` which is a bit weird since it's technically returning the last release.

Anyway, this behavior was broken everytime the Range header was missing (i.e almost always).


ejholmes:
Yeah, `ReleasesFirst` is a little weird. Maybe we should change all of the `*First` methods to `*Find`.

PR #614, opened 2015-07-24
Update fsouza/go-dockerclient to the last version.

### LINKED ISSUE #610: Trying to run an unknown command with "emp run" results in useless error message.
You get something like: `runner: start container: No such container: <sha>`. According to @ejholmes, it's a docker bug, and it's been fixed in `go-dockerclient`. Just need to upgrade.

benjamin-guillet:
This will fix the uninformative error message when trying to
run an unknown command in a docker container. Fix #610 .

Fixed in https://github.com/fsouza/go-dockerclient/commit/39c614e2d1e5b936b46298e873c26504419238e1


phobologic:
:+1:

ejholmes:
We should test this on staging for a bit just to make sure there's nothing wonky, especially with `emp run`. Last time I tried there was a bug in the connection hijacking that made `emp run` stall.

PR #618, opened 2015-07-28
Store registry along with repo to the app configuration

bronislav:
We need to store full repo including registry in order to be able to pull correct image.


bronislav:
I'm closing this issue since further investigation showed that problem in docker itself and not in the empire. docker/docker#15085

PR #619, opened 2015-07-29
Implement the emp info endpoint

benjamin-guillet:
Implement the emp info endpoint. Address https://github.com/remind101/emp/issues/13.

### DIFF
--- a/tests/cli/apps_test.go
+++ b/tests/cli/apps_test.go
@@ -32,11 +32,6 @@ func TestApps(t *testing.T) {
 }
 
 func TestAppInfo(t *testing.T) {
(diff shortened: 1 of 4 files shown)


tests/cli/apps_test.go
```diff
@@ -27,3 +30,21 @@ func TestApps(t *testing.T) {
...
+	}
+
+	run(t, []Command{
+		{
+			"create acme-inc",
+			"Created acme-inc.",
+		},
+		{
+			"info -a acme-inc",
+			regex,
```
ejholmes: You can just use `regex.MustCompile` and throw the regex here.

ejholmes:
Looks great! :+1:

Issue #615, opened 2015-07-25
Can't specify the app when deploying via emp


mhahn:
I'm curious how you all handle DB migrations within empire. The way I was thinking about doing it was having two apps within empire `api` and `api-migrations`.

`api` would have the external domain attached to it and several web processes running. `api-migrations` would have no external domain and no web processes.

for a release that required migrations, i would like to do something like the following:
```
$ emp deploy -a api-migrations mhahn/api
$ emp run -a api-migrations python manage.py migrate --no-input
# now that migrations are complete, deploying is safe
$ emp deploy -a api mhahn/api
```

DeploymentsCreateOpts supports specifying the app, but this isn't exposed via emp cli. Is this intentional? Is there a more preferred way to do what I'm looking for?

ejholmes:
@mhahn we basically treat it like we did when we were on Heroku. Since ECS is going to do a zero downtime deploy (spin up new processes, health check them, tear down the old ones after connections have drained), any release with migrations needs to be backwards compatible with the last release. So, when we're about to merge a PR, we separate out the migrations first and deploy those, run them, then deploy the code that uses them.

ejholmes:
Regarding specifying the app when deploying, yeah, we want to support that. Just needs a new API endpoint (probably `POST /apps/{app}/deploys`) and CLI support.

mhahn:
I think I'm missing something when trying to test these changes.

I've followed the steps listed here: https://github.com/remind101/empire#development.

After making the changes I do the following:

```
# within $GOPATH/src/github.com/remind101/emp
$ godep go build # to build the new cli with my changes

# within $GOPATH/src/github.com/remind101/empire
$ make build
$ docker-compose up
```

i can see that the cli changes are present and the new URL is being hit in the output from docker-compose, but its returning a 404. I tried testing adding some basic URLs within https://github.com/remind101/empire/blob/master/server/heroku/heroku.go#L19 but all of them return 404s. Did I miss some build step or is there somewhere else the URLs should be added?

mhahn:
and by test I mean, test against a local empire dev server. i wrote unit tests and those are passing.

ejholmes:
If you're making changes to empire, you'll need to rebuild the image that docker-compose uses:

```console
$ docker-compose build
$ docker-compose up
```

mhahn:
ah, cool.

[...]

PR #622, opened 2015-07-30
Deploy specific app

mhahn:
Server side fix for https://github.com/remind101/empire/issues/615.

Two things I wanted to call out:
- The change to `AppsFindOrCreateByRepo`. I had talked to Barrett about this being a design choice on your guys end. The change I made makes sense to me, but I just started working with this stuff today. Let me know if there is some implication I'm unaware of by unlinking them.
- `emp deploy -a some-app remind101/acme-inc` will only work if `some-app` exists. This seemed like a much simpler choice than creating the app if it doesn't exist. If we create the app based on "-a" then that subtly changes the meaning of "-a" in relation to all other commands that use it. I'll defer to whatever you guys prefer on this though.

### DIFF
--- a/apps.go
+++ b/apps.go
@@ -28,10 +28,10 @@ var (
 // NamePattern is a regex pattern that app names must conform to.
 var NamePattern = regexp.MustCompile(`^[a-z][a-z0-9-]{2,30}$`)
 
-// NewAppNameFromRepo generates a new name from a Repo
+// AppNameFromRepo generates a name from a Repo
 //
 //	remind101/r101-api => r101-api
-func NewAppNameFromRepo(repo string) string {
+func AppNameFromRepo(repo string) string {
 	p := strings.Split(repo, "/")
 	return p[len(p)-1]
 }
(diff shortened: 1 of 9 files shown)


ejholmes:
This looks great! I'll play around with it a bit today. Couple of initial comments coming.

deployments.go
```diff
@@ -37,7 +37,7 @@ type DeploymentsCreateOpts struct {
 // deployer is an interface that represents something that can perform a
 // deployment.
 type deployer interface {
-	Deploy(context.Context, DeploymentsCreateOpts) (*Release, error)
+	Deploy(context.Context, *DeploymentsCreateOpts) (*Release, error)
```
ejholmes: Any reason to make this a pointer? In general for option structs like this, I think it's better to not use a pointer unless the nil value has a semantic meaning. I think passing in nil right now would cause a panic.

server/heroku/deployments.go
```diff
@@ -22,10 +22,19 @@ type PostDeployForm struct {
...
 func (h *PostDeploys) ServeHTTPContext(ctx context.Context, w http.ResponseWriter, req *http.Request) error {
+	opts, err := getDeploymentsCreateOpts(ctx, w, req)
+	if err != nil {
+		return err
+	}
+	h.Deploy(ctx, opts)
+	return nil
+}
+
+func getDeploymentsCreateOpts(ctx context.Context, w http.ResponseWriter, req *http.Request) (*empire.DeploymentsCreateOpts, error) {
```
ejholmes: Just a Go style thing. Can we make this `newDeploymentsCreateOpts`?

deployments.go
mhahn: Yea, this is just me being not very familiar with go. I wanted to return nil here: https://github.com/remind101/empire/pull/622/files#diff-71bb861dacfb9ab12070c76ed3a8ce95R37, but it makes sense that the Deploy func should be required to take the struct. Let me fix that.

apps.go
```diff
@@ -168,23 +168,13 @@ func (s *appsService) AppsEnsureRepo(app *App, repo string) error {
...
-	// If the app wasn't found, create a new app linked to this repo.
-	if err != gorm.RecordNotFound {
-		return a, nil
-	}
-
-	n := NewAppNameFromRepo(repo)
-
-	a, err = s.store.AppsFirst(AppsQuery{Name: &n})
+	n := AppNameFromRepo(repo)
+	a, err := s.store.AppsFirst(AppsQuery{Name: &n})
```
ejholmes: I think if we make this change, then we're actually not using the `repo` column anywhere and it could just be removed (doesn't have to happen in this PR).

I was originally thinking that the CLI would allow you to link a docker repo to an app, but I think I might like this approach better.

ejholmes:
This is great! Thanks!

PR #623, opened 2015-07-30
Exit LoadBalancers loop when there are no more returned

markpeek:
This was causing an infinite loop polling the AWS API when
there were no ELB's running in the region.


ejholmes:
Nice find! Can we add a test for this, similar to [TestELB_LoadBalancers](https://github.com/remind101/empire/blob/1afd70c8c7759d425804efa698784726f613cd3b/pkg/lb/elb_test.go#L102).

markpeek:
np...pushed a test for it.

ejholmes:
:+1: thanks @markpeek!

PR #625, opened 2015-07-30
Trace empire deployer.

ejholmes:
I'm seeing https://app.honeybadger.io/projects/43058/faults/14165058 occasionally, which isn't very helpful because the error is overridden by [tugboat.ErrFailed](https://github.com/remind101/empire/blob/792c4e7fb74005f85d15fb3253a0f4e940122fbb/server/github/deployer.go#L125). This should trace the original error as well.


benjamin-guillet:
:+1:

phobologic:
:+1:

Issue #626, opened 2015-07-31
Silent failure when IAM permissions missing


pikeas:
Since the current CloudFormation config has overly broad permissions (#579), I configured my Empire controller to use an IAM user role with no access rights as a test. My plan was to run `emp` commands that I expected to fail with PermissionDenied errors, open up the role, and iterate until everything works.

However, Empire is failing silently:

```
$ emp deploy remind101/acme-inc:latest
...
Status: Image is up to date for remind101/acme-inc:latest

$ emp ps -a acme-inc
<blank output>

$ emp scale -a acme-inc web=2
error: ServiceNotFoundException: Service not found.
	status code: 400, request id: []
```

No service or load balancer is created on AWS, but as far as Empire is concerned, the request completed successfully.

Some endpoints did error successfully. I needed to add `ecs:ListServices` to get this far. Why is service creation failing silently?

phobologic:
Are you checking the logs of the empire daemon?  If it isn't loud enough, you should also try running the Empire daemon with EMPIRE_AWS_DEBUG set.

If you'd like to see somewhat tighter policies, you could also checkout the stacker blueprint policies:

https://github.com/remind101/stacker/blob/master/stacker/blueprints/empire/policies.py

pikeas:
Yep, just discovered and enabled `EMPIRE_AWS_DEBUG`. I now have some evidence of the failures:

```
---[ RESPONSE ]--------------------------------------
HTTP/1.1 403 Forbidden
Content-Length: 386
Content-Type: text/xml
Date: Fri, 31 Jul 2015 02:41:25 GMT
X-Amzn-Requestid: a36131b3-372d-11e5-98b0-5363c1742e73
```

This is better, but the failed request IDs aren't displayed on CloudTrail - I'm looking into that to see if I need to add some additional config on the AWS side.

However, the fact that Empire is showing success when AWS is clearly returning 40x failures for the API calls seems like a Big Deal. Can you confirm that you're seeing the same behavior on your end by intentionally restricting Empire on a test cluster?

phobologic:
Hmm - I may not be able to get to it tonight, but how are you defining 'showing success'?  The deploy command successfully pulling down the image you mean and not throwing an error?

pikeas:
@phobologic As far as I know, the deploy command should a) download the image, b) create a task definition, c) create a service from that definition, and d) create a load balancer for the service.

[...]

PR #628, opened 2015-08-03
Expose AWS API errors when deploying

phobologic:
Fixes GH-626 - I'm sure there are more places where this will pop up,
but the deploy end point should be good to go now.

### DIFF
--- a/deployments.go
+++ b/deployments.go
@@ -92,22 +92,16 @@ func (s *deployerService) doDeploy(ctx context.Context, opts DeploymentsCreateOp
 		Description: desc,
 	})
 
-	if err != nil {
-		return nil, err
-	}
-
-	return r, nil
+	return r, err
 }
 
 // Deploy is a thin wrapper around doDeploy to handle errors & output more cleanly
 func (s *deployerService) Deploy(ctx context.Context, opts DeploymentsCreateOpts) (*Release, error) {
-	r, err := s.doDeploy(ctx, opts)
 	var msg jsonmessage.JSONMessage
+
+	r, err := s.doDeploy(ctx, opts)
 	if err != nil {
 		msg = newJSONMessageError(err)
-		if err = json.NewEncoder(opts.Output).Encode(&msg); err != nil {
-			return nil, err
-		}
 	} else {
 		msg = jsonmessage.JSONMessage{Status: fmt.Sprintf("Status: Created new release v%d for %s", r.Version, r.App.Name)}
 	}
--- a/deployments.go
+++ b/deployments.go
@@ -49,8 +49,8 @@ type deployerService struct {
 	*releasesService
 }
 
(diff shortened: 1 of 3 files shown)


deployments.go
```diff
@@ -92,9 +92,27 @@ func (s *deployerService) Deploy(ctx context.Context, opts DeploymentsCreateOpts
 		Description: desc,
 	})
 
-	if err := json.NewEncoder(opts.Output).Encode(&jsonmessage.JSONMessage{
-		Status: fmt.Sprintf("Status: Created new release v%d for %s", r.Version, r.App.Name),
-	}); err != nil {
+	if err != nil {
```
ejholmes: You can just remove this if and:

```go
return r, err
```

server/heroku/deployments.go
```diff
@@ -20,13 +20,17 @@ type PostDeployForm struct {
...
-// Serve implements the Handler interface.
+// ServeHTTPContext implements the Handler interface.
 func (h *PostDeploys) ServeHTTPContext(ctx context.Context, w http.ResponseWriter, req *http.Request) error {
 	opts, err := newDeploymentsCreateOpts(ctx, w, req)
 	if err != nil {
 		return err
 	}
-	h.Deploy(ctx, *opts)
+
+	_, err = h.Deploy(ctx, *opts)
```
ejholmes: We actually don't want to return an error here, since we're streaming the output. Maybe add a comment noting that we're explicitly ignoring the error for that reason.

deployments.go
```diff
@@ -92,9 +92,27 @@ func (s *deployerService) Deploy(ctx context.Context, opts DeploymentsCreateOpts
...
+	return r, nil
+}
+
+// Deploy is a thin wrapper around doDeploy to handle errors & output more cleanly
+func (s *deployerService) Deploy(ctx context.Context, opts DeploymentsCreateOpts) (*Release, error) {
+	r, err := s.doDeploy(ctx, opts)
+	var msg jsonmessage.JSONMessage
+	if err != nil {
+		msg = newJSONMessageError(err)
+		if err = json.NewEncoder(opts.Output).Encode(&msg); err != nil {
```
ejholmes: This would result in the error being encoded into the stream twice, no? Once here, then it falls down to the one below. I think you just want this:

```go
var (
        msg jsonmessage.JSONMessage
        enc = json.NewEncoder(opts.Output)
)

r, err := s.doDeploy(ctx, opts)
if err != nil {
        msg = newJSONMessageError(err)
} else {
        msg = jsonmessage.JSONMessage{Status: fmt.Sprintf("Status: Created new release v%d for %s", r.Ver
}

if err := enc.Encode(&msg); err != nil {
        return r, err
}

return r, err
```

ejholmes:
:+1:

PR #629, opened 2015-08-03
More specific LB error messages

### LINKED ISSUE #627: Better error messaging when a load balancer is deemed unsuitable
I uploaded an ssl-cert yesterday but was missing the intermediate certificate, when I got an error about that I went straight to the load balancer and updated to a previous cert I had uploaded.

Today I was getting:

```
$ emp set foo=bar -a services
error: currently assigned load balancer is not suitable for the given exposure
```

which makes sense after reading through `lbOk` with it failing on this section

```	
if p.SSLCert != lb.SSLCert {
    return false
}
```

instead of each of these sections returning "false", they should be returning a more descriptive error.

phobologic:
Fixes #627

### DIFF
--- a/pkg/service/lb.go
+++ b/pkg/service/lb.go
@@ -106,22 +106,56 @@ func lbTags(app string, process string) map[string]string {
 	}
 }
 
+type loadBalancerExposureError struct {
+	proc *Process
+	lb   *lb.LoadBalancer
+}
+
+func (e *loadBalancerExposureError) Error() string {
+	var lbExposure string
+	if !e.lb.External {
+		lbExposure = "private"
+	} else {
+		lbExposure = "public"
+	}
+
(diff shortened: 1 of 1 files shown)


ejholmes:
:+1:

mhahn:
:+1: thanks!

pkg/service/lb.go
```diff
@@ -106,23 +106,57 @@ func lbTags(app string, process string) map[string]string {
 	}
 }
 
+type loadBalancerExposureError struct {
```
ejholmes: Prefixing these with `loadBalancer` feels a little verbose since we're already in package lb. How about just `ExposureError` (exported so consumers can type assert on it). e.g.

```go
if err, ok := err.(*lb.ExposureError); ok {
        // Stuff
}
```

ejholmes: I take that back, we're not in package lb. `LoadBalancerExposureError` would be fine imo.

bmarini:
:+1:

ejholmes:
:+1:

PR #634, opened 2015-08-05
Add the app id environment variable in the task definition

benjamin-guillet:
So we can use it to create or find unique kinesis streams per app.


ejholmes:
:+1:

PR #635, opened 2015-08-05
The step of manually deleting the ECS service is no longer required.

ejholmes:
This was fixed in https://github.com/remind101/empire/pull/576.

I actually had trouble finding the uninstallation guide, we should probably put this near the installation section.


benjamin-guillet:
:+1:

PR #641, opened 2015-08-13
Add CrossZoneLoadBalancing to ELBs

### LINKED ISSUE #640: App ELBs should have cross-zone load balancing enabled
http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/enable-disable-crosszone-lb.html

ECS doesn't perfectly balance it's containers really, so this feature really needs to be enabled.

phobologic:
This will ensure that traffic is balanced across backend containers
evenly.

http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/how-elb-works.html#request-routing

Fixes #640


bmarini:
:+1:

ejholmes:
Sweet! We should send something out to the google group and advise people to update their existing load balancers.

:+1:

phobologic:
Yeah, I'll let the list know.

PR #645, opened 2015-08-20
Update aws sdk

benjamin-guillet:
This is prep work for using kinesumer which uses the latest aws sdk. This should also take care of the ThrottlingExceptions and automatically retry (introduced in sdk 0.6.8, we were using v0.6.1).


ejholmes:
Nice. :+1:

Issue #647, opened 2015-08-25
Add pagination on displaying tasks (`emp ps`)


benguillet:
If youh have more than a 100 tasks running for one app, you currently get this when trying to list them (`emp ps`):

```console
error: InvalidParameterException: taskIds can have at most 100 items.
	status code: 400, request id: []
```

Should take advantage of https://github.com/remind101/empire/issues/620

PR #648, opened 2015-08-26
Limit DescribeTasks to 100 tasks per call

ejholmes:
Fixes https://github.com/remind101/empire/issues/647

I also removed the manual pagination in favor of the builtin auto pagination via the `*Pages` calls.


benguillet:
:+1: !

phobologic:
:+1:

PR #649, opened 2015-08-26
Return process type in get processes endpoint

bmarini:
This allows me to add a `scale -l` option to the `emp` command that will query dynos and group them by type to show the current process scales.

### DIFF
--- a/tests/api/processes_test.go
+++ b/tests/api/processes_test.go
@@ -6,6 +6,34 @@ import (
 	"github.com/bgentry/heroku-go"
 )
 
+func TestProcessesGet(t *testing.T) {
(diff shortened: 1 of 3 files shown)


tests/api/processes_test.go
```diff
@@ -3,9 +3,37 @@ package api_test
 import (
 	"testing"
 
-	"github.com/bgentry/heroku-go"
+	heroku "github.com/bgentry/heroku-go"
```
bmarini: For some reason my goimports plugin keeps deleting heroko-go otherwise. I'll fix...

ejholmes:
Nice! :+1:

PR #651, opened 2015-08-27
Tail and streams logs from kinesis

benguillet:
This doesn't handle yet:
 - filtering by process type


logs.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/docker/docker/pkg/jsonmessage"
+	streamhttp "github.com/remind101/empire/pkg/stream/http"
+	"github.com/remind101/kinesumer"
+)
+
+func (e *Empire) StreamLogs(a *App, w http.ResponseWriter) error {
```
r101rosie: This should probably just take an io.Writer.

logs.go
```diff
@@ -0,0 +1,33 @@
...
+	_, err = k.Begin()
+	if err != nil {
+		return err
+	}
+	defer k.End()
+
+	for {
+		rec := <-k.Records()
+		msg := jsonmessage.JSONMessage{Status: string(rec.Data())}
+		if err := json.NewEncoder(rw).Encode(&msg); err != nil {
```
ejholmes: I think we should actually just stream raw lines of data as `text/plain`.

logs.go
```diff
@@ -0,0 +1,33 @@
...
+	"encoding/json"
+	"net/http"
+
+	"github.com/docker/docker/pkg/jsonmessage"
+	streamhttp "github.com/remind101/empire/pkg/stream/http"
+	"github.com/remind101/kinesumer"
+)
+
+func (e *Empire) StreamLogs(a *App, w http.ResponseWriter) error {
+	rw := streamhttp.StreamingResponseWriter(w)
```
ejholmes: This wrapping should happen in the http handler.

logs.go
ejholmes: Also, this is the only place where we add methods onto the Empire struct outside of `empire.go`. I think it would make sense to have a `logsService` interface, where the default implementation is basically just a null implementation, since not everyone is going to want to use kinesis.

Maybe something like:

```go
// logs.go

type logsService interface {
        StreamLogs(context.Context, *App, io.Writer) error
}

type nullLogsService struct{}

func (s *nullLogsService) StreamLogs(ctx context.Context, app *App, w io.Writer) error {
        io.WriteString(w, "Logs are disabled")
        return nil
}

type kinesisLogsService struct {
}

func (s *kinesisLogsService) StreamLogs(ctx context.Context, app *App, w io.Writer) error {
        // ....
}
```

```go
// empire.go

func (e *Empire) StreamLogs(ctx context.Context, app *App, w io.Writer) error {
        return e.logsService.StreamLogs(ctx, app, w)
}
```

And then maybe a CLI flag to enable kinesis streams.

ejholmes:
This is looking really awesome. I think it would be good to follow up with updating the demo cloudformation stack to use the [empire ami](https://github.com/remind101/empire_ami) and add logspout with kinesis enabled on that. But, probably not necessary for now.

Might be good to roll this branch out on staging for a bit to play around with it, but otherwise. :+1:

(could use some docs about how to use kinesis log streaming).

server/heroku/logs.go
```diff
@@ -0,0 +1,27 @@
...
+	*empire.Empire
+}
+
+func (h *PostLogs) ServeHTTPContext(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+	a, err := findApp(ctx, h)
+	if err != nil {
+		return err
+	}
+
+	err = h.StreamLogs(a, streamhttp.StreamingResponseWriter(w))
```
ejholmes: Oh, almost forgot, we should wrap this in `streamhttp.Heartbeat` to prevent ELB from closing the connection.

crohr:
@ejholmes I'd love some docs on that!

Issue #578, opened 2015-07-06
CloudFormation backend?


ejholmes:
Now that CloudFormation supports ECS resources, we could use CloudFormation directly as a backend, rather than using the ECS/ELB/Route53 API's directly. This would remove a lot of nitty gritty manual work like changing listeners and healthchecks.

ejholmes:
The more I think about this the better it sounds. I think there's a couple of key advantages here:

1. The biggest advantage is that we can templatize deployments. We can provide a base text/template for a cloudformation stack that takes in a `service.App` as data and creates the ECS services, linked ELB's, etc but this will allow people to add additional resources to each stack that gets created, even custom resources for provisioning third party services (e.g. kinesis streams, honeybadger projects, papertrail/sumo drains, etc) making it far more extensible than the current backend.
2. We'd ultimately end up removing a lot of code since we'd only need to generate a cloudformation template and talk to one API.
3. Adding support for automatically adding attached resources (RDS), would be really easy.

I think the only drawback is the migration from the old backend to the new backend. We could add a `Migration` implementation of `service.Manager` that uses the new cloudformation backend, and ensures the old ECS/ELB's are destroyed.

public:
In practice how does updating a task definition work in CloudFormation? It the docs it says changing the docker image you have deployed requires replacing the entire Task but does that manifest as a new version of a task or as destroying the task entirely and recreating it?

ejholmes:
CloudFormation will create a new task definition with a random name (it doesn't create a new revision of the existing task definition), which would be fine for Empire since we never actually use the revision directly.

ejholmes:
Also, if we do implement a cloudformation backend, we'd continue to maintain the existing backend and eventually deprecate it depending on how the new backend works.

public:
I kind of like this idea since being able to create new RDS instances and stuff when you deploy your app would be pretty handy sometimes. It does feel a little bit like feature creep at this stage though. I guess this is mostly because we'd rather use Terraform for managing these kind of things right now but that might not be a great reason to dislike this idea.

I have some vague idea of the concept of a "provisioning container" which Empire could execute on each deploy in some context with just enough IAM permissions to create or update the rest of the service. This would decouple Empire (the current service) from any particular provisioning scheme and would make it mostly a system that managed access controls. This has the advantage of meaning Empire can run with fewer permissions and would also potentially make the actual provisioning system pluggable. This could also be called an even more extreme example of feature creep I suppose :)

[...]

PR #657, opened 2015-09-17
Scheduler

ejholmes:
This moves what used to be in ./pkg/service to ./scheduler and individual implementations are broken out into subdirectories:

```
+ scheduler
   + ecs
   + kubernetes
   + cloudformation
   + docker
```

The motivation behind this is just to make this package more prominent, since it's the most important interface within Empire. Breaking out the different implementations into separate sub-directories also makes those implementations more isolated and easier to test.

I've included some stub implementations for kubernetes and docker, just as a reminder that someone should implement these someday :).

### DIFF
--- a/apps.go
+++ b/apps.go
@@ -8,7 +8,7 @@ import (
 	"time"
 
 	"github.com/jinzhu/gorm"
-	"github.com/remind101/empire/pkg/service"
+	"github.com/remind101/empire/scheduler"
 	"github.com/remind101/pkg/timex"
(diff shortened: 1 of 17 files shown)


ejholmes:
This is also some prep work for https://github.com/remind101/empire/issues/578

benguillet:
:+1:

bmarini:
:+1:

phobologic:
:+1:

Issue #491, opened 2015-06-09
Enhanced Procfile support? empire.yml?


ejholmes:
We've talked about this a little bit. Opening this up to keep a record of thoughts.

It would be nice to support a Procfile "upgrade" as an `empire.yml` file. This would potentially be formatted more closely to something like `docker-compose.yml` and allow for configuring things like:

1. Whether a dyno should get a load balancer attached.
2. Whether the load balancer should be exposed publicly or privately.
3. Healthcheck configurations.

An example of this format might look like this:

```yml
web:
  command: ./bin/web
  load_balancer:
    expose: external
    checks:
      - protocol: http
        path: /health
api:
  command: ./bin/api
  load_balancer:
    expose: internal
    checks:
      - protocol: tcp
worker:
  command: ./bin/worker
```

This would come with support for attaching a load balancer to any process.

public:
Any plans to support keeping this kind of configuration outside of the Docker image? Having to download many MB of Docker image just to extract a small text file makes the experience feel very inefficient sometimes :(

ejholmes:
There's a couple of reasons we opted for extracting a Procfile:

1. It matches more closely to what Heroku does. Everything we have already has a Procfile defined.
2. It works properly with branches. Say you have a branch `a` that changes the command for the `web` process. If you swap between branch `a` and `b`, you don't have to change any configuration manually. Less room for error.

True, it is a little slower to deploy, but it hasn't been a major pain point for us yet. If we could support both, that would definitely be worth considering. Maybe if no Procfile is defined, we fallback to manual configuration?

public:
Keeping the deployment configuration in sync with the repo sounds plausibly useful. One approach to this might be to copy docker-compose a bit like you described above expect with the addition of an `image` key that defines the image and tag to deploy?

ejholmes:
Interesting idea, that could be a really nice alternative to extracting a Procfile. Right now, the `/deploys` endpoint takes a simple POST payload with the image to deploy:

```json
{
  "image": "remind101/acme-inc:tag"
}
```

We could optionally extend this to allow the posting of something like a `docker-compose.yml` and it could use that to create the appropriate processes. The deployment tooling (tugboat in our case), could handle pulling our an extended Procfile using the GitHub api, which would be a lot faster than pulling a docker image.

Empire is still somewhat constrained to the idea of having a 1:1 relationship between an app and a docker image, so this will touch a lot of things, but sounds like a good idea.

I'll mull this over some more.

[...]

PR #659, opened 2015-09-28
Extended Procfile format

ejholmes:
This is mostly an RFC for the extended Procfile format that has been discussed in https://github.com/remind101/empire/issues/491.

My goal with this is to keep the scope of this small to start with initially. We've determined that the most pressing use for the new format would be to allow for the configuration of health checks, so I'm only defining attributes to specify the command and any health checks.

I'd also like to keep the extended procfile format 1:1 compatible with the docker-compose.yml format. I think that will open up a lot of possibilities to be able to quickly convert between the two.

In the long run, procfile.ProcessDefinition could replace the [scheduler.Process](https://github.com/remind101/empire/blob/master/scheduler/scheduler.go#L53) type, so scheduler.Scheduler would basically just serve as a means of running a Procfile (Where Empire would just extract the Procfile, merge in env variables, instance counts, etc, then build a new Procfile before handing it to the scheduler). This could provide some pretty nice parity with tools like docker-compose (maybe a docker-compose backend for running Empire in development?) and could warrant breaking out scheduler.Scheduler into it's own project in the future.


procfile/README.md
```diff
@@ -0,0 +1,52 @@
...
+The standard Procfile format is specified in https://devcenter.heroku.com/articles/procfile.
+
+### Extended
+
+The extended Procfile format is Empire specific and implements a subset of the attributes defined in the [docker-compose.yml](https://docs.docker.com/compose/yml/) format. The extended Procfile format gives you more control, and allows you to configure things like health checks for the individual processes. An example of an extended Procfile might look like:
+
+```yaml
+web:
+  command: ./bin/web
+  health_checks:
```
ejholmes: Note that I didn't nest this under a `load_balancer` key, which is talked about in https://github.com/remind101/empire/issues/491. I did this mostly to avoid coupling us to the idea of a load balancer, since it's kind of implementation specific (maybe we'll support some network overlay in the future? Who knows). This also makes it possible to add health checks to non-web processes.

procfile/README.md
```diff
@@ -0,0 +1,52 @@
...
+```
+
+The standard Procfile format is specified in https://devcenter.heroku.com/articles/procfile.
+
+### Extended
+
+The extended Procfile format is Empire specific and implements a subset of the attributes defined in the [docker-compose.yml](https://docs.docker.com/compose/yml/) format. The extended Procfile format gives you more control, and allows you to configure things like health checks for the individual processes. An example of an extended Procfile might look like:
+
+```yaml
+web:
```
ejholmes: Initially, Empire will still make the assumption that a **web** process will get an ELB attached to it. Eventually, this should be configurable. To keep `docker-compose.yml` compliance, I was thinking that you could just specify a port to bind inside the container to specify that you want it exposed.

```yaml
web:
  ports:
    - ":8080"
```

Leaving off the host port causes Empire to allocate a port for that process.

ejholmes: On a side note, we should move PORT allocation into the ECS implementation of scheduler.Scheduler, since it's specific to ECS.

procfile/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+Allows you to specify an array of health checks that should be performed against the process.
+
+```yaml
+health_checks:
+  - type: tcp
+  - type: http
+    path: /health
+    interval: 30
+    timeout: 30
```
ejholmes: We should probably also have attributes for healthy/unhealthy thresholds. I'm still trying to define a format here that will map well to ELB/Kubernetes/etc.

procfile/procfile.go
```diff
@@ -0,0 +1,138 @@
...
+			Command:      pd.Command,
+			HealthChecks: healthChecks,
+		})
+	}
+	return
+}
+
+// ProcessDefinition defines parameters about individual processes within the
+// Procfile.
+type ProcessDefinition struct {
```
ejholmes: I was also thinking of defining a `twelvefactor` package, that defines various types like:

```go
// We define an app as a list of processes that share an environment.
type App struct {
        Env map[string]string
        Processes []Process
}

type Process struct {
        Name string
        Command string
        // etc
}
```

then scheduler.Scheduler is a means of running `twelvefactor.App`, which is pretty cool.

procfile/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+Specifies the command that should be run when executing this process 
+
+```yaml
+command: ./bin/web
+```
+
+**Health Checks**
+
+Allows you to specify an array of health checks that should be performed against the process.
```
phobologic: Not sure how we'd handle an array of health checks in AWS - ELBs only have a single healthcheck, even if they have multiple listeners.

phobologic:
:+1:

procfile/README.md
ejholmes: Yeah. I think in the ECS implementation of the scheduler we'll just raise an error if the Procfile defines multiple health checks.

mcansky:
I like the idea but why put that in the Procfile ? The role of a Procfile is simple and clear, personally I would prefer to have health checks defined in a separate file. Few reasons :
- keep the Procfile simple, dedicated to defining processes
- avoid potential conflicts
- not much overhead to have a separate file for health checks definitions

Any specific reason to try to put everything in the Procfile ?

I understand the idea to be able to use the Procfile as an alternate docker-compose file but that's not the goal of Procfile originally.
Having a separate Healthfile (or something else) is tempting for me as it would allow an easy way for monitoring agents to pick it up and deal with it without having to worry about anything else in the file. Just like the Procfile allows currently for processes.

ejholmes:
Closing this out for now. We'll probably touch on this again at some point, but it's lower priority.

PR #660, opened 2015-10-04
Updating README to use docker-machine (instead of boot2docker)

hairyhenderson:
The boot2docker CLI is deprecated - people should be using
docker-machine now :)

Signed-off-by: Dave Henderson <dhenderson@gmail.com>


ejholmes:
Awesome! Thanks!

PR #664, opened 2015-10-09
bootstrap: set pipefail option

billwanjohi:
There are a couple of awscli calls whose results are piped.
When these calls fail, the script should immediately exit,
but without pipefail ignores the error.


ejholmes:
Thanks!

PR #678, opened 2015-10-18
Enhance release description for config updates

### LINKED ISSUE #675: Release log lies after unsetting an environment variable
After doing `emp unset FOO -a acme-inc`, `emp releases -a acme-inc` will tell you `Set FOO configs vars` as the new release log, which should be `s/Set/Unset`. It can be confusing when looking several releases back.

bmarini:
Fixes #675 cc @benguillet

### DIFF
--- a/configs.go
+++ b/configs.go
@@ -4,6 +4,7 @@ import (
 	"database/sql"
 	"database/sql/driver"
 	"fmt"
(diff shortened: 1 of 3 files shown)


benguillet:
Beautiful :+1:

Issue #677, opened 2015-10-16
Update to latest aws-sdk


ejholmes:
[RegisterTaskDefinition](http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_RegisterTaskDefinition.html) supports a lot of docker options now, which opens up a few doors for us. Probably the best thing we can start with is setting `app`, `process`, and `version` labels on the container.

PR #679, opened 2015-10-21
Set com.empire.* labels on containers

ejholmes:
Closes https://github.com/remind101/empire/issues/677

We can now set labels in task definitions. This sets the following labels on containers:

```
com.empire.app.id
com.empire.app.name
com.empire.app.process
com.empire.app.release
```

This allows third party tools to easily introspect a container to see what app it relates to. If, for example, you're using datadog, you can set the [`collect_labels_as_tags`](https://github.com/DataDog/docker-dd-agent/blob/master/conf.d/docker_daemon.yaml#L115) property so that you can filter docker metrics by Empire application.

Docker [recommends that third party tools prefix labels using reverse DNS notation](https://docs.docker.com/userguide/labels-custom-metadata/#label-keys-namespaces). We don't own empire.com, but I think it would be weird to use `com.remind.empire` for people not at Remind. Thoughts?

(first two commits are godeps, [review this instead](https://github.com/remind101/empire/commit/f1425150d5395a5b3940c43e51def0c2f043a86d))


benguillet:
:+1: Yiiss. I think `com.empire.*` is fine

bmarini:
:+1: I also think its fine

ejholmes:
Just an FYI. After this change, a docker daemon providing docker API >= 1.18 is required.

ejholmes:
This also requires the latest version of the ECS agent (>= 1.5): https://github.com/aws/amazon-ecs-agent/releases/tag/v1.5.0

PR #680, opened 2015-10-21
Remove com.* from labels.

ejholmes:
After talking with @phobologic, since we don't own `empire.com` we should just remove the `com.` prefix. `empire.app` should give us enough namespacing and it reads better.


phobologic:
:+1: sorry for the late comments :)

PR #681, opened 2015-11-24
Support multiple environments for Github deployments.

benguillet:
Allow for environment aliases like `prod` for `production` for instance.


ejholmes:
This is great :+1:. There is a small disadvantage in doing this because the environment will look inconsistent in tugboat, but that's not a big deal.

PR #687, opened 2015-12-12
Auth redesign

ejholmes:
I've added Godeps in this, so you probably want to ignore the first commit.

This is a redesign of the authentication and authorization layer to fix some bugs and improve the existing behavior, as well as open it up for the addition of more granular authorization in the (near?) future.

But, primarily, this does the following:

1. The idea of **authentication** and **authorization** have been decoupled from each other. This fixes a potential security vulnerability where if a user leaves a github organization, their empire access tokens would still be valid. Organization membership is always checked now (<strike>however, the implementation should probably cache in memory for some reasonable amount of time for perf reasons</strike> this now gets cached for 30 minutes in memory, if the authorization check is positive).
2. You can now use an existing auth token to obtain a new auth token.
3. You can now use github credentials and otp for normal API requests (although, you're still better off using those creds to obtain an access token).
4. A nice side effect of this is that the entire API can now be wrapped in authorization instead of each individual endpoint. This makes it less likely that we add an endpoint and forget to wrap it with authentication.
5. The flaky test failures on `server/authorization/github/client_test.go` have been fixed.

Configuration is also logged on boot now:

```
server_1   | 2015/12/14 03:45:16 Using GitHub authentication backend with the following configuration:
server_1   | 2015/12/14 03:45:16   ClientID: e9a709caf6124bb46c23
server_1   | 2015/12/14 03:45:16   ClientSecret: ****
server_1   | 2015/12/14 03:45:16   Scopes: [repo_deployment read:org]
server_1   | 2015/12/14 03:45:16   GitHubAPI:
server_1   | 2015/12/14 03:45:16 Adding GitHub Organization authorizer with the following configuration:
server_1   | 2015/12/14 03:45:16   Organization: remind101
```

**TODO**

* [x] Add an in memory caching implementation of the Authorizer interface.
* [ ] Add an Authenticator that can accept github access tokens. This would be really nice to support for automation: just create a new github personal access token and supply it as the password in basic auth.
* [ ] Add an Authorizer implementation that check's team membership.


phobologic:
Awesome stuff man. :+1:

Issue #653, opened 2015-09-14
Scaling memory size with decimals doesn't work.


ejholmes:
```
$ emp scale process=1:256:1.00gb
error: invalid memory format
```

PR #694, opened 2015-12-13
Allow floating point numbers when scaling process size.

ejholmes:
Fixes https://github.com/remind101/empire/issues/653

This allows you to provide floating point numbers for the memory portion of a process size when scaling:

```console
$ emp scale web=1:256:0.5gb -a acme-inc
Scaled acme-inc to web=1:1X.
$ emp scale web=1:256:0.9gb -a acme-inc
Scaled acme-inc to web=1:256:921.60mb.
$ emp scale web=1:256:0.95gb -a acme-inc
Scaled acme-inc to web=1:256:972.80mb.
```


phobologic:
:+1:

PR #683, opened 2015-11-25
Use CreatedAt/UpdatedAt/StoppedAt from ECS API.

ejholmes:
Closes https://github.com/remind101/empire/issues/682
Fixes https://github.com/remind101/empire/issues/555

```console
$ emp ps -a acme-inc
v1.web.02510d1a-ca9f-4bda-b9d4-d87e4e3a4b7f  1X  RUNNING  26s  "acme-inc server"
$ # kill task
$ emp ps -a acme-inc
v1.web.b46508ce-0aaf-4100-8192-268d6040182e  1X  RUNNING   2s  "acme-inc server"
```

Still need to test this. I'm not seeing the new fields returned with the aws CLI.

[First commit is the actual changes](https://github.com/remind101/empire/commit/ec1810f997090c11d396bf2e4599a5a6c3b1b605), second is just godep updates.

Depends on https://github.com/remind101/kinesumer/pull/20

**Test**

* [ ] Log streaming still works.
* [x] stoppedAt/createdAt/startedAt are never nil.

PR #696, opened 2015-12-13
Package 12factor and new Scheduler interface

ejholmes:
Consider this a proposal for a v2 version of the Scheduler interface. Not necessarily looking to get this merged quickly.

The more I think about it, the more I realize that our Scheduler interface is just a generic interface for running 12factor apps against a scheduling backend, like ECS, Nomad, Kubernetes, etc.

**Empire's sole purpose is to take a manifest, perform some transformations on it (merging in environment variables, release version, image) and then pass it off to a scheduler**. I think there's a lot of benefits to making this fact more explicit in the code base.

**Goals of this PR**

1. Implement a higher quality ECS implementation. Better tests (functional tests against ECS), faster performance (parallelize as much as possible).
2. Define terminology for what we consider a "12factor" application, and provide Go types that mirror that definition (see `package twelvefactor`).
3. Support multiple provisioning backends for the ECS scheduler (**cloudformation** and **raw** api).
4. Potentially produce an isolated binary that can run 12factor applications without all of Empire. (e.g. Here's my Procfile and .env file, run this with the ECS backend)
5. Keep things generic enough so that other 12 factor PaaS's can use it.

This could probably be merged before we actually change Empire to start consuming it.

**TODO**

* [ ] Set logging driver: https://github.com/remind101/empire/pull/731
* [ ] Set correct UpdatedAt in tasks: https://github.com/remind101/empire/pull/683
* [ ] Update SSL cert on load balancer: https://github.com/remind101/empire/pull/683
* [ ] Ulimits https://github.com/remind101/empire/pull/761
* [ ] ELB stuff
* [ ] Don't use `TaskDefinition` to get Task information, pull the tasks and then retrieve the task definition for the task.


12factor/scheduler.go
```diff
@@ -0,0 +1,60 @@
...
+
+// Runner is an interface that wraps the basic Run method, providing a way to
+// run a 12factor application.
+//
+// Implementors should ensure that any existing processes that don't exist in
+// the newly submitted process list are removed. For example, if a "web" process
+// was previously defined, then only a "worker" process was submitted, the
+// existing "web" process should be removed.
+type Runner interface {
+	Run(Manifest) error
```
ejholmes: Just a thought. This might be better named `Up` and we can have an optional `Down` method that scales everything to 0 (e.g. Maintenance mode).

12factor/twelvefactor.go
```diff
@@ -0,0 +1,102 @@
...
+	// The container image for this app.
+	Image string
+
+	// The shared environment variables for the individual processes.
+	Env map[string]string
+}
+
+// Process represents an individual Process of an App, which defines the command
+// to run within the container image.
+type Process struct {
```
ejholmes: One thing missing from this right now is port mappings. My thought is that, if you provide a host/container port mapping, then it's implementation specific as to what that does (e.g. with ECS backend, you'll probably get an ELB). How to keep this generic enough so that it can describe SSL/HTTP/TCP/etc across all implementations is challenging.

ejholmes:
Gonna go ahead and close this. New CloudFormation scheduler was merged in https://github.com/remind101/empire/pull/803

Issue #458, opened 2015-05-27
Post Empire events to an SNS topic.


ejholmes:
This plays off of @bmarini's idea to add webhooks. SNS might be easier because we can build less. It would be sweet to have events for things like ELB's getting created, processes being scaled, etc.

Some cool consumers we could build:

1. Something that listens for ELB creation events and uses [this](http://dev.librato.com/v1/put/sources/:name) librato api to rename the source to something more meaningful.
2. An app that creates a base librato dashboard when new apps are created.

crohr:
Having some kind of webhook support would also be useful to post deployment/scaling notifications to slack/irc/whatever.

mhahn:
I was just thinking of this today as well. It would also be useful for
auditing who has run a container and when

Issue #658, opened 2015-09-18
Inject admin events into logstream


ejholmes:
Now that we have log streaming, it would be nice to also be able to write admin events into the apps log stream for things new deployment events, scaling, etc.

mhahn:
@ejholmes this should be done with https://github.com/remind101/empire/pull/836

mhahn:
Reopening this because it isn't currently working with the new empire daemon blueprints.

We don't currently report any errors from publishing to the stream, so need to add that.

PR #698, opened 2015-12-15
Event stream

ejholmes:
Related https://github.com/remind101/empire/issues/658
Closes https://github.com/remind101/empire/issues/458

This is a little spike into an events system, starting with **run**, **scale**, **deploy**, **restart** and **rollback** events. Note that this _doesn't_ attempt to give you any visibility into ECS events, just Empire triggered events for auditing.

I'm starting with an SNS implementation of the Events interface, which would allow you to write SQS consumers (or just lambda functions) that processes the events in whatever way you want (maybe post to a slack channel, or post to a datadog event stream). Eventually, we may provide other implementations, like webhooks (but that can be done with SNS easily by creating an http subscription).

**TODO**

* [ ] Add an `Events` implementation that writes into the apps log stream https://github.com/remind101/empire/issues/658
* [x] Make publishing events happen in a goroutine.
* [x] Update demo CloudFormation to use SNS events backend.
* [x] Docs on configuring SNS event stream


server/heroku/heroku.go
```diff
@@ -141,3 +143,24 @@ func RangeHeader(r *http.Request) (headerutil.Range, error) {
...
+
+// key used to store context values from within this package.
+type key int
+
+const (
+	userKey key = 0
+)
+
+// WithUser adds a user to the context.Context.
+func WithUser(ctx context.Context, u *empire.User) context.Context {
```
ejholmes: Moving these functions from empire core to here so that they can't be used within empire core. Providing the user that's performing an action should happen in the options struct provided.

cmd/empire/factories.go
```diff
@@ -177,6 +184,27 @@ func newKinesisLogsStreamer(c *cli.Context) (empire.LogsStreamer, error) {
 	return empire.NewKinesisLogsStreamer(), nil
 }
 
+// Events ==============================
+
+func newEventStream(c *cli.Context) (empire.EventStream, error) {
```
phobologic: Would things like this be easier to test, etc, if the functions took the arguments they actually used?  Like in this case it seems kind of wasteful to pass the entire cli.Context in, when all we really care about is a single variable - it also makes it harder to read, since I'm not actually sure, looking at this, what it needs as it hands the cli context further down the stack.

cmd/empire/main.go
```diff
@@ -66,6 +69,12 @@ var Commands = []cli.Command{
 				Usage: "Whether to run the migrations at startup or not",
 			},
 			cli.StringFlag{
+				Name:   FlagEventsBackend,
+				Value:  "",
+				Usage:  "The backend implementation to use to send event notifactions",
```
phobologic: Is there an easy way with codegangsta's CLI to give a list of choices for an option and display that with the help?

docs/cloudformation.json
```diff
@@ -488,10 +489,17 @@
...
               {
                 "Name": "EMPIRE_EC2_SUBNETS_PUBLIC",
                 "Value": { "Fn::Join": [ ",", [{ "Ref": "PubSubnetAz1" }, { "Ref": "PubSubnetAz2" }] ] }
+              },
+              {
+                "Name": "EMPIRE_EVENTS_BACKEND",
+                "Value": "sns"
+              },
+              {
+                "Name": "EMPIRE_SNS_TOPIC",
```
phobologic: There should probably be a condition for this - requiring that folks create an events topic is a bit annoying.

cmd/empire/factories.go
ejholmes: I think that would be worth it if we pull the factories out into a separate package, but that gets cumbersome because we have to define Config structs for everything. Some factories require [more than 1 config](https://github.com/remind101/empire/blob/master/cmd/empire/factories.go#L83-L118) value, so this keeps things easy to change without needing to updating config structs/method definitions.

cmd/empire/main.go
ejholmes: Not out of the box, but I think we could implement our own StringFlag based on [GenericFlag](https://github.com/codegangsta/cli/blob/master/flag.go#L59-L70). I'll take a stab at that.

docs/cloudformation.json
ejholmes: Good call. I'll update.

ejholmes:
Ok. I think this is good to go. I also added an example Lambda function in the docs that can be used to publish Empire events to a slack channel.

phobologic:
:+1:

Issue #662, opened 2015-10-05
Error trying to update SSL certificate


mhahn:
I was trying to update the ssl-cert for an app and got the following:

```
$ emp ssl-cert-add -a services crt.pem key.pem
error: DeleteConflict: Certificate: BLAH is currently in use by arn:aws:elasticloadbalancing:us-west-2:BLAH:loadbalancer/BLAH. Please remove it first before deleting it from IAM.
	status code: 409, request id: [c6e50af2-6b90-11e5-91fc-4df76678f0a4]
```

mhahn:
The error from AWS makes sense since empire is attempting to delete the certificate without disconnecting it from the load balancer.

It seems like empire should be more gracefully detatching, deleting, and then adding the new SSL certificate.

mhahn:
there is no "UpdateLoadBalancer" only "CreateLoadBalancer" which will create a load balancer with an SSL cert if it exists before it creates the process in ECS.

there should be an `UpdateLoadBalancer` call that can be called from ssl-cert-add and friends. i'll work on adding that.

mhahn:
http://docs.aws.amazon.com/ElasticLoadBalancing/latest/APIReference/API_SetLoadBalancerListenerSSLCertificate.html @phobologic pointed out this documentation, which `ssl-add-cert` should be calling if the load balancer exists

PR #700, opened 2015-12-17
Allow SSL certs to be updated on ELB's in ECS backend

ejholmes:
This allows the ECS backend to update the SSL cert on the associated ELB if it changes.

A better approach would be to provision resources using CloudFormation like talked about in https://github.com/remind101/empire/pull/696, but that's a ways off and will require a substantial migration.

### DIFF
--- a/pkg/lb/elb.go
+++ b/pkg/lb/elb.go
@@ -102,6 +102,10 @@ func (m *ELBManager) CreateLoadBalancer(ctx context.Context, o CreateLoadBalance
 	}, nil
 }
 
+func (m *ELBManager) UpdateLoadBalancer(ctx context.Context, opts UpdateLoadBalancerOpts) error {
+	return nil
+}
+
 // DestroyLoadBalancer destroys an ELB.
 func (m *ELBManager) DestroyLoadBalancer(ctx context.Context, lb *LoadBalancer) error {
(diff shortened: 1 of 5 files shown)


pkg/lb/elb.go
```diff
@@ -102,6 +102,25 @@ func (m *ELBManager) CreateLoadBalancer(ctx context.Context, o CreateLoadBalance
...
+		if err := m.updateSSLCert(ctx, opts.Name, *opts.SSLCert); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (m *ELBManager) updateSSLCert(ctx context.Context, name, certID string) error {
+	_, err := m.elb.SetLoadBalancerListenerSSLCertificate(&elb.SetLoadBalancerListenerSSLCertificateInput{
```
ejholmes: This should actually ensure that there's a 443 listener in case there was previously no cert attached.

ejholmes:
Just realized that https://github.com/remind101/empire/issues/662 actually still prevents this. I wonder if it would just be better if the UI for adding an SSL cert was more of an "attachment" flow:

```console
$ aws iam upload-server-certificate --server-certificate-name ProdCert
```

```console
$ emp ssl-attach ProdCert -a <app>
```

phobologic:
I'm all for changing the behavior to something like you're talking about, for all the reasons you mention.

mhahn:
huge +1, this would be amazing

PR #701, opened 2015-12-17
Remove existing SSL management.

### LINKED ISSUE #471: ssl: one certificate per web app may be an issue
Right now, when you add an SSL certificate, it always creates a new one.  Even if, for example, you have a *.<domain> certificate.

An account only has a limit of 20 certificates:

http://docs.aws.amazon.com/IAM/latest/UserGuide/LimitationsOnEntities.html

It might be worth building the certificate name by combining the CN of the cert with the timestamp  and then hashing them.  That way if services share the same cert, we don't have to use up another cert in the IAM system.

### LINKED ISSUE #662: Error trying to update SSL certificate

ejholmes:
Fixes #471 
Fixes #662

This still needs client support (`emp certs-attach`) and associated API endpoint, but this would change how certs are managed via Empire, in that you don't use Empire to manage certs, just attach existing certs to apps.

Note that https://github.com/remind101/empire/pull/700 will still be required to fix the fact that ELB's can't be updated to use new certs.

The basic flow for attaching a cert to an Empire application will be:

```console
$ aws iam upload-server-certificate --server-certificate-name ProdCert
$ emp certs-attach ProdCert -a <app>
```

I think this will also ultimately work better when we add more backends, since we won't have to try to support varying CertManager implementations.

Also note that this will be backwards compatible with existing certificates inside Empire (see migration).

**TODO**

* Sanity check migration


server/heroku/heroku.go
```diff
@@ -57,12 +57,6 @@ func New(e *empire.Empire, authenticator auth.Authenticator) httpx.Handler {
 	// OAuth
 	r.Handle("/oauth/authorizations", &PostAuthorizations{e}).Methods("POST")
 
-	// SSL
```
ejholmes: We should map these to http handlers that explain that they were removed.

phobologic:
This looks good so far - are you planning to put the new SSL attachment stuff in this same PR before merging?

ejholmes:
Yep, I'll be adding the new endpoint shortly with support in `emp`.

certs.go
```diff
@@ -0,0 +1,25 @@
...
+}
+
+func (s *certsService) CertsAttach(ctx context.Context, app *App, cert string) error {
+	app.Cert = cert
+
+	if err := s.store.AppsUpdate(app); err != nil {
+		return err
+	}
+
+	if err := s.releaser.ReleaseApp(ctx, app); err != nil {
```
ejholmes: Instead of re-releasing, this should probably creat a new release.

ejholmes: I'll follow up with this in a separate PR that makes the API simpler for creating new releases for apps.

PR #702, opened 2015-12-18
Add some deploy/config tests around public API.

ejholmes:
This adds some tests around the `Deploy` and `ConfigsApply` methods to make sure that they interact with the scheduler correctly. A little verbose, but they'll be nice for refactoring internals.


phobologic:
Awesome! :+1:

PR #707, opened 2015-12-31
Add create, destroy and set events

ejholmes:
This adds **create**, **destroy** and **set** events to the event stream and includes some work to cleanup the public API of Empire a little more.

### DIFF
--- a/apps.go
+++ b/apps.go
@@ -104,8 +104,8 @@ func (q AppsQuery) Scope(db *gorm.DB) *gorm.DB {
 	return scope.Scope(db)
 }
 
-// AppsFirst returns the first matching release.
-func (s *store) AppsFirst(scope Scope) (*App, error) {
+// AppsFind returns the first matching release.
+func (s *store) AppsFind(scope Scope) (*App, error) {
 	var app App
 	return &app, s.First(scope, &app)
 }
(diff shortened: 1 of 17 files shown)


apps.go
```diff
@@ -104,8 +104,8 @@ func (q AppsQuery) Scope(db *gorm.DB) *gorm.DB {
 	return scope.Scope(db)
 }
 
-// AppsFirst returns the first matching release.
-func (s *store) AppsFirst(scope Scope) (*App, error) {
+// AppsFind returns the first matching release.
+func (s *store) AppsFind(scope Scope) (*App, error) {
```
ejholmes: @benguillet brought up that `*First` was confusing and `*Find` would be a better name, so this is just a cosmetic change.

empire.go
```diff
@@ -169,9 +240,9 @@ func (e *Empire) DomainsDestroy(domain *Domain) error {
 	return e.domains.DomainsDestroy(domain)
 }
 
-// JobStatesByApp returns the JobStates for the given app.
-func (e *Empire) JobStatesByApp(ctx context.Context, app *App) ([]*ProcessState, error) {
```
ejholmes: The terminology we used to describe running tasks was confusing/inconsistent before (jobs, job state, process state), so I'm just called these `Tasks'`s now.

Issue #686, opened 2015-12-09
Add a Slack slash command interface


ejholmes:
I think it would be really cool to have built in support for an `/empire` slack command.

e.g.

```
/empire ps -a app
/empire restart -a app
/empire scale web=2 -a app
```

There's a Go library at https://github.com/ejholmes/slash for handling Slack slash commands.

No plans to implement this anytime soon, but I wanted to open it up to keep track of the idea.

mhahn:
if we setup: https://github.com/remind101/empire/issues/458 i could route those to slack. i'm going to look into setting this up. i get tired of running `emp ps -a app` over and over to determine if the app has been deployed :)

phobologic:
My main concern over this is the loss of authentication.  I think some places might be fine with that.  I think it largely also has to do with what commands we allow.

@mhahn I don't think that #458 would actually fix what you're saying.  The issue there is that Empire, once we make a change, does not actively monitor and report back when that change is complete.  So it'd be larger than what #458 is advocating.  It is annoying though to have so little feedback on whether changes you made worked or not.

phobologic:
Oops, wrong button :)

ejholmes:
So, I think we actually wouldn't lose authentication if we decided to do this. We can still oauth users that come in through the slack commands and do the same ACL stuff that we're planning to do.

Here's an [example](https://github.com/ejholmes/slashdeploy/blob/7b62ffb5f829d664f3bd64b67251e57e15eb3de3/auth/auth.go#L27-L100) of doing that.

phobologic:
Huh, that's pretty cool- though once someone authenticates they are forever authenticated it looks like?  Not sure that matters much.  Are you imagining that slack would have it's own end points for all the commands it would have, or would you use the existing command endpoints?

ejholmes:
Might require a separate table that stores a `slack user id` -> `empire access token` mapping, but I'm pretty sure we can reuse all of the authorization middleware that we use in the api control layer with this.

ejholmes:
I was imaging it would just be a separate control layer in `package server/slack`. It'd just be a consumer of `package empire` like the api is. I think this could be mvp'd with just a couple of commands. One of the most important for us would be an `/empire restart` and a slash command for enabling github deployments webhooks on a repo.

PR #708, opened 2015-12-31
Slack interface

ejholmes:
This is a little spike into a Slack interface to Empire for https://github.com/remind101/empire/issues/686. **Not ready to merge**

The way I built it, there's now a `cli` package that provides a CLI to Empire (whether it's in process or out of process doesn't matter) and there's just a slash handler that serves the CLI to Slack.

**Things still todo**

* [x] Wrap slack server with [slash.ValidateToken](https://godoc.org/github.com/ejholmes/slash#ValidateToken)
* [ ] Authenticate slack users against github. (Currently just runs as the `slack` user).
* [ ] Docs
* [ ] Add ability to disable certain commands.


cli/scale.go
```diff
@@ -0,0 +1,54 @@
...
+	"errors"
+	"strconv"
+	"strings"
+)
+
+var errInvalidScaleArg = errors.New("invalid argument")
+
+// This is a direct copy paste from
+// https://github.com/remind101/emp/blob/6fc34ef0633f5a26217023ede86b866bf4f17cc4/scale.go#L127-L168.
+//
```
sanjayprabhu: I assume it's not worth sharing any code between this and `emp`?

sanjayprabhu:
Nice, this is looking great!

cli/scale.go
ejholmes: That's definitely the end goal. If https://github.com/remind101/empire/pull/712 gets merged then this can be shared.

phobologic:
This just came up internally, but we should probably have a way for people to disable certain commands.  Here, for example, we'd definitely not want to allow `run` over slack, since that goes against some of our security principles.  `scale` is questionably on that spectrum as well.

ejholmes:
Added to the TODO section.

ejholmes:
This probably won't get any love anytime soon, so gonna close for now.

Issue #644, opened 2015-08-20
EOF error when deploying a non-existent tag


ejholmes:
This error is somewhat specific to us. If you have a repository that has a lot of tags in it (e.g. you tag every image with the git sha), when you try to deploy an image using a tag that doesn't exist, you'll see it stall for a minute or two and fail with:

```
$ emp deploy remind101/acme-inc:nonexistenttag
Pulling repository remind101/acme-inc
error: unexpected EOF
```

This happens because the behavior of the docker client is to first try to find the tag in the v2 registry, then fallback to pulling _all_ tags using the v1 registry API. Because there's so many tags, this times out and returns the EOF error.

This technically isn't an Empire bug, but tracking it here in case someone else runs into it.

See also https://github.com/docker/distribution/issues/797

ejholmes:
FYI. I'm not able to reproduce this on Docker 1.8.1.

PR #709, opened 2015-12-31
Expose errors in Tugboat properly

ejholmes:
This "fixes" https://app.honeybadger.io/projects/43058/faults/14165058 in that it makes the error visible. Previously, it would get silenced and people would be confused why they're deployment failed (tag wasn't found in the docker registry. See https://github.com/remind101/empire/issues/644). It's the biggest source of errors that we see from Empire, so it's annoying when you just see a blank error.

The behavior before was to write the error to the log stream, then update the deployment status in Tugboat as failed. But if there was an error writing to the log stream, then the error would just be silenced. It's better if we just update the deployment status in Tugboat passing it the actual error that happened.

Depends on https://github.com/remind101/tugboat/pull/41

**TODO**

* [ ] Merge https://github.com/remind101/tugboat/pull/41

### DIFF
--- a/pkg/dockerutil/dockerutil.go
+++ b/pkg/dockerutil/dockerutil.go
@@ -1,11 +1,13 @@
 package dockerutil
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
 
 	"github.com/docker/docker/pkg/jsonmessage"
+	"github.com/docker/docker/pkg/term"
 	"github.com/remind101/empire/pkg/image"
 )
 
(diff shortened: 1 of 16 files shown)


pkg/dockerutil/dockerutil.go
```diff
@@ -0,0 +1,73 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// DecodeJSONMessageStream wraps an io.Writer to decode a jsonmessage stream into
+// plain text. Bytes written to w represent the decoded plain text stream.
+func DecodeJSONMessageStream(w io.Writer) *DecodedJSONMessageWriter {
```
ejholmes: Also refactored out how the jsonmessage stream is docoded, so you can just wrap an io.Writer to decode the stream and write plain text.

phobologic:
:+1:

PR #713, opened 2016-01-03
Remove store

ejholmes:
This removes the internal `store` type, in favor of passing around a `*gorm.DB` instance. This is better, because it allows us to define better transactional boundaries, so we can wrap methods on the Empire struct in a transaction. This fixes that bug (thought we had an issue for it but couldn't find it) where deploying a docker image that didn't exist would still create an app within Empire, as well as some race conditions during release creation.


configs.go
```diff
@@ -156,18 +151,20 @@ func (s *configsService) Set(ctx context.Context, opts SetOpts) (*Config, error)
...
 			// It's possible to have config without releases, this handles that.
-			c, err := s.store.ConfigsFirst(ConfigsQuery{App: app})
+			c, err := configsFind(db, ConfigsQuery{App: app})
 			if err != nil {
 				if err == gorm.RecordNotFound {
-					return s.store.ConfigsCreate(&Config{
-						App:  app,
-						Vars: make(Vars),
-					})
+					// Return an empty config.
```
ejholmes: This actually exposed a problem with using `created_at` to determine what is the current config for an app. There was a case where, when setting config on an app that didn't have any releases, 2 configs would get created at about the same time; 1 empty config and 1 with the update vars. Returning an empty config here fixes that race, but using an auto incremented id to order would probably be better.

apps.go
```diff
@@ -217,27 +202,28 @@ func (s *scaler) Scale(ctx context.Context, opts ScaleOpts) (*Process, error) {
...
+	if err := processesUpdate(db, p); err != nil {
+		return nil, err
+	}
+
+	// If there are no changes to the process size, we can do a quick scale
+	// up, otherwise, we will resubmit the release to the scheduler.
+	if c == nil {
+		return p, s.Scheduler.Scale(ctx, release.AppID, string(p.Type), uint(quantity))
+	} else {
+		return p, s.releaser.Release(ctx, release)
```
ejholmes: This also fixes the issue where scaling with a new process size (e.g. `emp scale web=2:256:20m`) would require a new deployment to take effect.

releases.go
```diff
@@ -94,42 +102,82 @@ func (q ReleasesQuery) DefaultRange() headerutil.Range {
...
+// releasesService is a service for creating and rolling back a Release.
+type releasesService struct {
+	*Empire
+}
+
+// ReleasesCreate creates a new release then submits it to the scheduler.
+func (s *releasesService) ReleasesCreate(ctx context.Context, db *gorm.DB, r *Release) (*Release, error) {
+	// Lock all releases for the given application to ensure that the
+	// release version is updated automically.
+	if err := db.Exec(`select 1 from releases where app_id = ? for update`, r.App.ID).Error; err != nil {
```
ejholmes: Also found an issue where the previous lock we were obtaining wasn't sufficient to protect the atomicity of the Version of a release. See [TestEmpire_Deploy_Concurrent](https://github.com/remind101/empire/blob/395d91fc570b0e11a7553a899bb8e8c3a12b76f0/tests/empire/empire_test.go#L148-L206).

phobologic:
:+1:

PR #716, opened 2016-01-06
Show delta in ScaleEvent.

ejholmes:
When scaling, this will now show what the previous scale was as well as the delta:

```markdown
ejholmes scaled `web` on acme-inc from 5 to 10 (+5)
```
```markdown
ejholmes scaled `web` on acme-inc from 10 to 5 (-5)
```

### DIFF
--- a/empire.go
+++ b/empire.go
@@ -79,7 +79,7 @@ type Empire struct {
 	ExtractProcfile ProcfileExtractor
 
 	// EventStream service for publishing Empire events.
-	EventStream EventStream
+	EventStream
 }
 
 // New returns a new Empire instance.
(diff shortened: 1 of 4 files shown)


empire.go
```diff
@@ -429,12 +429,7 @@ func (opts ScaleOpts) Event() Event {
 
 // Scale scales an apps process.
 func (e *Empire) Scale(ctx context.Context, opts ScaleOpts) (*Process, error) {
-	p, err := e.scaler.Scale(ctx, opts)
-	if err != nil {
-		return p, err
-	}
-
-	return p, e.EventStream.PublishEvent(opts.Event())
+	return e.scaler.Scale(ctx, opts)
```
sanjayprabhu: We can't publish it here because we don't have the previous quantity?

sanjayprabhu:
So fast! Looks good to me.

:+1:

empire.go
ejholmes: Right.

Issue #714, opened 2016-01-06
Add API maintenance mode


ejholmes:
Sometimes it would be desireable to put the API in maintenance mode so that users can't alter running applications.

ejholmes:
Gonna close this for now. If it feels like it becomes necessary, we can re-open.

PR #717, opened 2016-01-06
Allow empire to be put into a readonly maintenance mode.

ejholmes:
Closes https://github.com/remind101/empire/issues/714

This allows you start Empire with a `--maintenance` flag. Doing this will cause any non-readonly actions to return an error indicating that maintenance mode is enabled for the API:

```console
$ emp set FOO=foobar -a acme-inc
error: The API is currently in maintenance mode: "We're performing a migration sorry!"
```

[Review on Reviewable](https://reviewable.io/reviews/remind101/empire/717)

### DIFF
--- a/empire.go
+++ b/empire.go
@@ -77,6 +77,10 @@ type Empire struct {
 
 	// EventStream service for publishing Empire events.
 	EventStream
+
+	// Set this to a non-nil value to put Empire in readonly maintenance
+	// mode.
+	*MaintenanceMode
 }
 
 // New returns a new Empire instance.
@@ -141,6 +145,10 @@ func (opts CreateOpts) Event() CreateEvent {
 
 // Create creates a new app.
 func (e *Empire) Create(ctx context.Context, opts CreateOpts) (*App, error) {
+	if err := e.MaintenanceMode; err != nil {
+		return nil, err
+	}
+
 	a, err := appsCreate(e.db, &App{Name: opts.Name})
 	if err != nil {
 		return a, err
@@ -167,6 +175,10 @@ func (opts DestroyOpts) Event() DestroyEvent {
 
 // Destroy destroys an app.
 func (e *Empire) Destroy(ctx context.Context, opts DestroyOpts) error {
+	if err := e.MaintenanceMode; err != nil {
+		return err
+	}
+
 	tx := e.db.Begin()
 
 	if err := e.apps.Destroy(ctx, tx, opts.App); err != nil {
@@ -227,6 +239,10 @@ func (opts SetOpts) Event() SetEvent {
 // Config. If the app has a running release, a new release will be created and
 // run.
 func (e *Empire) Set(ctx context.Context, opts SetOpts) (*Config, error) {
+	if err := e.MaintenanceMode; err != nil {
+		return nil, err
+	}
+
 	tx := e.db.Begin()
 
 	c, err := e.configs.Set(ctx, tx, opts)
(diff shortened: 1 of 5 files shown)


empire.go
```diff
@@ -444,6 +480,10 @@ func (opts DeploymentsCreateOpts) Event() DeployEvent {
 
 // Deploy deploys an image and streams the output to w.
 func (e *Empire) Deploy(ctx context.Context, opts DeploymentsCreateOpts) (*Release, error) {
+	if err := e.MaintenanceMode; err != nil {
+		return nil, err
```
ejholmes: This should also encode the error into the jsonmessage stream.

empire.go
```diff
@@ -141,6 +145,10 @@ func (opts CreateOpts) Event() CreateEvent {
 
 // Create creates a new app.
 func (e *Empire) Create(ctx context.Context, opts CreateOpts) (*App, error) {
+	if err := e.MaintenanceMode; err != nil {
```
phobologic: having to boilerplate this out seems like it's going to be forgotten/easily missed with added functionality.  Is there anyway we can make this a function of setting up a new url/route possibly?  Maybe even grouping our URLs by 'read'/'write' (also based on the method used to call them).

ejholmes: Another possibility is to do this as a decorator/middleware around the empire.Empire interface:

```go
type MaintenanceModeEmpire struct {
        *MaintenanceMode
        empire *Empire
}

func (e *MaintenanceModeEmpire) Create(ctx context.Context, opts CreateOpts) (*App, error) {
        if err := e.MaintenanceMode; err != nil {
                return nil, err
        }

        return e.empire.Create(ctx, opts)
}
```

So we would wrap the empire.Empire struct before passing to the server to consume. This would be a little more explicit, because you have to explicitly whitelist each method but ends up being more boilerplate, and it means that control layer packages need to accept an interface instead of a struct, which adds some indirection (but not that much). I can be sold either way.

ejholmes: One advantage of the decorator approach is we could generalize it a bit as an ACLEmpire and re-use it for access control. Something along the lines of:

```go
type Permissions map[string]error

type ACLEmpire struct {
        perms Permissions
        empire *Empire
}

func NewMaintenanceModeEmpire(err *MaintenanceMode, e *Empire) *ACLEmpire {
        return &ACLEmpire{
                empire: e,
                perms: Permissions{
                        "Create": err,
                },
        }
}

func (e *ACLEmpire) Create(ctx context.Context, opts CreateOpts) (*App, error) {
        if err := e.perms["Create"]; err != nil {
                return nil, err
        }

        return e.empire.Create(ctx, opts)
}
```

phobologic: What about a HttpHandler func that we can wrap 'write' methods with?  Less boilerplate, and it's explicit at the routing layer what is possible for blocking with maintenance mode.  Not sure that works well with a generic ACL system though.

ejholmes: We can probably start with that. I wanted to put this in the core API that the control layers consume so that it wouldn't get duplicated. If we put it in the control layers, than we have to duplicate it for the heroku api, github webhooks and slack commands (if we ever merge that PR in).

I think I'm still mostly in favor of just doing what I'm doing right now. You have to think about adding the maintenance mode check if you add a new method to the core API, but honestly, that interface is pretty stable and doesn't change very frequently.

ejholmes:
Gonna close this for now and re-open later if this becomes necessary.

PR #720, opened 2016-01-07
gliderlabs registrator compatibility

protomouse:
This PR sets SERVICE_NAME for [registrator](https://github.com/gliderlabs/registrator). This makes web processes discoverable by the app name in backends supported by registrator, such as Consul.


protomouse:
Fixed failing tests. Sorry about that.

releases.go
```diff
@@ -306,6 +306,9 @@ func newServiceProcess(release *Release, p *Process) *scheduler.Process {
 	if len(ports) > 0 {
 		env["PORT"] = fmt.Sprintf("%d", *ports[0].Container)
 
+		// gliderlabs registrator
+		labels[fmt.Sprintf("SERVICE_%d_NAME", *ports[0].Container)] = release.App.Name
```
ejholmes: Shouldn't this be an environment variable instead of a label?

protomouse: Registrator [reads both env and labels](https://github.com/gliderlabs/registrator/blob/master/bridge/util.go#L33), and the former takes precedence over the latter. My goal is to implement something that can be overridden by config:set. You could of course also achieve this by taking care not to clobber SERVICE_NAME passed by config. While the env implementation adds some pollution to the app (which was another motivator), it might on second thought also be a bit confusing using a label because there will (if overridden) be two values present upon container inspection.

ejholmes: Makes sense to me, I think starting with a label here is the best option! If you're happy with it, I'll merge.

ejholmes:
:+1:

releases.go
protomouse: Go for it.

ejholmes:
Had to revert this change as it looks like ECS doesn't support this format for labels: https://github.com/remind101/empire/issues/733.

Maybe an env var is the way to go? With the stuff coming in https://github.com/remind101/empire/issues/723, maybe this would be nice to only add if it was `--scheduler=ECS_Registrator` or something like that?

protomouse:
Oh crap. Looks like ECS is doing the right thing, and registrator is relying on Docker not conforming to its [own spec](https://docs.docker.com/engine/userguide/labels-custom-metadata/). Yeah, I'll add this as an env variable. Adding it as a separate scheduler configuration makes sense to me, as well. Let's put this on hold for the moment then.

PR #724, opened 2016-01-11
SSL Certificate Documentation

baversjo:
There's a new emp command for uploading ssl certificates, update docs to cover this method instead.

### DIFF
--- a/docs/ssl_certs.md
+++ b/docs/ssl_certs.md
@@ -6,7 +6,7 @@ Upload the certificate using the command below.
 The .pem file should contain a full chain of the public certificate and any root certificates.
 
 ```
-emp ssl-cert-add -a front cert.pem cert.key
+emp ssl-cert-add -a <app> cert.pem cert.key
 ```
 
 **Caveat**: Currently, attaching SSL certificates must happen before you deploy anything to the application (e.g. `emp create` then `emp ssl-cert-add` immediately after).
--- a/docs/ssl_certs.md
+++ b/docs/ssl_certs.md
@@ -2,16 +2,11 @@
 
 Empire allows you to attach IAM server certificates to an applications web process. Follow these steps to attach an SSL certificate.
 
-First, [upload a certificate to IAM](http://docs.aws.amazon.com/cli/latest/reference/iam/upload-server-certificate.html):
+Upload the certificate using the command below.
+The .pem file should contain a full chain of the public certificate and any root certificates.
 
-```console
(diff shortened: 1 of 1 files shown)


docs/ssl_certs.md
```diff
@@ -2,16 +2,11 @@
...
 
-```console
-$ aws iam upload-server-certificate --server-certificate-name myServerCertificate --certificate-body file://public_key_cert_file.pem --private-key file://my_private_key.pem --certificate-chain file://my_certificate_chain_file.pem
 ```
-
-Then attach it to the application:
-
-```console
-$ emp certs-attach arn:aws:iam::<accountId>:server-certificate/myServerCertificate -a <app>
+emp ssl-cert-add -a <app> cert.pem cert.key
```
ejholmes: This is actually the old method :). Empire no longer supports cert uploads. See https://github.com/remind101/empire/pull/701 for background.

We may re-introduce a command to upload certificates, but it would probably be per account instead of per app like it was before.

baversjo: Ah ok! Should we update the command list printed when running emp without any arguments? These methods are  currently listed:
```
   ssl                show ssl endpoint info
   ssl-cert-add       add a new ssl cert
   ssl-cert-rollback  add a new ssl cert
   ssl-destroy        destroy ssl endpoint
```

ejholmes: Ah! Good catch. These are supposed to be hidden, but it looks like they were still getting output from the usage. Opened a pr to fix that: https://github.com/remind101/empire/pull/726

Thanks for the heads up!

hajimenomomo:
Hey guys,
I successfully deployed an empire cluster using stacker, and deployed my first app, however I can't seem to find a way to add a new certificate to my app. 
```
emp cert-attach -a APP_NAME CERT_NAME
``` 
This outputs the following: error: Unexpected error: 404 Not Found.

My current empire version is 0.9.2 and the old ssl commands seem to not be available anymore.

Sincerely,

ejholmes:
Hey @hajimenomomo. This is happening because you're using a 0.9.3 client against a 0.9.2 daemon. Couple of things you can do:

1. Wait for https://github.com/remind101/empire/pull/727 to be merged and use 0.9.3 for daemon and client.
2. Use the docker image `remind101/empire:f20a2f6d43b4b940e58dd84219ac5b6573546188`, which is basically everything that's going into 0.9.3.
3. Downgrade to a 0.9.2 client which includes the ssl commands.

PR #726, opened 2016-01-11
Actually hide hidden commands.

ejholmes:
This hides commands marked as Hidden from the help output. Previous method was broken.

Brought up in https://github.com/remind101/empire/pull/724/files#r49390667


phobologic:
:+1:

PR #727, opened 2016-01-13
Release 0.9.3

ejholmes:
Bumps version to 0.9.3 and will build Linux/OS X binaries once merged.


ejholmes:
We should probably actually bump to 0.10.0 since there are backwards incompatible changes between HEAD daemon and 0.9.2 client:

https://github.com/remind101/empire/pull/724#issuecomment-171263059

phobologic:
Looks good - +1 for 0.10

ejholmes:
Closing and will re-open as 0.10.0

PR #729, opened 2016-01-14
Prettier README

ejholmes:
Adds the logo to the readme and makes the circle CI badge use the shield style so it's consistent with the docs badge.


phobologic:
:+1:

Issue #703, opened 2015-12-22
Add support for Amazon ECR


ejholmes:
How we pass a docker.AuthConfiguration down for pulls should be abstract a bit so that we can generate a docker.AuthConfiguration dynamically to work with ECR, and we can provide various AuthConfigurationProvider implementations:

```go
// Generates a docker.AuthConfiguration from a .dockercfg
FileProvider(file io.Reader) docker.AuthConfiguration {}
// Generates a docker.AuthConfiguration for an Amazon ECR registry.
ECRProvider() docker.AuthConfiguration {}
// Merges multiple docker.AuthConfigurations together.
MultiProvider(providers ...AuthConfigurationProvider) AuthConfigurationProvider
```

protomouse:
FWIW, I took a stab at this today: wowgroup/empire@14bee7db

ejholmes:
@protomouse awesome stuff. If you wanna open a PR, I'd love to get this merged in for the next release.

protomouse:
PR done.

JacopoDaeli:
@ejholmes @protomouse how can we use ECR with EMPIRE ?

ejholmes:
https://github.com/remind101/empire/pull/730 was just merged and will be in the next release.

@JacopoDaeli The only thing you should need to do is add the following permissions to your EC2 container instances instance profile: 

```
ecr:GetAuthorizationToken
ecr:BatchCheckLayerAvailability
ecr:GetDownloadUrlForLayer
ecr:BatchGetImage
```

PR #730, opened 2016-01-14
ECR support

protomouse:
This PR introduces multiple authentication backends, and adds support for the EC2 Container Registry. This addresses #703.


ejholmes:
Awesome! Overall this looks good. I have some minor nits around style consistency, but nothing major I think.

pkg/dockerauth/ecr.go
```diff
@@ -0,0 +1,76 @@
...
+
+type ECR interface {
+	GetAuthorizationToken(*ecr.GetAuthorizationTokenInput) (*ecr.GetAuthorizationTokenOutput, error)
+}
+
+type ecrAuthProvider struct {
+	svc ECR
+}
+
+func NewEcrAuthProvider(svc ECR) *ecrAuthProvider {
```
ejholmes: Can we use `ECR` here to be consistent? `NewECRAuthProvider`.

pkg/dockerauth/interface.go
```diff
@@ -0,0 +1,7 @@
+package dockerauth
+
+import "github.com/fsouza/go-dockerclient"
+
+type AuthProvider interface {
+	GetAuthConfiguration(registry string) (docker.AuthConfiguration, error)
```
ejholmes: We tend to not prefix method calls with `Get`. Can you change this method to just `AuthConfiguration`?

pkg/dockerauth/multi.go
```diff
@@ -0,0 +1,32 @@
...
+func NewMultiAuthProvider(providers ...AuthProvider) *multiAuthProvider {
+	return &multiAuthProvider{
+		providers: providers,
+	}
+}
+
+func (p *multiAuthProvider) GetAuthConfiguration(registry string) (docker.AuthConfiguration, error) {
+	for _, provider := range p.providers {
+		authConf, err := provider.GetAuthConfiguration(registry)
+		if err == nil {
```
ejholmes: Hmm, I wonder if we should have an `ErrNoAuthConfiguration` that we can return from AuthProvider implementations, and only try to next one if the error returned is equal to that, so we can bubble up actual errors, like we do in [package auth](https://github.com/remind101/empire/blob/19c5350e31210926504ed9cadf69a7f024866c66/server/auth/auth.go#L103-L109).

We probably wouldn't want errors like `No ECR authorization data for <registry>` from the ECR provider to go unnoticed.

ejholmes: Another option, instead of using an error, could be to return a `*docker.AuthConfiguration` and give semantic meaning to nil.

cmd/empire/factories.go
```diff
@@ -227,3 +229,21 @@ func newHBReporter(key, env string) (reporter.Reporter, error) {
...
+
+func newAuthProvider(c *cli.Context) (dockerauth.AuthProvider, error) {
+	provider := dockerauth.NewMultiAuthProvider()
+	provider.AddProvider(dockerauth.NewEcrAuthProvider(ecr.New(newConfigProvider(c))))
+
+	dockerConfigFile, err := os.Open(c.String(FlagDockerAuth))
+	if err == nil {
+		defer dockerConfigFile.Close()
+		dockerConfigProvider, err := dockerauth.NewDockerConfigAuthProvider(dockerConfigFile)
+		if err == nil {
```
ejholmes: I think we should return here if `err != nil`.

cmd/empire/factories.go
```diff
@@ -227,3 +229,21 @@ func newHBReporter(key, env string) (reporter.Reporter, error) {
...
 }
+
+// Auth provider =======================
+
+func newAuthProvider(c *cli.Context) (dockerauth.AuthProvider, error) {
+	provider := dockerauth.NewMultiAuthProvider()
+	provider.AddProvider(dockerauth.NewEcrAuthProvider(ecr.New(newConfigProvider(c))))
+
+	dockerConfigFile, err := os.Open(c.String(FlagDockerAuth))
+	if err == nil {
```
ejholmes: Same here, if a docker auth file was provided but can't be opened, we should bubble up that error.

pkg/dockerutil/client.go
```diff
@@ -72,7 +73,7 @@ func (c *Client) PullImage(ctx context.Context, opts docker.PullImageOptions) er
 		reg = "https://index.docker.io/v1/"
```
ejholmes: I think we should move this into the dockerConfigAuthProvider implementation now. Feels like a cleaner place to deal with this.

pkg/dockerutil/client.go
```diff
@@ -72,7 +73,7 @@ func (c *Client) PullImage(ctx context.Context, opts docker.PullImageOptions) er
...
 	}
 
-	if c, ok := c.Auth.Configs[reg]; ok {
+	if c, err := c.AuthProvider.GetAuthConfiguration(reg); err == nil {
```
ejholmes: We should bubble this error up if it's non-nil.

pkg/dockerauth/multi.go
protomouse: While I agree on the point of exposing errors, I think "keep trying to the best of our abilities, in priority order, until we run out of options" is the most suitable approach for authentication in general (making the actual backends fail-fast is of course desirable). OpenSSH and PAM comes to mind. The simplest thing would be to just log the errors then and there. Your thoughts?

cmd/empire/factories.go
protomouse: Makes sense.

cmd/empire/factories.go
protomouse: Aren't there several use cases that do not depend on credentials from Docker config? That is, if the file is not present on disk, it might be intentional. I do see #542, so perhaps there's something I'm missing?

protomouse:
Thanks for the feedback. The things I haven't commented on are already done or have to do with my other comments. I'd appreciate a final round of thoughts from you.

cmd/empire/factories.go
ejholmes: Absolutely. I just think that should be handled with the flag. If the `--docker.auth` flag is an empty string, then we shouldn't add a `dockerConfigAuthProvider` to the chain, instead of ignoring the error. Mostly to prevent misconfiguration (e.g. I really want to use my .dockercfg but I put it in the wrong place).

Basically, I think we should have this:

```go
if fname := c.String(FlagDockerAuth); fname != "" {
        f, err := os.Open(fname)
        if err != nil {
                return nil, err
        }

        p, err := dockerauth.NewDockerConfigAuthProvider(f)
        if err != nil {
                return nil, err
        }

        provider.AddProvider(p)
}
```

pkg/dockerauth/multi.go
ejholmes: I think there's actually 2 concepts of a "Multi" provider here:

1. One where we attempt to find the first AuthConfiguration that matches the registry.
2. One where we degrade gracefully in the case of errors.

I feel like no 1. is what we're mostly trying to achieve here so I actually think changing the return value to `(*docker.AuthConfiguration, error)` would be the best option, and then return `nil, nil` from the ECR provider if the regex doesn't match, and `nil, nil` from the dockerConfig provider if the registry isn't found in the map.

For example, if ECR goes down and I'm deploying an ECR image, seeing `No auth provider can provide for dkr....` would be confusing to me as a user. I don't necessarily expect it to try to fall through to the dockerConfig provider because it's guaranteed to not work.

Hopefully that makes sense.

protomouse:
Thanks for the update. I'll go ahead and implement your suggestions shortly.

protomouse:
I've added some commits implementing these changes. If this looks good to you, I'll squash.

docs/cloudformation.json
```diff
@@ -298,6 +298,13 @@
                 "ec2:Describe*",
                 "elasticloadbalancing:*",
                 "ecs:*",
+                "ecr:GetAuthorizationToken",
+                "ecr:BatchCheckLayerAvailability",
```
ejholmes: Minor but, `ecr:GetAuthorizationToken` looks like it should be the only permission Empire needs?

ejholmes:
Awesome. Looks good to me :+1: 

Feel free to squash and I'll merge.

docs/cloudformation.json
protomouse: I believe anything that wants to pull from ECR using the Docker API needs `GetAuthorizationToken`, `GetDownloadUrlForLayer`, `BatchGetImage` and `BatchCheckLayerAvailability`. These are included in [AmazonEC2ContainerServiceforEC2Role](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/instance_IAM_role.html). `GetRepositoryPolicy`, `DescribeRepositories` and `ListImages` shouldn't be necessary, as these pertain to AWS SDK operations. I copied from [AmazonEC2ContainerRegistryReadOnly](http://docs.aws.amazon.com/AmazonECR/latest/userguide/ecr_managed_policies.html#AmazonEC2ContainerRegistryReadOnly), which is why they're there.

ejholmes: Cool, makes sense.

protomouse:
Squashed and removed the three unnecessary permissions as per above. As expected, Empire works fine without them.

Issue #704, opened 2015-12-23
Support logConfiguration for ECS Task Definitions


brianz:
I'm looking into getting application logs into Loggly or some other hosted system.  Now, I think my ideal solution would use Docker's `syslog` driver to send application logs to the host where I'd then run a daemon to pick those up and ship them.  This design can be extended to any hosted system:

```
container app logs -> stdout -> host syslog -> wherever (Loggly, etc.)...totally configurable
```

One option for logs is https://github.com/gliderlabs/logspout suggested by @phobologic, however that has some limitations AFAICT:

- have to run another container on every Empire minion (arguably not a huge deal)
- can't ship to host's syslog (see https://github.com/gliderlabs/logspout/issues/118 ...bigger deal)

ECS Task Definitions support configuration for the `log_driver` and `log_opts` options:

http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#ContainerDefinition-logConfiguration

I *think* this would be the place to hook into the AWS API to set these options:

https://github.com/remind101/empire/blob/master/pkg/ecsutil/client.go#L60

Which I think is called from the scheduler:

https://github.com/remind101/empire/blob/master/scheduler/ecs/ecs.go#L343

I think there may be some repercussions from setting the `log_driver`, however I think the solution above would be really nice since it puts the owner of the Empire minion in full control over what happens to logs without needing to introduce another moving part.

phobologic:
So the big issue with this is that docker only allows a single log driver to be defined, and when you define anything but `json-file` or `journald` it disables `docker logs`/the api for streaming logs.  That will effectively break the `emp log` functionality.  Not sure if that's ok.

Regarding the logspout limitations - I think the first limitation isn't real, since empire already runs logspout, so it'd just involve modifying that configuration.  The second issue is a bit harder to deal with - largely due to needing to figure out what 'localhost' is for the host your docker containers are running on.  We ran into this when we started running statsd on our hosts, and resolved it with a local dnsmasq, though that is a little complicated.

WE've also talked about figuring out a way to hard code the `docker0` bridge IP (it seems to be the same on all our hosts, so this may not be necessary, but we haven't had a chance to dig into it yet to verify), then you can just configure all your daemons to talk to syslog via the network on that IP.

brianz:
It'd good to know that the Empire AMI runs logspout out of the box:

https://github.com/remind101/empire_ami/tree/master/ansible/roles/logspout

[...]

PR #731, opened 2016-01-20
Add a command line argument to set the logConfiguration

brianz:
First cut of configuring the ``--log-driver`` for from Empire.  Details of this are listed in #704.

In short, this allows controlling the logging driver when Empire starts up. Work remaining to do:

- validate the values passed in on the command line
- add support for log options
- add some tests

I tested this and it's working as expected when using ``empire server --docker.logdriver syslog``

I'd appreciate any feedback since this is my first time in Go.


cmd/empire/main.go
```diff
@@ -27,9 +27,10 @@ const (
 	FlagDBPath = "path"
 	FlagDB     = "db"
 
-	FlagDockerSocket = "docker.socket"
-	FlagDockerCert   = "docker.cert"
-	FlagDockerAuth   = "docker.auth"
+	FlagDockerSocket    = "docker.socket"
+	FlagDockerCert      = "docker.cert"
+	FlagDockerAuth      = "docker.auth"
+	FlagDockerLogDriver = "docker.logdriver"
```
ejholmes: I think this should actually live under the `ecs.*` namespace. The `docker.*` flags control options for connecting to the local docker daemon.

scheduler/ecs/ecs.go
```diff
@@ -47,6 +47,7 @@ type Scheduler struct {
 
 	cluster string
 	ecs     *ecsutil.Client
+	config  Config
```
ejholmes: I don't believe the `Scheduler` struct every needs to directly access the log driver, just the ProcessManager. This can be removed.

scheduler/ecs/ecs.go
```diff
@@ -308,6 +316,7 @@ type ecsProcessManager struct {
 	cluster     string
 	serviceRole string
 	ecs         *ecsutil.Client
+	config      Config
```
ejholmes: The `Config` struct should really only be used by the NewScheduler factory. Can we change this to:

```go
logDriver string
// or
logConfiguration *ecs.logConfiguration
```

And then change the signature of `taskDefinitionInput` to `func taskDefinitionInput(p *scheduler.Process, logConfiguration *ecs.LogConfiguration)`

ejholmes:
Awesome, other than those comments and tests, this looks like a good first step. It'd be nice if we could also provide a way to pass log driver options, but I don't think it's necessary for the first version.

cmd/empire/main.go
```diff
@@ -184,6 +186,18 @@ var EmpireFlags = []cli.Flag{
...
 	cli.StringFlag{
+		Name:   FlagECSLogDriver,
+		Value:  "json-file",
+		Usage:  "Log driver to use when running containers. Maps to the --log-driver docker cli arg",
+		EnvVar: "EMPIRE_ECS_LOG_DRIVER",
+	},
+	cli.StringSliceFlag{
+		Name:   FlagECSLogOpts,
+		Value:  &cli.StringSlice{},
+		Usage:  "Log driver to options. Maps to the --log-opt docker cli arg",
```
ejholmes: Can you give an example of the format you'd pass here? I'd assume `tag=foo,thing=bar`?

scheduler/ecs/ecs.go
```diff
@@ -78,6 +78,12 @@ type Config struct {
 
 	// AWS configuration.
 	AWS client.ConfigProvider
+
+	// Log driver to use when starting ECS tasks. Maps to the --log-driver docker cli arg
+	LogDriver string
+
+	// Log options to use when starting ECS tasks. Maps to the --log-opt docker cli arg
+	LogOpts []string
```
ejholmes: I think we should probably do the `foo=bar` parsing earlier (in cmd/empire) and take a `map[string]string` here instead.

Actually, maybe even just make this `LogConfiguration *ecs.LogConfiguration` and build the LogConfiguration in cmd/empire.

cmd/empire/main.go
brianz: `logopt` takes multiple arguments, which is exactly how it works with `docker run`:

```
empire server --ecs.logdriver syslog --ecs.logopt tag=brianz --ecs.logopt syslog-facility=daemon
```

...results in:

```
        "LogConfig": {
            "Type": "syslog",
            "Config": {
                "syslog-facility": "daemon",
                "tag": "brianz"
            }
        },
```

ejholmes:
Awesome, this looks good to me. If you can rebase on top of master I'll go ahead and merge. :+1:

brianz:
Rebasing was pretty gnarly so I merged in master.  Thanks again for the guidance @ejholmes.

PR #732, opened 2016-01-20
Update Cloudformation template to use latest ECS optimized AMI

jjbohn:
The old one stopped being able to be found by `aws ec2 describe-images` in `bin/bootstrap`


ejholmes:
Thanks! :+1:

rgabo:
There were also some important fixes around `yum` and security packages so always a good practice :wink:

PR #736, opened 2016-01-21
Refactor out an ImageBuilder interface.

ejholmes:
This is a pre-req for the "build on demand" feature in https://github.com/remind101/empire/wiki/Roadmap.

This refactors out an `ImageBuilder` interface in `package server/github`, which can be used to provide various implementations to support building a docker image on demand when a GitHub Deployment is triggered.

Right now, the only implementation is the `ImageFromTemplate` function which directly maps a GitHub deployment event to a docker image like before.

Once [Conveyor](https://github.com/remind101/conveyor) supports a build API, you'll be able to provide a CLI flag to enable Empire to trigger a build and wait for it to complete before deploying it to ECS. I'd like to provide implementations for Docker Hub and Quay if possible, but their API's kinda suck for manually triggering builds. Maybe we should make Conveyor into a paid service? ;)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,13 +3,16 @@
 REPO = remind101/empire
 TYPE = patch
 
+cmds: build/empire build/emp
+
+clean:
+	rm -rf build/*
+
 build/empire:
 	go build -o build/empire ./cmd/empire
 
 build/emp:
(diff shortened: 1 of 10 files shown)


phobologic:
:+1:

PR #737, opened 2016-01-21
Go 1.5.3

ejholmes:
To address [CVE-2015-8618](https://groups.google.com/forum/#!topic/golang-announce/MEATuOi_ei4)

Shouldn't actually effect Empire installations because we don't use a tls server directly.


phobologic:
:+1:

PR #738, opened 2016-01-21
Clarifies/cleans up cert-attach help output

phobologic:


cmd/emp/certs.go
```diff
@@ -8,7 +8,7 @@ import (
 
 var cmdCertAttach = &Command{
 	Run:      runCertAttach,
-	Usage:    "cert-attach",
+	Usage:    "cert-attach <aws_cert_name>",
```
ejholmes: Maybe `<aws_cert_arn>`? Currently, it has to be the full ARN.

phobologic: Cool, updated to make this more apparent throughout the help output.

ejholmes:
:+1:

Issue #739, opened 2016-01-21
emp run tries to pull from Docker Hub even though using Quay.io


shidenkai0:
Hi everyone,

I setup an empire cluster using stacker 0.6.0 with the latest blueprints, using Quay.io as registry, and managed to deploy my applications. Everything works properly except the emp run command, which tries to pull from docker hub, whereas my original deployment is from Quay.io. This results in the following:

```
runner: pull: Error: image username/repo:tag not found
```
Empire seems to be trying to pull from Docker Hub despite DOCKER_REGISTRY being set to Quay.io in the seed file.

Sincerely yours,

ejholmes:
Good catch. I'm 99% positive this is caused [here] (https://github.com/remind101/empire/blob/master/pkg/runner/runner.go#L73-L77), which doesn't duplicate the logic [here](https://github.com/remind101/empire/blob/master/empire.go#L565-L578) (putting the registry portion in the `Repository` field). We should extract that into pkg dockerutil and re-use it for pulling.

PR #740, opened 2016-01-22
Allow emp run to be used with unofficial docker registries.

ejholmes:
Fixes https://github.com/remind101/empire/issues/739.

This extracts the logic that we were using to pull from unofficial registries during deployments into the dockerutil client so it can be re-used for `emp run`.

Tested this locally with the quay.io/remind/acme-inc image and it works as expected.


phobologic:
:+1: awesome

PR #741, opened 2016-01-22
Allow emp run to be run against an EMPIRE_API_URL that includes ports

ejholmes:
Previously, if you tried to use `emp run` with an EMPIRE_API_URL that had the port in it (e.g. `http://localhost:8080`) you would get the following error:

```console
$ emp run bash -a acme-inc
error: dial tcp: too many colons in address 192.168.99.100:8080:80
```


phobologic:
:+1:

Issue #689, opened 2015-12-12
Move migrations to use github.com/rubenv/sql-migrate


ejholmes:
https://github.com/rubenv/sql-migrate is quite a bit better than https://github.com/mattes/migrate.

Primarily:

1. Can pass it a raw `*sql.DB` connection.
2. Can use go-bindata to provide the source of migrations, which would make Empire a completely self contained binary.

PR #742, opened 2016-01-23
Move migrations lib to github.com/rubenv/sql-migrate

ejholmes:
Closes https://github.com/remind101/empire/issues/689

This allows us to compile the migrations into the resulting binary, so that it's completely self contained, which also means we don't need to build the empire binary and run `empire migrate` before running tests.

Migrating the database down is also tested in this now, which caught a few errors and typos in the migrations.


phobologic:
:+1:

ejholmes:
Gonna close this out. https://github.com/remind101/empire/pull/786 made me realize that this still doesn't solve the full story for doing migrations, since github.com/rubenv/sql-migrate doesn't solve plain old Go migrations.

I'm pretty dissatisfied with all of the migration libraries for Go. All I really want is a simple lib that runs some sql and sets the id in a column, so I hacked something together in https://github.com/ejholmes/migrate. Only a couple hundred lines and does everything we need without all the unnecessary cruft. I'll open a PR with that as an example.

PR #745, opened 2016-01-29
GitHub team auth

mhahn:
This allows users to limit empire access to members of a specific team within a github organization.

Github doesn't support fetching teams by name without a non trivial amount of work, so you have to use the team id.

To get the team id:

`curl  -i -u <username>:<password> https://api.github.com/orgs/<organization>/teams`

testing:
- wrote tests
- tested locally that i was rejected when not a member of a team, allowed access once i was a member

### DIFF
--- a/cmd/empire/server.go
+++ b/cmd/empire/server.go
@@ -109,12 +109,12 @@ func newAuthenticator(c *cli.Context, e *empire.Empire) auth.Authenticator {
 	}
 
 	// After the user is authenticated, check their GitHub Team membership.
-	if teamId := c.String(FlagGithubTeam); teamId != "" {
+	if teamID := c.String(FlagGithubTeam); teamID != "" {
 		authorizer := githubauth.NewTeamAuthorizer(client)
-		authorizer.TeamId = teamId
+		authorizer.TeamID = teamID
 
 		log.Println("Adding GitHub Team authorizer with the following configuration:")
-		log.Println(fmt.Sprintf("  Team ID: %v ", teamId))
+		log.Println(fmt.Sprintf("  Team ID: %v ", teamID))
 
 		return auth.WithAuthorization(
 			authenticator,
--- a/cmd/empire/server.go
+++ b/cmd/empire/server.go
@@ -116,7 +116,11 @@ func newAuthenticator(c *cli.Context, e *empire.Empire) auth.Authenticator {
 		log.Println("Adding GitHub Team authorizer with the following configuration:")
 		log.Println(fmt.Sprintf("  Team ID: %v ", teamId))
 
-		return auth.WithAuthorization(authenticator, authorizer)
+		return auth.WithAuthorization(
+			authenticator,
+			// Cache the team check for 30 minutes
+			auth.CacheAuthorization(authorizer, 30*time.Minute),
+		)
 	}
 
 	return authenticator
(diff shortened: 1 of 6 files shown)


cmd/empire/server.go
```diff
@@ -108,5 +108,16 @@ func newAuthenticator(c *cli.Context, e *empire.Empire) auth.Authenticator {
 		)
 	}
 
+	// After the user is authenticated, check their GitHub Team membership.
+	if teamId := c.String(FlagGithubTeam); teamId != "" {
```
ejholmes: Minor, but can we change the casing from `Id` to `ID`. There's some doc from google about how it's not idiomatic to use camel casing for things like `ID` and we try to follow that.

server/auth/github/client.go
```diff
@@ -173,6 +177,29 @@ func (c *Client) IsMember(organization, token string) (bool, error) {
 	return true, nil
 }
 
+// IsTeamMember returns true if the given user is a member of the team.
+func (c *Client) IsTeamMember(teamId, name string, token string) (bool, error) {
```
ejholmes: Kinda feels like the method signature should be `(teamID, token string)` and then just fetch the github user to grab their login first. It's an extra call, but `empire.User.Name` may not always be their GitHub login in the future.

ejholmes:
Looks great :+1:

cmd/empire/server.go
```diff
@@ -108,5 +108,16 @@ func newAuthenticator(c *cli.Context, e *empire.Empire) auth.Authenticator {
...
+		authorizer := githubauth.NewTeamAuthorizer(client)
+		authorizer.TeamId = teamId
+
+		log.Println("Adding GitHub Team authorizer with the following configuration:")
+		log.Println(fmt.Sprintf("  Team ID: %v ", teamId))
+
+		return auth.WithAuthorization(authenticator, authorizer)
```
ejholmes: Should we cache this authorization, like we do with the [org auth](https://github.com/mhahn/empire/blob/github-team-auth/cmd/empire/server.go#L107)? Without caching, every empire API hit incurs a full 1-2 round trips to GitHub, which makes everything slow. An in memory ~30 minute cache has worked pretty well.

server/auth/github/client.go
mhahn: what if we added:
```
type User struct {
	Name        string `json:"name"`
	GitHubToken string `json:"-"`
    GitHubLogin string `json:"github_login"`
}
```

or some such thing.

I think the current flow of getting the user in `authenticate` and then having all github authorizers reference values from the empire user makes sense.

cmd/empire/server.go
mhahn: ah, i didn't realize this was on every API call, I thought it just happened once when the user logged in.

definitely makes sense to cache.

server/auth/github/client.go
ejholmes: Problem with that right now is backwards compat, since existing JWT tokens won't have the `GitHubLogin` field embedded. It wouldn't be a big deal to require Empire installations to re-generate a new signing secret and have users log in again, but if we can avoid that it's probably best.

I think I'd prefer just hitting the /user endpoint of the github api to grab their login using the given token, but either one is fine imo. Up to you.

mhahn: ok got it. just dug through how auth works for a request and I see now that we only call `authenticate` via github the first time.

mhahn:
thanks for the review @ejholmes!

PR #747, opened 2016-01-29
Add an ImageBuilder implementation for Conveyor

ejholmes:
This adds an implementation of the ImageBuilder interface backed by Conveyor. This means that if you're using Conveyor to build Docker images, it's no longer necessary to wait for the image to finish building before triggering a GitHub Deployment.

If a GitHub Deployment request is received by Empire, Empire will consult with Conveyor to obtain the Docker image for the given git sha, building it if necessary.

Not yet ready to merge, still needs some testing on both Conveyor's side and Empire's side.

### DIFF
--- a/cmd/empire/main.go
+++ b/cmd/empire/main.go
@@ -22,8 +22,11 @@ const (
 
 	FlagGithubWebhooksSecret           = "github.webhooks.secret"
 	FlagGithubDeploymentsEnvironments  = "github.deployments.environment"
+	FlagGithubDeploymentsImageBuilder  = "github.deployments.image_builder"
 	FlagGithubDeploymentsImageTemplate = "github.deployments.template"
 	FlagGithubDeploymentsTugboatURL    = "github.deployments.tugboat.url"
+
+	FlagConveyorURL = "conveyor.url"
 
 	FlagDBPath = "path"
 	FlagDB     = "db"
@@ -120,9 +123,15 @@ var Commands = []cli.Command{
 				EnvVar: "EMPIRE_GITHUB_DEPLOYMENTS_ENVIRONMENT",
 			},
 			cli.StringFlag{
+				Name:   FlagGithubDeploymentsImageBuilder,
+				Value:  "template",
+				Usage:  "Determines how the Docker image to deploy is determined when a GitHub Deployment event is received. Possible options are `template` and `conveyor`.",
+				EnvVar: "EMPIRE_GITHUB_DEPLOYMENTS_IMAGE_BUILDER",
+			},
+			cli.StringFlag{
 				Name:   FlagGithubDeploymentsImageTemplate,
 				Value:  github.DefaultTemplate,
-				Usage:  "A Go text/template that will be used to determine the docker image to deploy.",
+				Usage:  "A Go text/template that will be used to determine the docker image to deploy. This flag is only used when `--" + FlagGithubDeploymentsImageBuilder + "` is set to `template`.",
 				EnvVar: "EMPIRE_GITHUB_DEPLOYMENTS_IMAGE_TEMPLATE",
 			},
 			cli.StringFlag{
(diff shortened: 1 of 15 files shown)


server/github/builder.go
```diff
@@ -34,3 +37,40 @@ func ImageFromTemplate(t *template.Template) ImageBuilder {
...
+		client: c,
+	}
+}
+
+func (c *ConveyorImageBuilder) BuildImage(ctx context.Context, event events.Deployment) (image.Image, error) {
+	w := ioutil.Discard // TODO: Allow ImageBuilder implementations to write to the deployment logs?
+
+	a, err := c.client.Build(w, conveyor.BuildCreateOpts{
+		Repository: event.Repository.FullName,
+		Branch:     "", // TODO,
```
ejholmes: This should _probably_ be `event.Deployment.Ref`. Ultimately, Conveyor will need to support "branchless" builds, which should be pretty easy since the branch is only used to optimize the layer cache and isn't explicitly necessary to perform a build.

This is where things get a little weird, since we don't actually know if the ref that the deployment was triggered for was a branch, tag or sha.

ejholmes: Actually, I think we should not set Branch at all since we really have no idea what the ref was. In most cases, Conveyor will have already started the build, using the branch cache and Empire will just wait for it to complete.

ejholmes:
Note to self. We should definitely have some cancellations around deployment. Don't wanna have goroutines with pending deployments hanging around forever.

ejholmes:
Ok. I think this is good to merge now. Been running this in staging for the last couple of days and no major issues aside from https://github.com/ejholmes/cloudwatch/pull/4

cmd/empire/main.go
```diff
@@ -120,9 +123,15 @@ var Commands = []cli.Command{
...
+				Name:   FlagGithubDeploymentsImageBuilder,
+				Value:  "template",
+				Usage:  "Determines how the Docker image to deploy is determined when a GitHub Deployment event is received. Possible options are `template` and `conveyor`.",
+				EnvVar: "EMPIRE_GITHUB_DEPLOYMENTS_IMAGE_BUILDER",
+			},
+			cli.StringFlag{
 				Name:   FlagGithubDeploymentsImageTemplate,
 				Value:  github.DefaultTemplate,
-				Usage:  "A Go text/template that will be used to determine the docker image to deploy.",
+				Usage:  "A Go text/template that will be used to determine the docker image to deploy. This flag is only used when `--" + FlagGithubDeploymentsImageBuilder + "` is set to `template`.",
```
phobologic: Is this required if `FlagGithubDeploymentsImageBuilder` is set to template?

cmd/empire/main.go
```diff
@@ -131,6 +140,12 @@ var Commands = []cli.Command{
 				Usage:  "If provided, logs from deployments triggered via GitHub deployments will be sent to this tugboat instance.",
 				EnvVar: "EMPIRE_TUGBOAT_URL",
 			},
+			cli.StringFlag{
+				Name:   FlagConveyorURL,
+				Value:  "",
+				Usage:  "When combined with the `--" + FlagGithubDeploymentsImageBuilder + "` flag, this determines where the location of a Conveyor instance is to perform Docker image builds.",
```
phobologic: This is only if `FlagGithubDeploymentsImageBuilder` is set to conveyor right?

cmd/empire/main.go
phobologic: NM, I see the default

cmd/empire/server.go
```diff
@@ -46,8 +47,19 @@ func newServer(c *cli.Context, e *empire.Empire) http.Handler {
...
+	switch builder {
+	case "template":
+		tmpl := template.Must(template.New("image").Parse(c.String(FlagGithubDeploymentsImageTemplate)))
+		return github.ImageFromTemplate(tmpl)
+	case "conveyor":
+		s := conveyor.NewService(conveyor.DefaultClient)
+		s.URL = c.String(FlagConveyorURL)
+		return github.NewConveyorImageBuilder(s)
+	default:
+		panic(fmt.Sprintf("unknown image builder: %s", builder))
```
phobologic: Maybe return the allowed values?

server/github/builder.go
```diff
@@ -12,20 +14,20 @@ import (
 // ImageBuilder is an interface that represents something that can build and
 // return a Docker image from a GitHub commit.
 type ImageBuilder interface {
-	BuildImage(ctx context.Context, event events.Deployment) (image.Image, error)
+	BuildImage(ctx context.Context, w io.Writer, event events.Deployment) (image.Image, error)
```
phobologic: I'm actually not sure what the standard way to do this in Go is, but would it be useful to have some documentation on what these arguments are?  Like it's not immediately obvious what `w` is used for here.

phobologic:
:+1:

cmd/empire/main.go
ejholmes: Yeah, I'll update to clarify.

server/github/builder.go
ejholmes: Yeah, the description of this interface is a little lacking. I'll update.

Issue #750, opened 2016-02-02
Bootstrap script doesn't quote keypair


xiongchiamiov:
If you try to enter a keypair name that has a space in it, the bootstrap script complains:

    AWS SSH KeyName: Test Server
    Do you have a docker account & want to use it for private repo access? [y/N] 
    ==> Launching empire in AZs: us-east-1a us-east-1b, Cloudformation Stack empire-f49c73cf
    
    Error parsing parameter '--parameters': Expected: '=', received: 'EOF' for input:
    Server
          ^

Manually quoting this as `'Test Server'` works as you'd expect, but the script should [quote](http://mywiki.wooledge.org/Quotes) appropriately.

phobologic:
Good catch, simple fix - https://github.com/remind101/empire/pull/752

Issue #749, opened 2016-02-02
md5sum location disparity


xiongchiamiov:
`bin/bootstrap` checks to see if `/usr/bin/md5sum` exists, and if it does, instead tries to use `/usr/sbin/md5sum`.  These should be the same path, probably the former, or the path to the binary shouldn't be relied on and the `PATH` should be consulted.

phobologic:
https://github.com/remind101/empire/pull/752 for this as well.  Thanks.

PR #752, opened 2016-02-03
Fix bin/bootstrap issues

phobologic:
Fixes GH-750


phobologic:
Also Fixes GH-749

ejholmes:
:+1:

Issue #755, opened 2016-02-09
Add deployed version id to notification message


turadg:
Request to add the `env-v123` identifier in these messages:
https://remind.slack.com/archives/builds/p1455050430001916

Would make it much easier to cross reference when checking that the deployed frontend assets are reaching my browser.

PR #758, opened 2016-02-11
Adds environment & version to Deploy events

phobologic:
Fixes GH-755


empire.go
```diff
@@ -425,6 +425,9 @@ type DeploymentsCreateOpts struct {
 	// Image is the image that's being deployed.
 	Image image.Image
 
+	// Environment is the environment where the image is being deployed
+	Environment string
```
ejholmes: This is probably ok for now, but it'd probably be better in the long run if we had this on the Empire struct itself, so it works for non-github deployments. If I `emp deploy <image>`, I'd still expect to see the environment in the event.

That means we'd need an `--environment` flag.

empire.go
```diff
@@ -455,6 +462,9 @@ func (e *Empire) Deploy(ctx context.Context, opts DeploymentsCreateOpts) (*Relea
 		return r, err
 	}
 
+	event := opts.Event()
+	event.Release = r.Version
+
 	return r, e.PublishEvent(opts.Event())
```
ejholmes: Think this should be `PublishEvent(event)`.

ejholmes:
lgtm if you wanna just start with this :+1:

empire.go
phobologic: Hmm - yeah, I guess actually taking the environment FROM the github deployment isn't actually all that useful.  It's not like Empire runs multiple environments.  Let me think about that - doesn't seem like it'd need to be super far reaching?

ejholmes: I think the initial version could just be used for this.

phobologic: Ok, took a first stab at addressing this.  Let me know what you think.

empire.go
```diff
@@ -438,6 +448,10 @@ func (opts DeploymentsCreateOpts) Event() DeployEvent {
 	if opts.App != nil {
 		e.App = opts.App.Name
 	}
+
+	if opts.Environment != "" {
+		e.Environment = opts.Environment
```
ejholmes: I don't think environment should be on the DeploymentsCreateOpts at all, you can just set the Environment on the even in the `Deploy` method:

```go
event.Environment = e.Environment
```

Which means that you don't need the GetEnvironment() method and the server doesn't need to know anything about passing it.

server/server.go
```diff
@@ -19,6 +19,8 @@ var (
 type Options struct {
 	Authenticator auth.Authenticator
 
+	Environment string
```
ejholmes: This doesn't appear to get used.

phobologic:
Cool updated - thanks

empire.go
```diff
@@ -284,6 +287,10 @@ func (e *Empire) DomainsDestroy(ctx context.Context, domain *Domain) error {
 	return nil
 }
 
+func (e *Empire) GetEnvironment() string {
```
ejholmes: This isn't used anymore.

ejholmes:
Nice! Looks like it needs a rebase, but :+1: otherwise.

ejholmes:
Oh, can you update the changelog as well?

phobologic:
Done!

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * It's now possible to lock down the GitHub authorization to a specific team via the `--github.team.id` flag [#745](https://github.com/remind101/empire/pull/745).
 * Empire can now integrate with Conveyor to build Docker images on demand when using the GitHub Deployments integration [#747](https://github.com/remind101/empire/pull/747).
 * Stdout and Stdin from interactive run sessions can now be sent to CloudWatch Logs for longterm storage and auditing [#757](https://github.com/remind101/empire/pull/757).
+* Add `Environment` and `Release` to Deploy Events. `--environment` will likely be used for tagging resource later. [#758](https://github.com/remind101/empire/pull/758)
```
ejholmes: Nit, s/resource/resources/g

Issue #762, opened 2016-03-02
ECS logconfiguration should default to nil unless overriden


ejholmes:
After https://github.com/remind101/empire/pull/731, we default to json-file, which will override any `docker daemon` defaults (and could break backwards compatibility). We should default to a nil log configuration, so the docker defaults are respected.

PR #763, opened 2016-03-02
Default to the docker daemon default log driver.

ejholmes:
Fixes https://github.com/remind101/empire/issues/762

In certain configurations, it's more desirable to use the default logging driver provided when running `docker daemon`, which won't always be `json-file`.


phobologic:
:+1:

PR #764, opened 2016-03-04
Force json-file logging driver for attached runs.

ejholmes:
Attached runs depends on the `json-file` logging driver, so we want to make sure it's used regardless of the logging driver set at `docker daemon` time.


phobologic:
:+1:

PR #765, opened 2016-03-05
Allow running 'emp help' without setting EMPIRE_API_URL.

### LINKED ISSUE #744: HTTP client should be lazy loaded in emp
Right now, if you use a command that doesn't require an http client (e.g. help, about), you'll get the following error:

```bash
$ emp help about
error: EMPIRE_API_URL must be set
```

This is leftover cruft from forking hk, which initializes the http client in `main`.

A simple workaround is to set `EMPIRE_API_URL`, but it'd be a better experience if you didn't have to do this.

/cc @turadg

brianestlin:
Minor annoyance. LMK what you think @ejholmes .

Fixes #744

### DIFF
--- a/cmd/emp/help.go
+++ b/cmd/emp/help.go
@@ -14,7 +14,6 @@ import (
 
 var helpEnviron = &Command{
 	Usage:    "environ",
-	NoClient: true,
 	Category: "emp",
 	Short:    "environment variables used by emp",
 	Long: `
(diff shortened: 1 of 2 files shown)


ejholmes:
Awesome! :+1:

turadg:
:clap:

phobologic:
:+1: Woot

PR #775, opened 2016-03-31
scale -l now shows configured scale, not running

### LINKED ISSUE #769: emp scale -l is misleading
`emp scale -l`returns counts as seen by `emp ps` not what's currently configured. This means that during deployments or when processes cannot start, it returns the wrong values. I expect it to return the currently configured scaling settings, not the counts of processes running.

phobologic:
Fixes #769

### DIFF
--- a/releases.go
+++ b/releases.go
@@ -230,15 +230,15 @@ func createFormation(db *gorm.DB, release *Release) error {
 	return nil
 }
 
-// GetCurrentFormations gets the current formations for an app
-func GetCurrentFormations(db *gorm.DB, app *App) (*Formation, error) {
+// currentFormations gets the current formations for an app
+func currentFormation(db *gorm.DB, app *App) (Formation, error) {
 	// Get the current release
 	current, err := releasesFind(db, ReleasesQuery{App: app})
 	if err != nil {
 		return nil, err
 	}
 	f := current.Formation()
-	return &f, nil
+	return f, nil
 }
 
 // ReleasesLastVersion returns the last ReleaseVersion for the given App.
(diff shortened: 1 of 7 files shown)


sanjayprabhu:
What do you think about changing it to print one item per line (so we can pipe it to grep)? Currently for applications with a lot of processes, it's hard to find the process you are looking for.

phobologic:
Honestly I like having it's output be 'ready made' for just using with the regular `emp scale` command.  I actually do use grep with this, but there's two things that have to be considered there:

1. For some reason `emp scale -l` writes to stderr, so you need to ` 2>&1 `.  Annoying, but I'm sure there was a reason for that.  I should look into it.
2. Make sure your grep does color highlighting of matches.

You can also do `emp scale -l -a acme-inc 2>&1 | tr " " "\\n"` and it should work.  Obviously not as convenient though.

releases.go
```diff
@@ -230,6 +230,17 @@ func createFormation(db *gorm.DB, release *Release) error {
 	return nil
 }
 
+// GetCurrentFormations gets the current formations for an app
+func GetCurrentFormations(db *gorm.DB, app *App) (*Formation, error) {
```
ejholmes: No need to use a pointer for `map` types, since they're already passed by reference:

```go
func GetCurrentFormations(db *gorm.DB, app *App) (Formation, error)
```

ejholmes: Just for consistency with the rest of the functions like this, can we un-export it and change the pluralization? (also, general convention in go is to never use "get")

```go
func currentFormation(db *gorm.DB, app *App) (Formation, error)
```

ejholmes:
This looks great. :+1:

Issue #718, opened 2016-01-06
Need to stop using uuid from code.google.com


phobologic:
When installing the latest `emp` command:

```
# go get -u github.com/remind101/empire/cmd/emp
warning: code.google.com is shutting down; import path code.google.com/p/go-uuid/uuid will stop working
```

benguillet:
note: this is only a problem is you haven't activated go1.5+ vendoring (`export GO15VENDOREXPERIMENT=1`)

ejholmes:
I'll update the docs to be more explicit about requiring Go 1.5 and setting `GO15VENDOREXPERIMENT`.

phobologic:
Ahh - ok, I was on 1.5, but hadn't set that variable, I'll add that to my env.

PR #776, opened 2016-04-02
Update git repo url for uuid

jperichon:
Related to https://github.com/remind101/empire/issues/718

The repo has now been completely moved away from Google Code: fetching it fails as opposed to just return a warning like before.

```
❯ go get code.google.com/p/go-uuid/uuid
package code.google.com/p/go-uuid/uuid: unable to detect version control system for code.google.com/ path
```


jperichon:
Nevemind, this is not needed. For some environmental reason, `go build` was not using my `vendor` directory.

PR #778, opened 2016-04-06
Update ECS AMI

benguillet:
Update to the latest ECS AMI as seen in the docs: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/launch_container_instance.html. Latest should fix an issue with the agent that stop accepting incoming requests.


phobologic:
:+1:

PR #780, opened 2016-04-11
Validate parsed accessed tokens.

ejholmes:
I've seen cases where `Claims["User"]["Name"]` is an empty string. I don't have a good repro for when that can happen (last time I saw it, I had signed in when the GitHub API was having issues), but this will validate the access token before creating it and after decoding it.


phobologic:
:+1:

Issue #751, opened 2016-02-02
Verify that aws-sdk-go is retrying throttling errors


ejholmes:
We see ThrottlingException's pretty frequently when running `emp ps`. The aws clients should be retrying this with backoff, so the fact that we're still seeing this makes me question if it's actually retrying.

```console
$ emp ps 
error: ThrottlingException: Rate exceeded
    status code: 400, request id: 4e339002-c9df-11e5-97c0-d708bdd78d89
```

ejholmes:
Definitely looks like it's retrying, but the default retry limit is [3 retries](https://github.com/aws/aws-sdk-go/blob/9d7bc2d6ca2ada0468f705f0e9725ca97f8550c8/aws/client/client.go#L54) and with the algorithm in the [default retrier](https://github.com/aws/aws-sdk-go/blob/9d7bc2d6ca2ada0468f705f0e9725ca97f8550c8/aws/client/default_retryer.go#L34-L37), this will retry over the course of about 500ms, so it's still pretty easy to hit throttling while retrying.

Maybe the best workaround would be to initially bump this up to something like 6 (retry 6 times over ~5 seconds)? It means that `emp ps` and `emp deploy` would be slower for apps with big Procfiles, but atleast it wouldn't return an error.

For future reference, reproduced these using this go program:

```go
package main

import (
	"log"

	"github.com/aws/aws-sdk-go/aws"
	"github.com/aws/aws-sdk-go/aws/session"
	"github.com/aws/aws-sdk-go/service/ecs"
)

const concurrent int = 150

var client *ecs.ECS

func main() {
	config := &aws.Config{MaxRetries: aws.Int(15)}
	config.WithLogLevel(aws.LogDebugWithRequestRetries)
	client = ecs.New(session.New(config))

	var i int
	for {
		i++
		if err := get(); err != nil {
			log.Fatal(err)
		}
	}
}

func get() error {
	errCh := make(chan error, concurrent)

	for i := 0; i < concurrent; i++ {
		go func() {
			_, err := client.DescribeTaskDefinition(&ecs.DescribeTaskDefinitionInput{
				TaskDefinition: aws.String("047bc51a-8d16-4dba-97f9-43f53e22f008--web"),
			})
			errCh <- err
		}()
	}

	for i := 0; i < concurrent; i++ {
		err := <-errCh
		if err != nil {
			return err
		}
	}

	return nil
}
```

ejholmes:
[...]

PR #781, opened 2016-04-12
Reduce number of DescribeTaskDefinition calls.

ejholmes:
Related to https://github.com/remind101/empire/issues/751.

Previously we naively described the task definition for each ECS task when running `emp ps`. In reality, most tasks will share the same task definition, so this means we make a lot of duplicate and unnecessary `DescribeTaskDefinition` api calls.

This will describe each unique task definition in the list of tasks, then store it in a map. The calls to DescribeTaskDefinition could also be parallelized, but I'll save that for a different PR.


phobologic:
Awesome :+1:

PR #782, opened 2016-04-12
Describe task definitions in parallel.

ejholmes:
For apps with a large number of processes defined in the Procfile, describing each task definition sequentially can take some time. The ECS api doesn't provide an endpoint for describing multiple task definitions, so this describes them in parallel.

### DIFF
--- a/scheduler/ecs/ecs.go
+++ b/scheduler/ecs/ecs.go
@@ -238,19 +238,9 @@ func (m *Scheduler) Instances(ctx context.Context, appID string) ([]*scheduler.I
 		return instances, err
 	}
 
-	taskDefinitions := make(map[string]*ecs.TaskDefinition)
-	for _, t := range tasks {
-		k := *t.TaskDefinitionArn
-
-		if _, ok := taskDefinitions[k]; !ok {
-			resp, err := m.ecs.DescribeTaskDefinition(ctx, &ecs.DescribeTaskDefinitionInput{
-				TaskDefinition: t.TaskDefinitionArn,
-			})
-			if err != nil {
-				return instances, err
-			}
-			taskDefinitions[k] = resp.TaskDefinition
-		}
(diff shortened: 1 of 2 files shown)


scheduler/ecs/ecs.go
```diff
@@ -288,6 +278,41 @@ func (m *Scheduler) Instances(ctx context.Context, appID string) ([]*scheduler.I
...
+			ch <- struct {
+				resp *ecs.DescribeTaskDefinitionOutput
+				err  error
+			}{resp, err}
+		}(arn)
+	}
+
+	for arn := range taskDefinitions {
+		result := <-ch
+		if err := result.err; err != nil {
```
phobologic: Should we make this a little smarter when it comes to errors like the Throttling error?  Seems like we're asking for more of those with this, right?

ejholmes: I think with https://github.com/remind101/empire/pull/781, we're less likely to hit throttling even if we parallelize like this. With my test program, I had to do about 150 DescribeTaskDefinitions in parallel until I got throttled.

It would definitely still be possible if you did something like `while true; do emp ps; done`, but in that case all we can really do us increase the number of retries in the AWS client.

scheduler/ecs/ecs.go
```diff
@@ -288,6 +278,41 @@ func (m *Scheduler) Instances(ctx context.Context, appID string) ([]*scheduler.I
...
+			return nil, err
+		}
+		taskDefinitions[arn] = result.resp.TaskDefinition
```
ejholmes: Have to use result.resp.TaskDefinition.Arn as the key here. `arn` will be the incorrect value.

ejholmes:
I'll close this one for now, since it's mostly fast enough since https://github.com/remind101/empire/pull/781

PR #783, opened 2016-04-13
Remove ecs.ProcessManager interface

ejholmes:
This is a step towards allowing multiple "web" processes per app. I created a [list of steps](https://github.com/remind101/empire/wiki/Extended-Procfile-Roadmap) that will get us towards supporting an extended Procfile where we can define health checks, exposure settings, etc.

This is basically just collapsing LBProcessManager, and ecsProcessManager into ecs.Scheduler to make it easier to work with. No implementation changes within methods.

The next step will be to move port allocation within ecs.Scheduler.


scheduler/ecs/ecs.go
```diff
@@ -424,7 +397,83 @@ func (m *ecsProcessManager) updateCreateService(ctx context.Context, app *schedu
 	return m.createService(ctx, app, p)
 }
 
-func (m *ecsProcessManager) Processes(ctx context.Context, appID string) ([]*scheduler.Process, error) {
+func (m *Scheduler) loadBalancer(ctx context.Context, app *scheduler.App, p *scheduler.Process) (string, error) {
```
ejholmes: Nothing has actually changed within this method, it's identical to the old `LBProcessManager.CreateProcess` method.

phobologic: should this just be `createLoadBalancer` then? The name of the method was a little confusing - also, if not, maybe a small comment to describe the point of this?  `loadBalancer` doesn't scream 'this creates something' to me.

phobologic:
This is great. Seriously love this refactor, so much easier to understand. :+1:

PR #784, opened 2016-04-13
ECR auth: multi-region support

protomouse:
Since ECR has [launched in additional regions](https://aws.amazon.com/about-aws/whats-new/2016/03/amazon-ec2-container-registry-available-in-eu-ireland/), I've added support by passing a factory function to `NewECRAuthProvider`.

This also addresses a stupid oversight with my original implementation, when running Empire in a region different from the ECR image it is trying to pull.

### DIFF
--- a/cmd/empire/factories.go
+++ b/cmd/empire/factories.go
@@ -268,8 +268,11 @@ func newHBReporter(key, env string) (reporter.Reporter, error) {
 // Auth provider =======================
 
 func newAuthProvider(c *cli.Context) (dockerauth.AuthProvider, error) {
+	awsSession := newConfigProvider(c)
 	provider := dockerauth.NewMultiAuthProvider()
-	provider.AddProvider(dockerauth.NewECRAuthProvider(ecr.New(newConfigProvider(c))))
+	provider.AddProvider(dockerauth.NewECRAuthProvider(func(region string) dockerauth.ECR {
+		return ecr.New(awsSession, &aws.Config{Region: aws.String(region)})
+	}))
 
 	if dockerConfigPath := c.String(FlagDockerAuth); dockerConfigPath != "" {
 		dockerConfigFile, err := os.Open(dockerConfigPath)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 * `emp run` now works with unofficial Docker registries [#740](https://github.com/remind101/empire/pull/740).
 * `emp scale -l` now lists configured scale, not the running processes [#769](https://github.com/remind101/empire/pull/769)
 * Fixed a bug where it was previously possible to create a signed access token with an empty username [#780](https://github.com/remind101/empire/pull/780)
+* ECR authentication now supports multiple regions, and works independently of ECS region [#784](https://github.com/remind101/empire/pull/784)
 
 **Performance**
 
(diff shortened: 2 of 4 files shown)


protomouse:
Alright, hold this for a bit. I'm seeing authorization errors after deploying this to our staging environment.

cmd/empire/factories.go
```diff
@@ -269,7 +269,9 @@ func newHBReporter(key, env string) (reporter.Reporter, error) {
 
 func newAuthProvider(c *cli.Context) (dockerauth.AuthProvider, error) {
 	provider := dockerauth.NewMultiAuthProvider()
-	provider.AddProvider(dockerauth.NewECRAuthProvider(ecr.New(newConfigProvider(c))))
+	provider.AddProvider(dockerauth.NewECRAuthProvider(func(region string) dockerauth.ECR {
+		return ecr.New(newConfigProvider(c), &aws.Config{Region: aws.String(region)})
```
protomouse: Perhaps it's smarter to close over an instance of ConfigProvider, rather than construct it every time the factory is called. Looks like the AWS SDK will [create a copy](https://github.com/aws/aws-sdk-go/blob/af3026711cd873ccb4e87a7142fa8ca0be8f2aef/aws/session/session.go#L108) of the session, which should make it safe to do so.

ejholmes: You could also just close over a map so that the ecr clients are memoized. Something like:

```go
var clients map[region]*ecr.ECR
provider.AddProvider(dockerauth.NewECRAuthProvider(func(region string) dockerauth.ECR {
        var client *ecr.ECR
        client, ok := clients[region]
        if !ok {
                client = ecr.New(newConfigProvider(c), &aws.Config{Region: aws.String(region)})
                clients[region] = client
        }
        return client
}
```

Although, it's probably fine to instantiate a new client every time.

ejholmes:
Nice catch! This looks great. I had one comment, if you want to make that change. Otherwise, this looks good. Do you mind updating the CHANGELOG as well?

cmd/empire/factories.go
protomouse: If wonder if by doing memoization here you introduce thread-safety concerns. According to the docs, `http.Serve` runs each request in its own goroutine, and I don't think Go makes any atomicity guarantees in most cases.

I did a bit of benchmarking on session creation vs copying:

```go
func BenchmarkAwsSessionCreate(b *testing.B) {
  var svc *ecr.ECR

  for i := 0; i < b.N; i++ {
    svc = ecr.New(session.New(), &aws.Config{Region: aws.String("us-east-1")})
  }

  _ = svc
}

func BenchmarkAwsSessionCopy(b *testing.B) {
  sess := session.New()
  var svc *ecr.ECR

  for i := 0; i < b.N; i++ {
    svc = ecr.New(sess, &aws.Config{Region: aws.String("us-east-1")})
  }

  _ = svc
}
```

```
BenchmarkAwsSessionCreate-8	  200000	      9235 ns/op
BenchmarkAwsSessionCopy-8  	  300000	      5401 ns/op
```

So I think I'll move it out for now.

protomouse:
Ok, moved session creation out of the factory and added a row to "Bugs" in the CHANGELOG.

ejholmes:
Awesome. Thanks!

mcansky:
is there a specific documentation section on the use of these changes ?

phobologic:
@protomouse any chance you could provide some documentation for using ECR?  It's catching a few people up, and quite honestly I haven't used it before - digging into the code, but it's not quite clear what changes need to be made in config to use ECR so far.

protomouse:
@phobologic Sure. See #1034!

PR #785, opened 2016-04-14
Move port allocation to ECS scheduler.

ejholmes:
This is step 2 in https://github.com/remind101/empire/wiki/Extended-Procfile-Roadmap

This moves port allocation from Empire core into the ECS scheduler implementation and simplifies the interface between both.

Since the instance port limitation is an ELB limitation, I think it makes more conceptual sense to allocate an instance port when we create an ELB. With this change, when we create an ELB, we allocate a port from the pool, then release the port back to the pool when we delete that ELB. This allows us to provision any number of ELB's across the cluster without running into port conflicts.

The next step will be to collapse the [scheduler.Process.Exposure](https://github.com/remind101/empire/blob/e1beb5d5de9ab21ec69a40498c1f39668ba991db/scheduler/scheduler.go#L72-L73) and [scheduler.Process.SSLCert](https://github.com/remind101/empire/blob/e1beb5d5de9ab21ec69a40498c1f39668ba991db/scheduler/scheduler.go#L89) attributes into a single `Exposure` attribute that allows us to define HTTP/HTTPS/TCP/SSL exposures. I'll save that for a different PR.


migrations/0013_add_process_column_to_ports.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE ports ADD COLUMN taken text;
+UPDATE ports SET taken = 't' WHERE port = (SELECT port FROM ports WHERE app_id != NULL);
```
ejholmes: This should also set app_id to null. This migration should probably remove that column, since it'll be unused after this change.

scheduler/ecs/ecs.go
```diff
@@ -743,16 +757,6 @@ func (e *LoadBalancerExposureError) Error() string {
 	return fmt.Sprintf("Process %s is %s, but load balancer is %s. An update would require me to delete the load balancer.", e.proc.Type, e.proc.Exposure, lbExposure)
 }
 
-// LoadBalancerPortMismatchError is returned when the port stored in the data store does not match the ELB instance port
-type LoadBalancerPortMismatchError struct {
```
ejholmes: This error case is no longer possible, since the ELB determines the host port once it's been created.

phobologic:
:+1:

PR #786, opened 2016-04-15
Use []string internally for commands

ejholmes:
Passing a string is error prone, because we have to parse it in a way that matches bash's behavior (e.g. single quotes are 1 argument). It's a lot simpler and less prone to errors if we do this parsing at the very edge of user input, then use a `[]string` internally, since a `[]string` is a more accurate representation of command arguments (e.g. `os.Args` is a []string).

We should change the `command` field on the processes table to a postgres array and store it as such internally, but that requires a migration that uses the shellwords lib, which we don't have good support for right now.

### DIFF
--- a/processes.go
+++ b/processes.go
@@ -4,6 +4,7 @@ import (
 	"database/sql"
 	"database/sql/driver"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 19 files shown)


processes.go
```diff
@@ -57,20 +59,49 @@ func (p ProcessType) Value() (driver.Value, error) {
...
+	//	Command{"echo", "hello world"}
+	//
+	// Then stringify it:
+	//
+	//	"echo hello world"
+	//
+	// Then parse it again:
+	//
+	//	Command{"echo", "hello", "world"}
+	return driver.Value(c.String()), nil
```
ejholmes: Now that I think about it, this may not be possible without the migration, since parsing -> stringifying -> parsing again isn't pure.

procfile/extractor.go
```diff
@@ -50,7 +49,7 @@ func (e *CMDExtractor) Extract(img image.Image) (Procfile, error) {
 		return pm, err
 	}
 
-	pm["web"] = strings.Join(i.Config.Cmd, " ")
```
ejholmes: This change actually fixes a bug if you specified a Docker CMD like:

```
CMD ["nginx", "-g", "daemon off;"]
```

processes.go
phobologic: The beginning of the spiral :)

ejholmes: Indeed :P

phobologic:
:+1:

ejholmes:
Once https://github.com/remind101/empire/pull/787 is merged, I'll cherry pick https://github.com/remind101/empire/commit/f362b39492c9c91685ef3dadb16301ef0cc5ff36 onto this, which changes the processes column to a jsonb type and migrates all of the data to be stored as an array.

ejholmes:
@phobologic now that #787 is merged, I added another commit to this that migrates the database to store the command as an array in a jsonb column. Tested this migration with our staging database, which worked as expected.

phobologic:
👍 looks good

processes.go
```diff
@@ -129,7 +159,7 @@ func (cm CommandMap) Value() (driver.Value, error) {
 	for k, v := range cm {
 		m[string(k)] = sql.NullString{
 			Valid:  true,
-			String: string(v),
+			String: v.String(),
```
ejholmes: Dang, just realized that this is a problem. By stringifying it here, we lose information.

It would technically be fixed with https://github.com/remind101/empire/pull/792, but might be worth fixing in a separate PR so we can roll out the recent migrations more incrementally to check for problems.

ejholmes: I think right now, it would only be a problem during a rollback, since that's the only time the slug would actually be loaded from the database.

PR #787, opened 2016-04-15
Switch migrations lib to github.com/remind101/migrate

### LINKED ISSUE #689: Move migrations to use github.com/rubenv/sql-migrate

ejholmes:
Closes #689

A simpler alternative to https://github.com/remind101/empire/pull/742 and a pre-req to https://github.com/remind101/empire/pull/786. This pulls the database migrations into the binary (which also means the binary is now completely self contained) and makes it easy for us to generate migrations that include Go code:

```go
migrate.Migration{
        ID: 1234,
        Up: func(tx *sql.Tx) error {
                // Any Go code.
                return nil
        },
}
```

The schema generated is identical as before.


migrations/migrations.go
```diff
@@ -0,0 +1,247 @@
+package migrations
+
+import "github.com/ejholmes/migrate"
+
+var Migrations = []migrate.Migration{
```
ejholmes: Generated this automatically from the sql migrations with:

```ruby
#!/usr/bin/env ruby

#fileRegex = /(\d{4})_(\w*)\.(up|down)\.sql/
fileRegex = /0*(\d*?)_(\w*)\.(up|down)\.sql/

migrations = Dir["migrations/*"].each_with_object({}) do |f, migrations|
  match = fileRegex.match(f.gsub('migrations/', ''))
  id, name, dir = match[1], match[2], match[3]
  id = Integer(id)
  migrations[id] ||= {}
  migrations[id][dir] = File.read(f)
end

puts "package migrations"
puts ""
puts "import \"github.com/ejholmes/migrate\""
puts ""
puts "var migrations = []migrate.Migration{"
migrations.each do |id, migration|
  up = migration["up"].split(';').map(&:strip).compact
  down = migration["down"].split(';').map(&:strip).compact

  puts "        {"
  puts "                ID:   #{id},"
  puts "                Up:   migrate.Queries([]string{"
  up.each do |query|
  break if query == ""
  puts "                        `#{query}`,"
  end
  puts "                }),"
  puts "                Down: migrate.Queries([]string{"
  down.each do |query|
  break if query == ""
  puts "                        `#{query}`,"
  end
  puts "                }),"
  puts "        },"
end
puts "}"

```

phobologic:
This looks good - but it might be worth having at least one more set of eyes look at it since it gets into the database for these migrations.

bmarini:
👍

PR #788, opened 2016-04-16
Change scheduler.Exposure type.

ejholmes:
This is step 4 in https://github.com/remind101/empire/wiki/Extended-Procfile-Roadmap#step-4---refactor-how-we-define-process-exposure

This changes how exposure settings are communicated in the `scheduler` package. After this, we can add a `TCPExposure` type and change the load balancer package to be able to create tcp load balancers.

This is all internal for now and will eventually be exposed in the extended Procfile.

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -10,27 +10,6 @@ import (
 	"golang.org/x/net/context"
 )
 
-type Exposure int
-
-func (e Exposure) String() string {
-	switch e {
-	case ExposeNone:
-		return "none"
-	case ExposePrivate:
-		return "private"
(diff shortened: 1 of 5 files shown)


scheduler/scheduler.go
```diff
@@ -73,11 +52,37 @@ type Process struct {
 
 	// ulimit -u
 	Nproc uint
+}
+
+// Exposure controls the exposure settings for a process.
+type Exposure struct {
```
ejholmes: This could also allow the consumer to specify the container port to use, but mostly unnecessary for now.

phobologic:
👍

Issue #789, opened 2016-04-19
Lock migrations


ejholmes:
Once https://github.com/remind101/migrate/pull/3 is merged, pull that in here. Some recent migrations are long running (relatively speaking), and it would be good to ensure that only 1 process is ever running migrations.

PR #790, opened 2016-04-19
Lock migrations.

ejholmes:
Closes https://github.com/remind101/empire/issues/789

This will cause the migrations to obtain an advisory lock before running, so we can ensure that migrations aren't run twice across multiple processes (easy to do with the `--automigrate` flag).

Once https://github.com/remind101/migrate/pull/3 is reviewed, I'll update the dependency here.

### DIFF
--- a/db.go
+++ b/db.go
@@ -15,6 +15,8 @@ type DB struct {
 	*gorm.DB
 
 	uri string
+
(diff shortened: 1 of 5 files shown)


phobologic:
👍

PR #791, opened 2016-04-20
Package 12factor

ejholmes:
For part of Step 5/6 in https://github.com/remind101/empire/wiki/Extended-Procfile-Roadmap#step-5---refactor-how-we-store-process-configuration

This extracts just the new `App`, `Process` and `Manifest` types from https://github.com/remind101/empire/pull/696 and switches the current scheduler to use it.

The `twelvefactor` package just defines a more accurate Go representation of what we determine to be a twelve factor application.

The primary change is that the `Image` is now defined on the `App` and the shared environment/labels are defined on the `App` as well, which is a more accurate representation of a twelve factor app.

### DIFF
--- a/12factor/12factor.go
+++ b/12factor/12factor.go
@@ -1,0 +1,121 @@
+// Package twelvefactor provides types to represents 12factor applications,
+// which are defined in http://12factor.net/
+package twelvefactor
+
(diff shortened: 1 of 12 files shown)


ejholmes:
Gonna close this. There's some nice changes in this, but they can be done in smaller increments.

PR #792, opened 2016-04-21
Extended Procfile and new Formation storage

ejholmes:
Step 5 and Step 6 of https://github.com/remind101/empire/wiki/Extended-Procfile-Roadmap

This defines the initial schema for the extended Procfile format and changes how we store process configuration (formation) in postgres.

Previously, when we deployed something we would:

1. Extract the `Procfile` from the Docker image.
2. Convert the `procfile.Procfile` type to a `CommandMap (map[ProcessType]Command)`.
3. Store this as an hstore column on the `slugs` table.
3. Insert individual `Processes` in the `processes` table for each command in the procfile.

This was problematic because:

a. The previous `process_types` column on the slugs table wouldn't allow us to store additional process configuration from the extended procfile format.
b. The previous `processes` table would also be difficult to extend to store new configuration settings for the process (e.g. health checks, exposure, etc).

With this change, how we store the process configuration is more generalized by storing a `Formation` on the `slugs` and `releases` tables. Now, when we deploy something we:

1. Extract the `Procfile` from the Docker image.
2. Store this in yaml format in the `procfile` column on the `slugs` table.
3. Convert the Procfile to a Formation.
3. Merge the new Formation from the slug with the existing Formation from the previous release and store this in the `formation` column on the `releases` table.

The formation is stored in json format in a jsonb column, and would look something like this:

```json
{
  "adhoc": {
    "Memory": 536870912,
    "Command": [
      "newrelic-admin",
      "run-program",
      "python",
      "scripts/scheduler.py",
      "--start_date=$START_DATE",
      "--end_date=$END_DATE",
      "--tasks=$TASKS",
      "--nocp"
    ],
    "CPUShare": 256
  },
  "daily": {
    "Memory": 536870912,
    "Command": [
      "newrelic-admin",
      "run-program",
      "python",
      "scripts/scheduler.py",
      "--nocp"
    ],
    "CPUShare": 256
  },
  "worker": {
    "Memory": 536870912,
    "Command": [
      "newrelic-admin",
      "run-program",
      "python",
      "scripts/worker.py"
    ],
    "CPUShare": 256
  }
}
```

This will allow us to easily extend the `empire.Process` type to allow for additional attributes, like health checks, exposure settings, etc. I think this might end up being simpler than trying to map the `Formation` to a set of tables.

With this change, Empire will now also begin supporting the new extended Procfile format, which initially just allows for the `command`:

```yaml
web:
  command: ./bin/web
worker:
  command: ./bin/worker
```

But very easy to extend with new settings moving forward.

**TODO**

* [x] Documentation

### DIFF
--- a/constraints.go
+++ b/constraints.go
@@ -1,0 +1,81 @@
+package empire
+
+import (
+	"encoding/json"
+	"fmt"
+
+	. "github.com/remind101/empire/pkg/bytesize"
+	"github.com/remind101/empire/pkg/constraints"
+)
+
+var (
+	Constraints1X = Constraints{constraints.CPUShare(256), constraints.Memory(512 * MB), constraints.Nproc(256)}
+	Constraints2X = Constraints{constraints.CPUShare(512), constraints.Memory(1 * GB), constraints.Nproc(512)}
+	ConstraintsPX = Constraints{constraints.CPUShare(1024), constraints.Memory(6 * GB), 0}
+
+	// NamedConstraints maps a heroku dynos size to a Constraints.
+	NamedConstraints = map[string]Constraints{
+		"1X": Constraints1X,
+		"2X": Constraints2X,
+		"PX": ConstraintsPX,
+	}
+
+	// DefaultConstraints defaults to 1X process size.
+	DefaultConstraints = Constraints1X
+)
+
+// Constraints aliases empire.Constraints type to implement the
+// json.Unmarshaller interface.
+type Constraints constraints.Constraints
+
+func parseConstraints(con string) (*Constraints, error) {
+	if con == "" {
+		return nil, nil
+	}
+
+	if n, ok := NamedConstraints[con]; ok {
+		c := Constraints(n)
+		return &c, nil
+	}
+
+	c, err := constraints.Parse(con)
+	if err != nil {
+		return nil, err
+	}
+
+	r := Constraints(c)
+	return &r, nil
+}
+
+func (c *Constraints) UnmarshalJSON(b []byte) error {
+	var s string
+
+	if err := json.Unmarshal(b, &s); err != nil {
+		return err
+	}
+
+	cc, err := parseConstraints(s)
+	if err != nil {
+		return err
+	}
+
+	if cc != nil {
+		*c = *cc
+	}
+
+	return nil
+}
+
+func (c Constraints) String() string {
+	for n, constraint := range NamedConstraints {
+		if c == Constraints(constraint) {
+			return n
+		}
+	}
+
+	if c.Nproc == 0 {
+		return fmt.Sprintf("%d:%s", c.CPUShare, c.Memory)
+	} else {
+		return fmt.Sprintf("%d:%s:nproc=%d", c.CPUShare, c.Memory, c.Nproc)
+	}
+}
--- a/docs/deploying_an_application.md
+++ b/docs/deploying_an_application.md
@@ -17,13 +17,22 @@ To see an example application, you can look at [remind101/acme-inc]
 
 Empire was modeled after the Heroku API.  As such, many concepts and commands are similar.  One concept which has bled over and which is a requirement is that of a [Procfile][procfile].  For your application to actually work, you must include a Procfile at the root level of your application.
 
-Imagine a Python application which simply ran celery.  A `Procfile` for such a service could look like this:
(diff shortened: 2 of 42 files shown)


constraints.go
```diff
@@ -0,0 +1,81 @@
+package empire
```
ejholmes: Nothing in this file is actually new, just moved it out of `processes.go`.

migrations.go
```diff
@@ -301,4 +303,104 @@ ALTER TABLE apps ADD COLUMN exposure TEXT NOT NULL default 'private'`,
 			`ALTER TABLE processes ADD COLUMN command text not null`,
 		}),
 	},
+	{
+		ID: 15,
+		Up: func(tx *sql.Tx) error {
```
ejholmes: Tested this migration with our staging database and it ran fine.

docs/deploying_an_application.md
```diff
@@ -17,13 +17,22 @@ To see an example application, you can look at [remind101/acme-inc]
...
 
-Imagine a Python application which simply ran celery.  A `Procfile` for such a service could look like this:
+Empire supports two Procfile formats; `standard` and `extended`. The `standard` format is probably what you're used to if you've used Heroku before and simply maps a process name to a command to run:
 
+```yaml
+worker: celery -A tasks worker --loglevel=info
 ```
-workerservice: celery -A tasks worker --loglevel=info
+
+The `extended` format is Empire specific and allows you to define additional configuration for processes, like exposure settings (http/https/tcp/ssl), health checks, and more. An example of the Procfile above in the `extended` format would simply be:
```
phobologic: Are you planning on adding these features in this release?  If not, I'd just mention that these are things we want to allow in the future.

migrations.go
```diff
@@ -301,4 +303,104 @@ ALTER TABLE apps ADD COLUMN exposure TEXT NOT NULL default 'private'`,
...
+			_, err := tx.Exec(`ALTER TABLE slugs ADD COLUMN formation jsonb`)
+			if err != nil {
```
phobologic: Might be nice to have a comment about what each migration is about going forward, just to give some context.

processes.go
```diff
@@ -98,182 +56,83 @@ func (c Command) String() string {
...
-// Process holds configuration information about a Process Type.
+// Process holds configuration information about a Process.
 type Process struct {
-	ReleaseID string
-	ID        string
-	Type      ProcessType
-	Quantity  int
-	Command   Command
-	Constraints
+	Command  Command              `json:"Command,omitempty"`
```
phobologic: How is a process tied to a Release now?

phobologic:
Big commit - exciting stuff! 👍

docs/deploying_an_application.md
ejholmes: Not in this PR, but most likely in the next release. I can update these docs for now.

migrations.go
ejholmes: Yeah good call. I'll add a comment to the top of the migrations.

processes.go
ejholmes: The `Process` type is just part of the `Formation` type, which is a `map[string]Process`. The `Formation` gets json encoded and stored in the `formation` column in postgres.

migrations.go
```diff
@@ -378,4 +379,107 @@ ALTER TABLE apps ADD COLUMN exposure TEXT NOT NULL default 'private'`,
 			`ALTER TABLE slugs ADD COLUMN process_types hstore not null`,
 		}),
 	},
+
+	// This migration changes that way we store process configuration for
+	// releases and slugs, to instead store a Formation in JSON format.
+	{
+		ID: 15,
+		Up: func(tx *sql.Tx) error {
```
ejholmes: Note to self. Add a `Down` migration here.

ejholmes:
Ok, made a small change to this. Instead of extracting a Formation from the Procfile and storing that in the Slug, we just store the raw Procfile and convert that to a Formation in memory. I think that'll be better in the long run, since we're storing the source of truth.

releases.go
```diff
@@ -122,6 +127,7 @@ func (s *releasesService) Rollback(ctx context.Context, db *gorm.DB, opts Rollba
 		App:         app,
 		Config:      r.Config,
 		Slug:        r.Slug,
+		Formation:   r.Formation,
```
ejholmes: One of the really cool things about this change is that the rollback operation is significantly simpler. Since we now have a Formation object, we can just copy the old formation to the new release.

PR #793, opened 2016-04-21
Store process_types as json.

ejholmes:
This addresses https://github.com/remind101/empire/pull/786#discussion_r60528030.

https://github.com/remind101/empire/pull/792 will supercede this, but want to fix this migration before that one is merged.


phobologic:
👍

PR #794, opened 2016-04-22
Enable single transaction mode for migrations.

ejholmes:
When we release the next version of Empire, it's going to contain at least 3 fairly large database migrations. When users are upgrading, it'd be more desirable if all of those migrations are treated atomically. If any of the migrations fail, it's better if the entire migration rolls back so that they can easily downgrade to the previous version instead of getting stuck in an inconsistent migration state.

Depends on https://github.com/remind101/migrate/pull/5. Once that is merged, I'll update the dep here.

### DIFF
--- a/db.go
+++ b/db.go
@@ -36,10 +36,18 @@ func OpenDB(uri string) (*DB, error) {
 		return nil, err
 	}
 
+	m := migrate.NewPostgresMigrator(conn)
+	// Run all migrations in a single transaction, so they will be rolled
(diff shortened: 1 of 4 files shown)


phobologic:
👍 Also, we should think about the messaging here.  We should be pretty explicit letting people know, somehow, that they should backup their empire database before doing this upgrade or something.

ejholmes:
Yep, definitely. Before we release the next version, we should add something to the doc about upgrading Empire.

PR #800, opened 2016-04-26
Allow any process to expose an HTTP/TCP service

ejholmes:
This adds a new `ports` setting to the extended Procfile format, which allows you to attach a load balancer to any process in the app, not just the "web" process.

This is definitely a WIP and looking for feedback. There's some open questions, and a lot of polish needed.

If you want to play around with this locally, you can `emp deploy` [`remind101/acme-inc:extended`](https://github.com/remind101/acme-inc/compare/extended)

**TODO**

* [x] Moar tests.
* [x] <strike>Allow tcp/ssl load balancers to actually get created.</strike> First version of this will be http/https only
* [x] <strike>If the process originally had no load balancer, you can't add a load balancer to it (ECS limitation). The workaround is to remove the process definition, then add it again with the `expose` setting. This is awkward though, because the existing ECS service kinda sticks around in a draining state for a while.</strike> Not this PR.
* [x] <strike>Currently, I'm thinking that the extended Procfile format shouldn't make any assumptions about the `web` process. But, when people switch to the extended Procfile format, they'll almost certainly forget to add the expose setting. Need to think about a good way to prevent this.</strike> I think this is fine once https://github.com/remind101/empire/pull/801 is merged.
* [x] <strike>Right now, I'm not creating CNAME's for anything other than the web process. That could probably be better.</strike> Probably not in this PR.
* [x] Don't remove ELB's when deleting a process? https://github.com/remind101/empire/pull/801
* [x] <strike>For https/ssl exposures, it will use the single attached certificate on the app. People might want to use different certs for different processes.</strike> Not this PR.
* [x] Migration


ejholmes:
>  If the process originally had no load balancer, you can't add a load balancer to it (ECS limitation). The workaround is to remove the process definition, then add it again with the expose setting. This is awkward though, because the existing ECS service kinda sticks around in a draining state for a while.

To account for this, I've considered changing how we name ECS services to include a hash of the exposure settings. So if you change a processes exposure, it'll be a new ECS service and new ELB (like `<app-id>--web--<sha1 of exposure>`).

Probably wouldn't do that in this PR though.

scheduler/ecs/ecs.go
```diff
@@ -514,8 +514,12 @@ func (m *Scheduler) loadBalancer(ctx context.Context, app *scheduler.App, p *sch
 			Tags:     tags,
 		}
 
-		if e, ok := p.Exposure.Type.(*scheduler.HTTPSExposure); ok {
+		switch e := p.Exposure.Type.(type) {
+		case *scheduler.HTTPSExposure:
 			opts.SSLCert = e.Cert
+		case *scheduler.HTTPExposure:
+		default:
+			return nil, fmt.Errorf("can't create %s load balancer", e.Protocol())
```
ejholmes: // TODO

extractor.go
```diff
@@ -212,20 +212,29 @@ func formationFromStandardProcfile(p procfile.StandardProcfile) (Formation, erro
...
 			Command: cmd,
+			Expose:  exposure,
 		}
 	}
 
 	return f, nil
 }
 
-func formationFromExtendedProcfile(p procfile.ExtendedProcfile) (Formation, error) {
+func formationFromExtendedProcfile(app *App, p procfile.ExtendedProcfile) (Formation, error) {
```
ejholmes: These functions could use some unit tests.

extractor.go
```diff
@@ -190,20 +190,20 @@ func firstFile(tr *tar.Reader) ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
-func formationFromProcfile(p procfile.Procfile) (Formation, error) {
+func formationFromProcfile(app *App, p procfile.Procfile) (Formation, error) {
 	switch p := p.(type) {
 	case procfile.StandardProcfile:
-		return formationFromStandardProcfile(p)
+		return formationFromStandardProcfile(app, p)
```
phobologic: nitpick, but maybe `formationFromLegacyProcfile`?

processes.go
```diff
@@ -63,6 +63,14 @@ type Process struct {
 	Memory   constraints.Memory   `json:"Memory,omitempty"`
 	CPUShare constraints.CPUShare `json:"CPUShare,omitempty"`
 	Nproc    constraints.Nproc    `json:"Nproc,omitempty"`
+	Expose   *Exposure            `json:"Expose,omitempty"`
+}
+
+// Exposure holds exposure settings for the Process.
+type Exposure struct {
+	Protocol string `json:"Protocol"`
+	External bool   `json:"External"`
```
phobologic: Is it worth maybe making this a string, rather than a bool?  I could see a world where we have different 'types' of subnets, and where the ELB goes is based on this.  Maybe not worth dealing with here.

extractor.go
ejholmes: I've been referring to the `map[string]string` procfile format as "standard", but I could switch to "legacy". If I switch it, I'd want to do it across the board. What do you think?

processes.go
ejholmes: Yeah. Definitely could see different "visibilities" in the future. I opted to just go with a bool for now, and figure we can just run a migration later if we need to open it up.

ejholmes:
Ok, I think this is probably mergable once https://github.com/remind101/empire/pull/801 is merged. I've addressed all of the todo's, added some tests and added a migration.

procfile/README.md
```diff
@@ -42,3 +42,13 @@ Specifies the command that should be run when executing this process
...
 command: ./bin/web
 ```
+
+**Expose**
+
+Specifies the exposure settings for the process. This allows you to expose a tcp/http/https/ssl service on any process in the application.
+
+```yaml
+expose:
+  type: tcp // Possible options are tcp/http/https/ssl
```
ejholmes: One scenario that I haven't really thought through with this yet is when you want to have multiple listeners. For example, our existing behavior when creating https load balancers is to create both an https and an http listener. Maybe we should do something like this:

```yaml
web:
  visibility: public
  ports:
    - "80:8080":
       protocol: http
    - "443:8080":
       protocol: https
    - "1234:8081":
       protocol: tcp
    - "4321:8081":
       protocol: ssl
```

This might be nice because it offers pretty close parity with `ports` in [docker-compose.yml](https://docs.docker.com/compose/compose-file/#ports).

The awkward bit is how we communicate the protocol, which we need to create the correct listeners on the load balancer. We could provide some sensible defaults, like:

```yaml
web:
  visibility: public
  ports:
    - 80:80 # http listener
    - 443:80 # https listener
    - "1234:1234" # tcp listener
    - # ??? not really a good way to handle an ssl listener.
```

This would also mean we could avoid setting the PORT environment variable and just let containers bind to the port they specify.

ejholmes: Actually, I need to check if this would even be possible with ECS. To do this, I think we'd need to allocate an instance port per container port (e.g. port 8080 and port 8081 get different instance ports). Last time I checked, you could only add 1 LoadBalancer object when creating an ECS service, and the load balancers can't be updated once the ecs service is created.

Maybe the format above is good, but we just need some validations for ECS/ELB limitations.

ejholmes: Just tested this with ECS. It looks like it's possible, but their docs don't make it very intuitive. When we create an ECS, we provide an array of [ecs.LoadBalancer](http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_LoadBalancer.html). That makes you provide a ContainerPort and ContainerName, but those don't seem to serve any purpose other than validation. So, if we create a load balancer with multiple listeners, and a task definition with multiple port mappings, that should be fine.

procfile/procfile.go
```diff
@@ -47,11 +95,16 @@ func Parse(r io.Reader) (Procfile, error) {
 
 // ParseProcfile takes a byte slice representing a YAML Procfile and parses it
 // into a Procfile.
-func ParseProcfile(b []byte) (Procfile, error) {
-	p, err := parseStandardProcfile(b)
+func ParseProcfile(b []byte) (p Procfile, err error) {
+	p, err = parseStandardProcfile(b)
```
mhahn: does it make sense to have some sort of version specification in the Procfile?

I know vagrant did this with Vagrantfiles. I would think of `StandardProcfile` as v1, `ExtendedProcfile` as v2, you might have some v3, v4 etc in the future that you don't know about yet. This current method works for the two that are markedly different, but I think could get confusing in the future and implicitly being dropped into one format when you wanted to be in another.

mhahn:
looks awesome 👍

procfile/procfile.go
ejholmes: Yeah, definitely was thinking about it. The only reason I didn't was because it means we'd probably want to nest the processes. e.g.

```yaml
version: 'v1'
processes:
  web:
    command: ./bin/web
```

which would be fine, but mostly wanted to avoid it if possible. I figure if the time comes where we need a v3, then we can do that and v1/v2/v3 will still be easy to differentiate.

procfile/README.md
mhahn: i like having the ability for each port mapping to define its protocol.

you could also support inference for the http/https use case just because it is pretty common.

ie:

```
web:
  visibility: public
  ports:
    - 80:8080
    - 443:8080
    - 4321:8081
      protocol: tcp
```

procfile/procfile.go
mhahn: yea, makes sense

ejholmes:
For the record. Putting this on hold until CloudFormation is the default scheduling backend, which will make everything we want to do with extended Procfile much simpler.

procfile/README.md
v-yarotsky: I'm sooo looking forward to this

ejholmes:
I've updated this with complete support in the CloudFormation backend. You can now define a Procfile like this:

```yaml
web: 
  command: acme-inc server
  ports:
    - "80:80"
    - "8080:8080":
        protocol: "tcp"
api:
  command: acme-inc api
  ports:
    - "80:80"
```

Allowing you to define pretty sophisticated scenarios. HTTP, HTTPS, TCP and SSL protocols are all supported.

The `web` process will still get some default exposure settings if a `ports: ` key is not provided. It's essentially equivalent to:

```yaml
web:
  ports:
    - "80:8080":
        protocol: "http"
```

And web is still special, in that it gets a CNAME at `<app>.empire`.

Also, the visibility (`external`/`internal`) is still controlled via `emp domains`, but I'd plan on allowing that to be overridden with a `visibility: ` key, like this:

```yaml
web:
  visibility: public
  ports:
    - "80:8080"
```

If you wanna try it out, you can run this and deploy the `remind101/acme-inc:extended` image.

This requires some changes in `scheduler.Exposure` that would be really hard to support in the legacy ECS backend, so you'll notice that this PR currently removes the legacy ECS backend. I think, we should hold off on merging this until we've released 0.11.0, [which is the release where we'll make the CloudFormation backend the default](https://github.com/remind101/empire/wiki/Scheduler-Deprecation-Plan#0110), at which point `master` can become the tip for 1.0.0.

phobologic:
👍 awesome

ejholmes:
Ok, rebased this. This should be mergeable once https://github.com/remind101/empire/pull/1001 has been merged.

ejholmes:
ALB wasn't released when I originally opened this, and there are some limitations when using ALB (see the last commit). Basically, when using ALB + dynamic port mapping, you cannot have multiple listeners going to multiple container ports. It makes sense when you think about it; ECS is registering the container in a target group with a single dynamically allocated port, so alb has no way to know how to route to multiple ports.

The below will work fine when using ELB + instance ports, however, if ALB is enabled, it will yield an error now:

```yaml
web:
  ports:
    - "80:80"
    - "8080:8080"
```

This is totally fine with ALB + dynamic ports (all host ports map to a single container port):

```yaml
web:
  ports:
    - "80:8080"
    - "443:8080":
         protocol: https
    - "8080:8080"
```

I think this is only a limitation with dynamic port mapping, so if we allocated an instance port for ALB, instead of passing `Host: 0`, then it would work the same as ELB. Maybe we could make that configurable?

releases.go
```diff
@@ -315,6 +321,17 @@ func newSchedulerProcess(release *Release, name string, p Process) *scheduler.Pr
...
 
+	var exposure *scheduler.Exposure
+	// For `web` processes defined in the standard procfile, we'll
+	// generate a default exposure setting and also set the PORT
+	// environment variable for backwards compatability.
+	if name == webProcessType && len(p.Ports) == 0 {
+		exposure = standardWebExposure(release.App)
+		env["PORT"] = "8080"
+	} else {
+		exposure = processExposure(release.App, p)
```
phobologic: Is there no way to share the port info?  SEems like we should keep the env var around, no matter what, if at all possible.  Users may be relying on it.

phobologic: Ahh, good point - so this only happens when there are multiple ports defined?  That seems fine.

ejholmes: Right. For backwards compatibility, if you defined a `web` proc, but no `ports:` then it's equivalent to:

```yaml
web:
  environment:
    PORT: "8080"
  ports:
    - "80:8080":
        protocol: "http"
    - "443:8080": # If a cert is provided
        protocol: "https"
```

phobologic: Sounds good!  Ignore me!

ejholmes:
I'll go ahead and merge this for now. There will probably be some follow up polish, and additional docs, but overall this seems to be working great.

Issue #798, opened 2016-04-22
Add some protection around destroying ELB's


ejholmes:
It would really suck if someone happened to remove the `web` process in the Procfile, then deployed it, since that would result in Empire removing the existing ELB for that process.

IMO, we probably shouldn't ever actually remove a provisioned ELB until the app is deleted.

PR #801, opened 2016-04-27
Only destroy load balancers when the app is destroyed.

ejholmes:
Fixes https://github.com/remind101/empire/issues/798

I think it's a lot safer if we only ever destroy a load balancer when the entire app is destroyed, rather than per process. With the existing behavior, it would be possible to deploy a Procfile without the `web` process and the load balancer would be destroyed, making it more difficult to recover from (if you have CNAME's pointed at it).

With the new behavior, it does mean that you could potentially accumulate unused load balancers, but those can always be destroyed manually if it's a real problem.

### DIFF
--- a/scheduler/ecs/ecs.go
+++ b/scheduler/ecs/ecs.go
@@ -193,7 +193,7 @@ func (m *Scheduler) Submit(ctx context.Context, app *scheduler.App) error {
 	return nil
 }
 
-// Remove removes any ECS services that belong to this app.
+// Remove removes all of the AWS resources for this app.
 func (m *Scheduler) Remove(ctx context.Context, appID string) error {
 	processes, err := m.Processes(ctx, appID)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


scheduler/ecs/ecs.go
```diff
@@ -527,16 +527,18 @@ func (m *Scheduler) loadBalancer(ctx context.Context, app *scheduler.App, p *sch
 	return l, nil
 }
 
-func (m *Scheduler) removeLoadBalancer(ctx context.Context, app string, p string) error {
-	l, err := m.findLoadBalancer(ctx, app, p)
+func (m *Scheduler) removeLoadBalancers(ctx context.Context, app string) error {
```
phobologic: Was this called somewhere else before?  I'm not sure how this stops the previous behavior - only that it adds the behavior that it will now make sure that the LBs are destroyed when `Remove` is called?

ejholmes: So, `Remove` is called when the entire app is destroyed. Previously, this was in `RemoveProcess`, which is called to remove an individual process in the app (say for example, you had a `worker1` process, but then you changed its name to `worker`, we remove the irrelevant ecs service).

PR #802, opened 2016-04-27
Add support for creating TCP load balancers in ECS scheduler.

ejholmes:
This should be a pre-req to https://github.com/remind101/empire/pull/800.

This adds support to the ECS scheduler for creating TCP/SSL load balancers for tcp/ssl exposures.


ejholmes:
Might wanna hold off on this until https://github.com/remind101/empire/pull/800/files#r61201001 is addressed.

ejholmes:
Going to address https://github.com/remind101/empire/pull/800/files#r61201001 first.

Issue #770, opened 2016-03-17
EOF errors when making large changes


phobologic:
While scaling all processes down to 0 while doing some maintenance (and then scaling back up after the maintenance), after hanging for a bit, I eventually got an `EOF` error.  My guess is that this is due to the ELB timing out since no data is being sent back to the client.

It also makes it hard to tell which apps were scaled down and which weren't, so you have to check on that and then re-run the truncated command.

```
# emp scale -a acme-inc proc1=1:1X proc2=1:1X proc3=1:2X proc4=1:2X proc5=1:2X proc6=6:1024:1.00gb proc7=20:1X proc8=1:512:6.00gb web=25:PX
error: EOF
```

Also, the next couple of replies get an error from ECS due to limited #'s of deployments.  Not sure there's really anything that can be done about that though, other than wait for the existing deployments to finish:

```
error: InvalidParameterException: 5 active deployments exist, unable to create new deployment.
    status code: 400, request id: ae119f19-eb41-11e5-a4dc-0f01aa337913
```

ejholmes:
I was just thinking about this, and ended up digging into how Go handles TCP keep alives by default. Turns out, any accepted TCP connections from `http.Server` get [TCP keepalives enabled](https://golang.org/src/net/http/server.go?s=64484:64531#L2426), but the [TCP_KEEPINTVL](http://man7.org/linux/man-pages/man7/tcp.7.html) option is set to 3 minutes (we set a 60 second idle connection timeout).

If we increased the idle connection timeout to something like 4 minutes, this might help, although, we'd also need to enable keep alives from the client connection as well.

Issue #684, opened 2015-11-26
Unable to find tasks or services easily in ECS console


gileze33:
We have 20+ services running on empire, and have been having some issues where old deploys hang around for hours. 

I'm trying to work out if it's an issue with ECS or empire, but I can't see any way of finding the service in question without having to go into every single service that's running, open a task definition and then look at which docker image it's related to.

Is there an easier way to quickly link an app or task for an app from the emp command line with the appropriate part in ECS?

crohr:
What I found is that you can `emp info -a service`, take the UUID returned, and use that to filter tasks in the ECS cluster. But it's still a bit clumsy.

benguillet:
Hey @gileze33, tasks definitions and services are named on ECS after the Empire app id and the process type, so all you have to do is get the app id from the deploy hanging with `emp info -a <app>`

rgabo:
@benguillet can you list what you believe are benefits to naming services/task definitions after the Empire app id instead of the a more human-readable naming convention?

The cryptic naming has been killing us when it comes to looking at ELB, ECS and when looking at monitoring/metrics in Datadog and New Relic. It's not that we can build tooling around it, but it is unnecessary complexity.

Thanks!

ejholmes:
Couple of reasons we decided to use the app id instead of the name (which
we used initially):

1. To support renaming. Keying everything off the unique app id means we
have to change less in aws if the app name changes. Granted we don't
support `emp rename` right now, it's still leaves things open for it.
2. We ran into a lot of bugs/issues when deleting an ECS service and
re-creating a new one of the same name (e.g. `emp destroy && emp deploy`).
Mainly the ECS service getting stuck in a drainig state indefinitely.
Always using a unique identifier got around this.

Our hope was that ECS would eventually support tags, but that hasn't
happened yet, so it does add some annoyance. We've felt it too, although it
hasn't been painful enough for us to warrant going back to using app name
(and the migration involved in that).

Eventually, I'd like to use CloudFormation to provision the apps resources
instead of talking to ECS/ELB/Route53 directly, which would solve most of
these issues. No time frame on that though.

rgabo:
thanks for the explanation @ejholmes, we have high hopes for Docker/ECS tags as well.

[...]

Issue #665, opened 2015-10-11
Deregister Task Definitions when app is destroyed


ianneub:
Minor feature request: It would be nice if all the previous Task Definitions were deregistered when an app is destroyed.

As far as I can tell after destroying an app all the old Task Definitions are left lying around.

_Side note: Nice work so far. Thanks for open sourcing this!_

bmarini:
Hey @ianneub, yeah that's a good call. We should be cleaning up those task definitions.

Issue #590, opened 2015-07-17
ELB Healthcheck Intervals are too lax


phobologic:
We use the default HealthCheck when building ELBs, and those are far too lax.  We might want to pull this into the extended procfile support. 

http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/elb-healthchecks.html

ejholmes:
This one got closed an accident from https://github.com/remind101/empire/pull/803, but very easy to change now.

Right now, the cloudformation stack doesn't specify any health checks, which means you can alter the health check by hand after the ELB is created (like before). I think it'd be good to wait for health checks to be defined in the extended Procfile before we add health checks to the CloudFormation stack, so that they can still be configured.

PR #803, opened 2016-04-28
Implement CloudFormation backend.

ejholmes:
Fixes https://github.com/remind101/empire/issues/556
Fixes https://github.com/remind101/empire/issues/684
Probably fixes https://github.com/remind101/empire/issues/665
Probably fixes https://github.com/remind101/empire/issues/770
Can Fix https://github.com/remind101/empire/issues/590 easily
Closes https://github.com/remind101/empire/issues/560
Closes https://github.com/remind101/empire/issues/578
Makes https://github.com/remind101/empire/issues/630 trivial
Makes https://github.com/remind101/empire/issues/706 trivial
Makes https://github.com/remind101/empire/issues/723 trivial
Makes https://github.com/remind101/empire/issues/796 trivial
Makes https://github.com/remind101/empire/issues/797 trivial

WIP so, not everything is implemented (about 80-90%).

The deeper I go into adding the extended Procfile, the more I feel like we're just reimplementing terraform/cloudformation within Empire, which is not what I want to spend my time doing.

This adds support for a CloudFormation backend, so that AWS resources for an app are managed entirely by CloudFormation. We just pass a `scheduler.App` to a template and get a CloudFormation stack that represents the application.

There's massive benefits to doing this:

1. Way quicker to extend with new functionality. We just need to update the stack template, and boom.
2. Resources will be named better. We should prefix the stack with the app name (and possibly an environment if provided) so that ecs services, load balancers, etc are easier to find.
3. Cleaning up after ourselves after destroying an application is super simple. Just delete the stack.
4. We get an audit log of all the stack updates made.

The only disadvantage is we'll need to come up with a simple and safe migration plan (that's easy for other users of Empire as well), since this will require all new ECS/ELB resources for existing apps.

This will make almost everything we do inside Empire vastly simpler as we add features moving forward, so I think it's worth tackling now.

**TODO**

* [x] Name stacks based on the app name and environment.
* [x] Allocate instance ports.
* [x] Implement updating the existing stack.
* [x] Tests
* [x] Upload templates to s3
* [ ] Update aws sdk
* [x] Store stack name in a transaction.
* [x] Implement run.
* [x] Create service role.
* [ ] When deleting, wait for stack to stabilize.
* [x] Prefix stack name.
* [x] Store template based on app name.
* [x] Parameterize scaling.
* [x] Set environment tag.
* [ ] Update CHANGELOG.
* [ ] Auto-subscribe sqs queue to sns queue
* [x] Don't start custom resources server if sqs queue isn't set.
* [x] Error out of cloudformation scheduler is sqs queue isn't set.
* [x] Error out of template if there's any duplicate process names.


ejholmes:
Regarding migration plans, my initial thoughts would be that we just run the old and new backend in parallel, but only use the new backend for newly created apps. Once we're comfortable with it, we should have a mostly (or entirely) automated way to schedule an existing release using the new scheduler, then destroy the old resources using the old scheduler. In psuedo code, something like:

```
func Migrate(release) {
  cloudformationScheduler.Submit(release)
  // Wait 30 minutes or something
  oldECSScheduler.Remove(release.App.ID)
}
```

docs/cloudformation.json
```diff
@@ -301,6 +301,7 @@
             {
               "Effect": "Allow",
               "Action": [
+                "cloudformation:*",
```
ejholmes: When this is ready, need to determine exact permissions that Empire needs (and maybe we can lock this down so it can only access stacks that it created).

scheduler/cloudformation/cloudformation.go
```diff
@@ -1,5 +1,354 @@
...
+	// TODO: Should we just update a parameter in the stack instead?
+	_, err = s.ecs.UpdateService(&ecs.UpdateServiceInput{
+		Cluster:      aws.String(s.Cluster),
+		DesiredCount: aws.Int64(int64(instances)),
+		Service:      aws.String(serviceArn),
+	})
+	return err
+}
+
+func (s *Scheduler) Run(ctx context.Context, app *scheduler.App, process *scheduler.Process, in io.Reader, out io.Writer) error {
```
ejholmes: Still need to implement this, but it will be pretty much identical to the old scheduler.

scheduler/cloudformation/template.go
```diff
@@ -0,0 +1,269 @@
...
+			loadBalancers = append(loadBalancers, map[string]interface{}{
+				"ContainerName": p.Type,
+				"ContainerPort": ContainerPort,
+				"LoadBalancerName": map[string]string{
+					"Ref": loadBalancer,
+				},
+			})
+			resources[loadBalancer] = map[string]interface{}{
+				"Type": "AWS::ElasticLoadBalancing::LoadBalancer",
+				"Properties": map[string]interface{}{
```
ejholmes: Should add some tags to this with app name and process name.

scheduler/cloudformation/cloudformation.go
```diff
@@ -1,5 +1,408 @@
...
+	if err := s.Template.Execute(buf, app); err != nil {
+		return err
+	}
+
+	desc, err := s.cloudformation.DescribeStacks(&cloudformation.DescribeStacksInput{
+		StackName: aws.String(stackName),
+	})
+
+	if err, ok := err.(awserr.Error); ok && err.Message() == fmt.Sprintf("Stack with id %s does not exist", stackName) {
+		if _, err := s.cloudformation.CreateStack(&cloudformation.CreateStackInput{
```
ejholmes: For the initial version of this, we may wanna add a stack policy that disallows removal of  the ELB for any process named "web".

ejholmes: Although it may not be necessary now. For example, when using this backend, when I add a domain to an app and deploy, cloudformation will provision a new ELB and ecs services, then swap DNS over to the new ELB.

scheduler/cloudformation/cloudformation.go
```diff
@@ -1,5 +1,416 @@
...
+
+	tags := []*cloudformation.Tag{
+		{Key: aws.String("empire.app.id"), Value: aws.String(app.ID)},
+		{Key: aws.String("empire.app.name"), Value: aws.String(app.Name)},
+	}
+
+	if err, ok := err.(awserr.Error); ok && err.Message() == fmt.Sprintf("Stack with id %s does not exist", stackName) {
+		if _, err := s.cloudformation.CreateStack(&cloudformation.CreateStackInput{
+			StackName:    aws.String(stackName),
+			TemplateBody: aws.String(buf.String()),
```
ejholmes: Will need to upload to an s3 bucket and use TemlateBodyUrl here since these stacks can get way larger than the 51kb limit for inline template body.

scheduler/cloudformation/template.go
```diff
@@ -0,0 +1,281 @@
...
+			sg := t.InternalSecurityGroupID
+			subnets := t.InternalSubnetIDs
+
+			if p.Exposure.External {
+				scheme = schemeExternal
+				sg = t.ExternalSecurityGroupID
+				subnets = t.ExternalSubnetIDs
+			}
+
+			instancePort := int64(9000) // TODO: Allocate a port
```
ejholmes: One thing that is a little more challenging in a CloudFormation world is instance port allocation. Because we no longer have direct control over creating and deleting load balancers, it's hard to allocate and release ports for them.

I was thinking that we could just add a custom resource with a lambda function that allocates allocates and releases ports, so it's just native CloudFormation (and can be re-used outside of Empire).

ejholmes: Or maybe better yet, use SNS backed custom resources and have Empire listen on an SQS queue to provision port allocations. Keeps everything more self contained.

ejholmes:
Ok, once https://github.com/remind101/empire/pull/811 is merged, this is feature complete and has all the existing functionality of the old scheduler. My be some minor bugs to fix, but this works beautifully in my tests locally.

Once this is merged, I'll work on an automated migration scheduler that will migrate apps from the old scheduler to the new one.

ejholmes:
Also, it's worth mentioning that this is currently only enabled when the `--scheduler` flag is explicitly set to `cloudformation`. The old scheduler will still be the default for the time being.

scheduler/cloudformation/cloudformation.go
```diff
@@ -1,5 +1,554 @@
...
+	buf := new(bytes.Buffer)
+	if err := s.Template.Execute(buf, app); err != nil {
+		return err
+	}
+
+	key := fmt.Sprintf("%x", sha1.Sum(buf.Bytes()))
+
+	_, err = s.s3.PutObject(&s3.PutObjectInput{
+		Bucket:      aws.String(s.Bucket),
+		Key:         aws.String(fmt.Sprintf("/%s", key)),
```
mhahn: since we have app id, i would include that in this key.

i could see it being useful to view different stack "versions" for a specific app.

scheduler/cloudformation/cloudformation.go
```diff
@@ -1,5 +1,554 @@
...
+				return err
+			}
+		}
+	} else if err == nil {
+		stack := desc.Stacks[0]
+		status := *stack.StackStatus
+
+		// If there's currently an update happening, wait for it to
+		// complete.
+		if strings.Contains(status, "IN_PROGRESS") {
```
mhahn: it would be nice to handle the `ROLLBACK_COMPLETE` status too, the only way to move forward with that state is to delete the stack. not sure if you want to handle that deletion or prompt the user to.

scheduler/cloudformation/cloudformation.go
```diff
@@ -1,5 +1,554 @@
...
+
+	resp, err := s.ecs.DescribeTasks(&ecs.DescribeTasksInput{
+		Cluster: aws.String(s.Cluster),
+		Tasks:   arns,
+	})
+
+	return resp.Tasks, err
+}
+
+// Services returns the names of the map that maps the name of a process to the
```
mhahn: this comment is confusing

scheduler/cloudformation/template.go
```diff
@@ -0,0 +1,300 @@
...
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/aws/aws-sdk-go/service/route53"
+	"github.com/remind101/empire/pkg/bytesize"
+	"github.com/remind101/empire/scheduler"
+)
+
+const (
+	// For HTTP/HTTPS/TCP services, we allocate an ELB and map it's instance port to
+	// the container port. This is the port that processes within the container
+	// should bind to. Tihs value is also exposed to the container through the PORT
```
mhahn: typo "Tihs" -> This

scheduler/cloudformation/cloudformation.go
```diff
@@ -1,5 +1,554 @@
...
+	if err != nil {
+		return err
+	}
+
+	serviceArn, ok := services[process]
+	if !ok {
+		return fmt.Errorf("no %s process found", process)
+	}
+
+	// TODO: Should we just update a parameter in the stack instead?
```
mhahn: +1

scheduler/cloudformation/cloudformation.go
```diff
@@ -1,5 +1,554 @@
...
+	_, err := s.ecs.StopTask(&ecs.StopTaskInput{
+		Cluster: aws.String(s.Cluster),
+		Task:    aws.String(instanceID),
+	})
+	return err
+}
+
+// Scale scales the ECS service for the given process to the desired number of
+// instances.
+func (s *Scheduler) Scale(ctx context.Context, appID string, process string, instances uint) error {
```
mhahn: reading through this it seems like this should handle scaling memory and cpu in the template as well?

scheduler/cloudformation/cloudformation.go
ejholmes: Good call. I'll update.

scheduler/cloudformation/cloudformation.go
ejholmes: Yeah, I waffled on that. I figure that that cause should be pretty rare, and it can be handled manually by resolving the issue that caused the rollback, deleting the stack, then re-deploying.

scheduler/cloudformation/cloudformation.go
ejholmes: Cool, I'll make that change in this PR too.

mhahn:
👍 all of this looks awesome

scheduler/cloudformation/cloudformation.go
ejholmes: Done in https://github.com/remind101/empire/pull/803/commits/84f55e5e33a2e7cb4cb4015316bd7a1a6cc003f0. The template is now stored in a folder prefixed with the app name and id.

scheduler/cloudformation/cloudformation.go
ejholmes: Ok, did this in https://github.com/remind101/empire/pull/803/commits/5b7ef2034c23372a33374dbc5e798c9e29539f08. Scale is now controlled through parameters in the stack.

ejholmes:
Ok. Gonna go ahead and merge this in. This won't become the default scheduling backend until there's an easy way to migrate from the old backend to the new one.

My hope is that this becomes the default backend in the next release, with an automated migrator. Then we can remove the old backend, which frees us up to move faster on extended Procfile.

There will probably be some minor bugs to fix in this, but we can deal with those as we see them.

dekz:
Eric this is so incredibly dope. I totally agree with your assessment with the extended Procfile and am glad Empire didn't go down that route.

ejholmes:
@dekz :metal: super excited about this as well. Opens up a lot of possibilities for Empire.

Issue #767, opened 2016-03-17
Add 'commit messages' for things like emp run, emp scale, emp set


phobologic:
It'd be good to have some context when these things happen.

Maybe a flag to make it required or not, and a `-m` option to add it automatically from the command line, or else you get prompted?

I think this is a good set of commands to start with:

- emp run
- emp scale
- emp set

@sanjayprabhu

mhahn:
Discussed a little bit with @phobologic 

Rough decisions for v1:
- empire already supports events, the things we want to track commit messages for are a subset of those events.
- instead of worrying about storage for v1, add support for adding a commit message to those events

For v2, we can add an alternative event backend that also publishes events to some persistent storage (S3 or RDS etc)

mhahn:
v1:
- [ ] emp cli should support `-m`  for a subset of commands (similar to `-a` )
- [ ] emp api needs to be able to handle the commit message for those events

v1.1:
- [ ] should have some flag that allows you to require commit messages for a set of actions for the empire server
- [ ] if the server requires a commit message, it should return an error saying commit message is required

v1.2:
- [ ] the client should handle the error from the server and have some UI for seeing if the user wants to enter the required commit message

ejholmes:
We could store the commit message on the `description` column on releases, but maybe it'd be better to store it in a separate column.

phobologic:
Yeah, the weird part with storing this is the fact that we are looking for commits in other events that don't create releases, like scaling events.

ejholmes:
Ah yeah, good call. Although honestly, maybe those should create new releases? The only reason we didn't was to have the same behavior as Heroku.

sanjayprabhu:
+1 to scaling events creating releases. Would love to be able to see them
in emp releases.

On Wed, Apr 27, 2016, 15:46 Eric Holmes <notifications@github.com> wrote:

> Ah yeah, good call. Although honestly, maybe those should create new
> releases? The only reason we didn't was to have the same behavior as Heroku.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/remind101/empire/issues/767#issuecomment-215254193>
>
-- 
Sanjay

phobologic:
I'm waffling on the idea.  It feels weird in the 12factor world.  I'd almost be more prone to support an 'Events' table and an `emp events` command that let you look at all the old events, whether for a single app or all of an environment.

sanjayprabhu:
[...]

PR #808, opened 2016-04-29
Commit messages

mhahn:
v1 implementation for: https://github.com/remind101/empire/issues/767

- added a message to each event defined in events.go
- went with `OptionalMessage` to enable the server to control which commands require a commit message in the future
- commit message is passed for all relevant actions with the `Commit-Message` header. I didn't use `X-Commit-Message` because apparently that is deprecated? https://tools.ietf.org/html/rfc6648. I don't have a preference and am not very opinionated around REST services.

I added tests around the formatting of messages, but if there are some other areas you think should have some coverage let me know!

### DIFF
--- a/cmd/emp/main.go
+++ b/cmd/emp/main.go
@@ -55,7 +55,7 @@ func (c *Command) PrintLongUsage() {
 }
 
 func (c *Command) FullUsage() string {
-	parts := make([]string, 0, 2)
+	var parts []string
 	if c.NeedsApp || c.OptionalApp {
 		parts = append(parts, "[-a <app or remote>]")
 	}
--- a/cmd/emp/main.go
+++ b/cmd/emp/main.go
@@ -25,11 +25,12 @@ var (
 
 type Command struct {
 	// args does not include the command name
-	Run         func(cmd *Command, args []string)
-	Flag        flag.FlagSet
-	NeedsApp    bool
-	OptionalApp bool
-	NoClient    bool
+	Run             func(cmd *Command, args []string)
+	Flag            flag.FlagSet
+	NeedsApp        bool
+	OptionalApp     bool
+	NoClient        bool
+	OptionalMessage bool
 
 	Usage    string // first word is the command name
 	Category string // i.e. "App", "Account", etc.
@@ -54,8 +55,19 @@ func (c *Command) PrintLongUsage() {
 }
 
 func (c *Command) FullUsage() string {
+	parts := make([]string, 0, 2)
 	if c.NeedsApp || c.OptionalApp {
-		return c.Name() + " [-a <app or remote>]" + strings.TrimPrefix(c.Usage, c.Name())
+		parts = append(parts, "[-a <app or remote>]")
+	}
+
+	if c.OptionalMessage {
+		parts = append(parts, "[-m <message>]")
+	}
+
+	if len(parts) > 0 {
+		// NB: no space between last two placeholders because of
+		// strings.TrimPrefix behavior
+		return fmt.Sprintf("%s %s%s", c.Name(), strings.Join(parts, " "), strings.TrimPrefix(c.Usage, c.Name()))
 	}
 	return c.Usage
 }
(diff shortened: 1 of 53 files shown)


cmd/emp/main.go
```diff
@@ -54,8 +55,19 @@ func (c *Command) PrintLongUsage() {
 }
 
 func (c *Command) FullUsage() string {
+	parts := make([]string, 0, 2)
```
mhahn: i was iffy on whether or not this is the best practice for this time of operation

pkg/heroku/heroku.go
```diff
@@ -62,23 +62,43 @@ type Client struct {
...
-	return c.APIReq(v, "PUT", path, body)
+	return c.APIReq(v, "PUT", path, body, nil)
 }
 
 func (c *Client) Delete(path string) error {
-	return c.APIReq(nil, "DELETE", path, nil)
+	return c.APIReq(nil, "DELETE", path, nil, nil)
+}
+
+func (c *Client) GetWithHeaders(v interface{}, path string, headers http.Header) error {
```
mhahn: i went with the `WithHeaders` suffix to avoid having to go through and add a `nil` parameter to every existing invocation of these functions.

pkg/heroku/heroku.go
```diff
@@ -269,3 +292,15 @@ func (lr *ListRange) SetHeader(req *http.Request) {
...
 }
+
+type RequestHeaders struct {
+	CommitMessage string
+}
+
+func (r *RequestHeaders) Headers() http.Header {
+	headers := http.Header{}
+	if r.CommitMessage != "" {
+		headers.Set("Commit-Message", r.CommitMessage)
```
mhahn: this should be a constant, but wasn't sure if there was an existing best practice for where to put those

server/heroku/heroku.go
```diff
@@ -167,3 +167,7 @@ func UserFromContext(ctx context.Context) *empire.User {
 	}
 	return u
 }
+
+func findMessage(r *http.Request) (string, error) {
```
mhahn: i went with `(string, error)` return value here to support a flag where all commit messages could be required in the future.

pkg/heroku/heroku.go
```diff
@@ -96,7 +116,7 @@ func (c *Client) Delete(path string) error {
 //   nil         no body
 //   io.Reader   body is sent verbatim
 //   else        body is encoded as application/json
-func (c *Client) NewRequest(method, path string, body interface{}) (*http.Request, error) {
+func (c *Client) NewRequest(method, path string, body interface{}, headers http.Header) (*http.Request, error) {
```
ejholmes: IMO, it might be a little cleaner to just do this:

```go
req, err := NewRequest("GET", "/", nil)
if err != nil {
        return err
}
req.Header.Set(..)
```

pkg/heroku/heroku.go
ejholmes: Probably just put in in the top of this file (there's a couple other constants defined there).

cmd/emp/main.go
ejholmes: You can just do this:

```go
var parts []string
```

ejholmes:
Other than those comments, I don't have much to add. I played around with it locally. This is awesome. 👍 

Change you update the CHANGELOG too?

cmd/emp/main.go
mhahn: cool

pkg/heroku/heroku.go
mhahn: ah ok cool

pkg/heroku/heroku.go
mhahn: i was trying to keep it similar to how the heroku client allows global additional headers:

https://github.com/remind101/empire/blob/master/pkg/heroku/heroku.go#L147

I also think your suggestion would work if everything that needed the request header was called from `APIReq`, but `run` accesses `NewRequest` directly:
https://github.com/remind101/empire/blob/master/cmd/emp/run.go#L125

mhahn:
done

PR #814, opened 2016-05-03
Migrate apps to CloudFormation scheduler

ejholmes:
This is my initial thought on how we might migrate to the new CloudFormation based scheduler:

1. All existing apps will continue to use the old scheduler.
2. New apps will use the new scheduler.
3. When you want to migrate an app to the new scheduler, you set an environment variable.

**TODO**

* [x] Delete from `scheduler_migration` when destroyed.
* [x] Docs about upgrading Empire to the CloudFormation backend.
* [x] Add existing apps to scheduler_migration table.

### DIFF
--- a/migrations.go
+++ b/migrations.go
@@ -532,4 +532,17 @@ ALTER TABLE apps ADD COLUMN exposure TEXT NOT NULL default 'private'`,
 			`DROP TABLE stacks`,
 		}),
 	},
+
+	// This migration adds a table that gets used to migrate apps from the
+	// old ECS backend to the shiny new CloudFormation backend.
+	{
+		ID: 17,
+		Up: migrate.Queries([]string{
+			`CREATE TABLE scheduler_migration (app_id text NOT NULL, backend text NOT NULL)`,
+			`INSERT INTO scheduler_migration (app_id, backend) SELECT id, 'ecs' FROM apps`,
+		}),
+		Down: migrate.Queries([]string{
+			`DROP TABLE scheduler_migration`,
+		}),
+	},
 }
--- a/scheduler/cloudformation/migration.go
+++ b/scheduler/cloudformation/migration.go
@@ -1,0 +1,212 @@
+package cloudformation
+
+import (
+	"database/sql"
+	"errors"
+	"fmt"
+	"io"
+
+	"golang.org/x/net/context"
+
+	"github.com/remind101/empire/scheduler"
+	"github.com/remind101/empire/scheduler/ecs"
+)
+
+// This is the environment variable in the application that determines what step
+// of the migration we should transition to. A basic migration flow would look
+// like:
+//
+// 1. `emp set EMPIRE_SCHEDULER_MIGRATION=step1`: CloudFormation stack is
+//    created without any DNS changes.
+// 2. User removes the old CNAME manually in the AWS Console, then sets the
+//    `DNS` parameter in the CloudFormation stack to `true`.
+// 3. `emp set EMPIRE_SCHEDULER_MIGRATION=step2`: The old AWS resources are
+//    removed.
+// 4. `emp unset EMPIRE_SCHEDULER_MIGRATION`: All done.
+const MigrationEnvVar = "EMPIRE_SCHEDULER_MIGRATION"
+
+// ErrMigrating is returned when the application is being migrated.
+var ErrMigrating = errors.New("app is currently being migrated to a CloudFormation stack. Sit tight...")
+
+// This is a scheduler.Scheduler implementation that wraps the newer
+// cloudformation.Scheduler and the older ecs.Scheduler to migrate applications
+// over the the new CloudFormation based scheduler.
+//
+// It uses a sql table to determine what scheduling backend should be used. Apps
+// can be migrated from the ecs scheduler to the cloudformation scheduler by
+// using the Migrate function.
+type MigrationScheduler struct {
+	// The scheduler that we want to migrate to.
+	cloudformation interface {
+		scheduler.Scheduler
+		SubmitWithOptions(context.Context, *scheduler.App, SubmitOptions) error
+	}
+
+	// The scheduler we're migrating from.
+	ecs interface {
+		scheduler.Scheduler
+		RemoveWithOptions(context.Context, string, ecs.RemoveOptions) error
+	}
+
+	db *sql.DB
+}
+
+// NewMigrationScheduler returns a new MigrationSchedeuler instance.
+func NewMigrationScheduler(db *sql.DB, c *Scheduler, e *ecs.Scheduler) *MigrationScheduler {
+	return &MigrationScheduler{
+		db:             db,
+		cloudformation: c,
+		ecs:            e,
+	}
+}
+
+// Backend returns the scheduling backend to use for the given app.
+func (s *MigrationScheduler) Backend(appID string) (scheduler.Scheduler, error) {
+	backend, err := s.backend(appID)
+	if err != nil {
+		return nil, err
+	}
+
+	switch backend {
+	case "ecs":
+		return s.ecs, nil
+	case "cloudformation":
+		return s.cloudformation, nil
+	default:
+		return nil, ErrMigrating
+	}
+}
+
+// backend returns the name of the backend to use for operations.
+func (s *MigrationScheduler) backend(appID string) (string, error) {
+	var backend string
+	err := s.db.QueryRow(`SELECT backend FROM scheduler_migration WHERE app_id = $1`, appID).Scan(&backend)
+
+	// For newly created apps.
+	if err == sql.ErrNoRows {
+		return "cloudformation", nil
+	}
+
+	return backend, err
+}
+
+func (s *MigrationScheduler) Submit(ctx context.Context, app *scheduler.App) error {
+	state, err := s.backend(app.ID)
+	if err != nil {
+		return err
+	}
+
+	desiredState := app.Processes[0].Env[MigrationEnvVar]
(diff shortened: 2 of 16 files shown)


scheduler/cloudformation/migration.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return err
+	}
+	return b.Submit(ctx, app)
+}
+
+// Migrate submits the app to the CloudFormation scheduler, waits for the stack
+// to successfully create, then removes the old API managed resources using the
+// ECS scheduler.
+func (s *MigrationScheduler) Migrate(ctx context.Context, app *scheduler.App) error {
```
ejholmes: This is where the magic happens. The ordering is less than ideal, since it will inevitably incur some downtime.

Ideally, this would:

1. Create the CloudFormation stack.
2. Wait for it to update.
3. Swap DNS over to the new stack.
4. Once you've verified things are working, remove the old resources.

ejholmes: Ok, a little better now. It will create the stack without any dns changes, destroy the old resources, then update the new stack with the CNAME.

scheduler/cloudformation/migration.go
```diff
@@ -0,0 +1,172 @@
...
+		Wait: true,
+	}); err != nil {
+		return err
+	}
+
+	_, err := s.db.Exec(`UPDATE scheduler_migration SET backend = 'cloudformation' WHERE app_id = $1`, app.ID)
+	return err
+}
+
+func (s *MigrationScheduler) Remove(ctx context.Context, appID string) error {
```
ejholmes: This should also remove the row for the app from `scheduler_migration`.

migrations.go
```diff
@@ -532,4 +532,17 @@ ALTER TABLE apps ADD COLUMN exposure TEXT NOT NULL default 'private'`,
 			`DROP TABLE stacks`,
 		}),
 	},
+
+	// This migration changes that way we store process configuration for
```
mhahn: old comment?

ejholmes: Bah. Copy paste.

scheduler/cloudformation/migration.go
```diff
@@ -0,0 +1,172 @@
...
+// cloudformation.Scheduler and the older ecs.Scheduler to migrate applications
+// over the the new CloudFormation based scheduler.
+//
+// It uses a sql table to determine what scheduling backend should be used. Apps
+// can be migrated from the ecs scheduler to the cloudformation scheduler by
+// using the Migrate function.
+type MigrationScheduler struct {
+	// This is the amount of time to wait for the new CloudFormation stack
+	// to stabilize before swapping over to it. 5-10 minutes is probably
+	// enough for the load balancer to starting being resolvable.
```
mhahn: s/starting/start

scheduler/cloudformation/migration.go
```diff
@@ -0,0 +1,172 @@
...
+	// Don't make any DNS changes.
+	if err := s.cloudformation.SubmitWithOptions(ctx, app, SubmitOptions{
+		Wait:  true,
+		NoDNS: true,
+	}); err != nil {
+		return err
+	}
+
+	// Wait for some time for everything to start running.
+	<-time.After(s.WaitTime)
```
mhahn: is there a way we can make this a health check function instead?

maybe supplying it with a randomish DNS value and checking against that until we get a success response?

scheduler/cloudformation/migration.go
```diff
@@ -0,0 +1,172 @@
...
+
+	// Remove the old AWS resources.
+	if err := s.ecs.Remove(ctx, app.ID); err != nil {
```
mhahn: why not swap the DNS before killing the old AWS resources?

mhahn: i guess because `ecs.Remove` removes the existing DNS record?

ejholmes: Yeah, right now that's hidden, but we could expose that on the ecs scheduler. Might be worth it.

I was thinking, we could also make the migration a 2-3 step process:

1. The new stack is created, and DNS is swapped over to it.
2. DNS is swapped.
3. The old resources are removed.

For high load services, this might be ideal, because it gives you time to pre-warm ELB's.

Moving through each step could also potentially be controlled by just setting environment variables with `emp set`, like:

1. `emp set EMPIRE_SCHEDULER_MIGRATION=step1`.
  * Wait for the stack to be created.
  * Ask AWS to pre-warm the ELB.
2. `emp set EMPIRE_SCHEDULER_MIGRATION=step2`.
  * DNS is swapped to the new stack.
  * Check that everything is functioning as expected.
3. `emp set EMPIRE_SCHEDULER_MIGRATION=step3`.
  * Old resources are removed.

mhahn:
The process of switching over new apps to cloudformation looks great.

For migrating an existing app, my ideal would be:
- bring up new stacks
- test they're working (maybe with some temporary dns record?)
- swap dns record from old stacks to new stacks
- check to make sure everything is working again
- remove old stacks

my hesitation around the 10 minute wait time as well as waiting to remove the old app before bringing up the new one is just issues i've had previously where you can't delete a resource because something else depends on it or bringing up a stack takes longer than you expect. i don't think the removal of something with a dependency is an issue if you assume empire has managed the whole process, but that would be my worst case scenario thinking.

ejholmes:
Yep, agreed. I think making it a 3 step manual process like that would be ideal.

scheduler/cloudformation/migration.go
mhahn: are you thinking of adding a `migrate` cmd?

it seems like that could be useful to handle these migrations, ie. when i'm ready to migrate:

```
emp migrate -a acme-inc
```

and then you could control those flow steps through that.

i like the idea of doing the swap and then allowing the user to confirm everything is working properly, rollback if necessary, or continue forward.

i'm wary of the auto destruction of resources. there is always some edge case that doesn't get handled and then that user is hosed.

ejholmes: I just added a commit that I think addresses all of the concerns, and makes it easy to move forward safely. Basically, these are the steps you would run:

1. `emp set EMPIRE_SCHEDULER_MIGRATION=step1`: The new CloudFormation stack is created, with the `DNS` parameter set to `false`.
2. At this point, you would manually remove the old `<app>.empire` CNAME, and then update the `DNS` parameter in the CloudFormation stack to `true`. I originally thought of automating this as another step, but I think it just ends up being easier/safer to be able to make the DNS changes manually, which makes it easy to rollback if there's problems.
3. `emp set EMPIRE_SCHEDULER_MIGRATION=step2`: This removes the old AWS resources.
4. `emp unset EMPIRE_SCHEDULER_MIGRATION`: Finalizes the migration, and unlocks the app.

ejholmes:
Ok, I added some tests for the different migration steps and scenarios and addressed all the comments. I'll probably hold off on merging this until I have a chance to test migrating a couple of apps in our staging environment to see if any issues pop up.

scheduler/cloudformation/migration.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+}
+
+// backend returns the name of the backend to use for operations.
+func (s *MigrationScheduler) backend(appID string) (backend string, err error) {
+	err = s.db.QueryRow(`SELECT backend FROM scheduler_migration WHERE app_id = $1`, appID).Scan(&backend)
+
+	// For newly created apps.
+	if err == sql.ErrNoRows {
+		backend = "cloudformation"
```
mhahn: curious when you decide to use named returns vs not

mhahn:
👍

scheduler/cloudformation/migration.go
ejholmes: Actually this is a named return only because of an existing change I had in here, where the return value was `(string, string, error)`. Generally only used named returns to remove ambiguity in the return values.

ejholmes: And fixed.

ejholmes:
Also, been thinking about how the scheduler changes will play out in new releases, and when to deprecate the old scheduler. My initial plan is in https://github.com/remind101/empire/wiki/Scheduler-Deprecation-Plan.

ejholmes:
Discovered an issue with this when moving one of our larger apps that generates a (very) large CloudFormation template (~830kb), which goes ways over the 460kb limit when using s3 templates: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html

So, this means for large stacks, we'll need to generate multiple templates, then combine them into a single stack using nested templates. I think that should be pretty easy to do, and shouldn't have too many downsides. We can just generate a separate stack file for each process, then submit a really small template to CloudFormation.

ejholmes:
Actually, nested stacks may not work well since they actually create separate cloudformation stacks, which complicates how we return the running tasks for an app. Not impossible if we have to go down that route, but then it makes it easier to hit the cloudformation stack limit.

This is a pretty exception case, so maybe there's something that we can do to reduce the size of templates. Most of it is duplication in task definitions anyway.

ejholmes:
Alright, the vast majority of the size was just because we were using json.MarshalIndent. Getting rid of that cuts the file size in about half. May be some more that can be done to minify templates further, but for the 99%, this will be fine I think.

ejholmes:
Ok, used this to migrate 7 applications in our staging environment, and I'm pretty happy with the flow. For future reference, there's a [Scheduler Migration Guide](https://github.com/remind101/empire/wiki/Scheduler-Migration-Guide) in the wiki that walks through the steps that this implements.

PR #815, opened 2016-05-03
Support specifying a flag to require messages for all commands

mhahn:
I initially thought this would be a good candidate for middleware, but after going down that path it really didn't make any sense. We would have to white list paths we didn't want to have the message required for (ie. auth etc.) and I already have `findMessage` implemented in all the actions where a message should be required.

I'll add support for gracefully handling the error within emp.

### DIFF
--- a/empire.go
+++ b/empire.go
@@ -155,6 +155,11 @@ func (opts CreateOpts) Event() CreateEvent {
 
 // Create creates a new app.
 func (e *Empire) Create(ctx context.Context, opts CreateOpts) (*App, error) {
+
+	if e.MessagesRequired && opts.Message == "" {
+		return nil, &MessageRequiredError{}
+	}
+
 	a, err := appsCreate(e.db, &App{Name: opts.Name})
 	if err != nil {
 		return a, err
@@ -185,6 +190,11 @@ func (opts DestroyOpts) Event() DestroyEvent {
 
 // Destroy destroys an app.
 func (e *Empire) Destroy(ctx context.Context, opts DestroyOpts) error {
+
+	if e.MessagesRequired && opts.Message == "" {
+		return &MessageRequiredError{}
+	}
+
 	tx := e.db.Begin()
 
 	if err := e.apps.Destroy(ctx, tx, opts.App); err != nil {
@@ -249,6 +259,11 @@ func (opts SetOpts) Event() SetEvent {
 // Config. If the app has a running release, a new release will be created and
 // run.
 func (e *Empire) Set(ctx context.Context, opts SetOpts) (*Config, error) {
+
+	if e.MessagesRequired && opts.Message == "" {
+		return nil, &MessageRequiredError{}
+	}
+
 	tx := e.db.Begin()
 
 	c, err := e.configs.Set(ctx, tx, opts)
@@ -340,6 +355,11 @@ func (opts RestartOpts) Event() RestartEvent {
 // Restart restarts processes matching the given prefix for the given Release.
 // If the prefix is empty, it will match all processes for the release.
 func (e *Empire) Restart(ctx context.Context, opts RestartOpts) error {
+
+	if e.MessagesRequired && opts.Message == "" {
+		return &MessageRequiredError{}
+	}
+
 	if err := e.apps.Restart(ctx, e.db, opts); err != nil {
 		return err
 	}
@@ -464,6 +484,11 @@ func (opts RollbackOpts) Event() RollbackEvent {
 // Rollback rolls an app back to a specific release version. Returns a
 // new release.
 func (e *Empire) Rollback(ctx context.Context, opts RollbackOpts) (*Release, error) {
+
+	if e.MessagesRequired && opts.Message == "" {
+		return nil, &MessageRequiredError{}
+	}
+
 	tx := e.db.Begin()
 
 	r, err := e.releases.Rollback(ctx, tx, opts)
@@ -517,6 +542,11 @@ func (opts DeploymentsCreateOpts) Event() DeployEvent {
 
 // Deploy deploys an image and streams the output to w.
 func (e *Empire) Deploy(ctx context.Context, opts DeploymentsCreateOpts) (*Release, error) {
+
+	if e.MessagesRequired && opts.Message == "" {
+		return nil, &MessageRequiredError{}
+	}
+
 	tx := e.db.Begin()
 
 	r, err := e.deployer.Deploy(ctx, tx, opts)
@@ -578,6 +608,11 @@ func (opts ScaleOpts) Event() ScaleEvent {
 
 // Scale scales an apps process.
 func (e *Empire) Scale(ctx context.Context, opts ScaleOpts) (*Process, error) {
+
+	if e.MessagesRequired && opts.Message == "" {
+		return nil, &MessageRequiredError{}
+	}
+
 	tx := e.db.Begin()
 
 	p, err := e.apps.Scale(ctx, tx, opts)
--- a/empire.go
+++ b/empire.go
@@ -108,11 +108,6 @@ func New(db *DB, options Options) *Empire {
 	return e
 }
 
-// ShouldRequireMessages returns a boolean if messages are required for actions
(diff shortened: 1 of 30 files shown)


server/heroku/heroku.go
```diff
@@ -169,6 +170,15 @@ func UserFromContext(ctx context.Context) *empire.User {
...
-	return r.Header.Get(heroku.CommitMessageHeader), nil
+type shouldRequireMessages interface {
+	ShouldRequireMessages() bool
+}
+
+func findMessage(r *http.Request, e shouldRequireMessages) (string, error) {
+	var err error
+	h := r.Header.Get(heroku.CommitMessageHeader)
+	if e.ShouldRequireMessages() && h == "" {
+		err = fmt.Errorf("Header '%s' is required", heroku.CommitMessageHeader)
```
mhahn: I was curious if this was the best way to return an error. I'll handle this within emp and with the current error just search for the string "CommitMessage" and required or something, but wasn't sure if there was a better way to do this.

mhahn:
v1.1 of https://github.com/remind101/empire/issues/767

pkg/heroku/heroku.go
```diff
@@ -250,14 +250,18 @@ type errorResp struct {
...
 	if res.StatusCode/100 != 2 { // 200, 201, 202, etc
 		var e errorResp
 		err := json.NewDecoder(res.Body).Decode(&e)
 		if err != nil {
 			return errors.New("Unexpected error: " + res.Status)
 		}
-		return Error{error: errors.New(e.Message), Id: e.Id, URL: e.URL}
+		ret := Error{error: errors.New(e.Message), Id: e.Id, URL: e.URL}
+		if e.Id == "message_required" {
+			panic(ret)
```
ejholmes: Kinda sucks that we have to panic here. But I can't think of a much better way to do this check globally since the run actions don't return an error (although, that would probably be a good refactor later).

empire.go
```diff
@@ -79,6 +79,9 @@ type Empire struct {
 
 	// RunRecorder is used to record the logs from interactive runs.
 	RunRecorder RunRecorder
+
+	// MessagesRequired is a boolean used to determine if messages should be required for events.
+	MessagesRequired bool
```
ejholmes: I wonder if this should actually cause methods on `empire.Empire` to return something like an `empire.ErrMessageRequired`. It would end up being annoying boilerplate, because you'd have to add an `if e.MessagesRequired { return ErrMessageRequired }` to everything, but it forces consumers of the `empire` package to provide a message.

Or, if we really only want to require it from the heroku server, maybe just move this to `server/heroku` instead.

ejholmes:
Awesome. Other than those comments, this looks good to me. Also, CHANGELOG :).

pkg/heroku/heroku.go
mhahn: yea, that was my initial thought, but they `printFatal` everywhere.

empire.go
mhahn: i would think requiring commit messages would be something that empire requires and the heroku client just knows how to handle. i'll take a look at this tomorrow.

mhahn:
@ejholmes addressed the comments and implemented the `HijackReadWriter`.

tested all the commands and everything looks to be working as expected

server/heroku/processes.go
```diff
@@ -100,23 +99,19 @@ func (h *PostProcess) ServeHTTPContext(ctx context.Context, w http.ResponseWrite
...
+			Header:   "HTTP/1.1 200 OK\r\nContent-Type: application/vnd.empire.raw-stream\r\n\r\n",
 		}
-		defer closeStreams(inStream, outStream)
-
-		fmt.Fprintf(outStream, "HTTP/1.1 200 OK\r\nContent-Type: application/vnd.empire.raw-stream\r\n\r\n")
-
+		defer stream.Close()
 		// Prevent the ELB idle connection timeout to close the connection.
-		defer close(streamhttp.Heartbeat(outStream, 10*time.Second))
+		defer close(streamhttp.Heartbeat(stream, 10*time.Second))
```
ejholmes: I wonder if the heartbeat could be a problem, since this will cause the connection to be hijacked.

server/heroku/processes.go
```diff
@@ -100,23 +99,19 @@ func (h *PostProcess) ServeHTTPContext(ctx context.Context, w http.ResponseWrite
 	}
 
 	if form.Attach {
-		inStream, outStream, err := hijackServer(w)
-		if err != nil {
-			return err
+		stream := &hijack.HijackReadWriter{
+			Response: w,
+			Header:   "HTTP/1.1 200 OK\r\nContent-Type: application/vnd.empire.raw-stream\r\n\r\n",
```
ejholmes: Super minor, but this might be nice if we just passed an `http.Header` here and did the rest of the formatting within HijackReadWriter.

server/heroku/processes.go
ejholmes: I think in reality, we'd almost never hit that case, because errors returned from Run would rarely take longer than 10 seconds.

empire.go
```diff
@@ -386,6 +409,10 @@ func (opts RunOpts) Event() RunEvent {
 func (e *Empire) Run(ctx context.Context, opts RunOpts) error {
 	event := opts.Event()
 
+	if e.MessagesRequired && opts.Message == "" {
```
ejholmes: Maybe pull this into a little helper function:

```go
if err := e.requiresMessage(opts.Message); err != nil {
        return err
}
```

ejholmes:
This is great 👍

server/heroku/processes.go
ejholmes: I guess, we could special case HijackServer to not hijack if a single null character is written. Or add an option to streamhttp.Heartbeat to only send heartbeats after an initial call to Write has happened.

mhahn: i don't think we can avoid cases where we've already hijacked the connection and have started writing to it and then an error is returned from Run.

An example of this is if you try and run something that doesn't exist on the container:

```
$ ./build/emp run -a acme-inc worker -m test
Pulling from remind101/acme-inc
Digest: sha256:14d3e8043b5aae9fc730ae95774be993af5a0f7a16ef4852bde934b3da71fe0e
runner: start container: API error (500): Cannot start container 611e31671e0eb330a221adecc872935594c653fecfb78d445872039ad6a35f17: [8] System error: exec: "worker": executable file not found in $PATH
```

https://github.com/remind101/empire/pull/815/commits/7f751b44ed3efcc57f0b9d35ace44d0a5951be0b falls back to the existing behavior to just write the error to the stream and return `nil` if the connection has already been hijacked.

Given that, I'm not sure it's worth worry about the heartbeat and just allowing it to write the error to the stream.

mhahn:
ok, i think this is good to go.

one other minor thing i fixed was when an error was returned previously you would get:

```
[mhahn:empire(interactive-commit-messages)]$ ./build/emp run -a acme-inc worker -m test
Pulling from remind101/acme-inc
Digest: sha256:14d3e8043b5aae9fc730ae95774be993af5a0f7a16ef4852bde934b3da71fe0e
runner: start container: API error (500): Cannot start container e2a890f3fd4ac30bf9580d3f87f46c7f575ce033bb859d49a9b99ab8dde15e25: [8] System error: exec: "worker": executable file not found in $PATH
                                                                                 [mhahn:empire(master)]$
```

we're now adding an extra carriage return to the output and it cleans it up:

```
[mhahn:empire(interactive-commit-messages)]$ ./build/emp run -a acme-inc worker -m test
Pulling from remind101/acme-inc
Digest: sha256:14d3e8043b5aae9fc730ae95774be993af5a0f7a16ef4852bde934b3da71fe0e
runner: start container: API error (500): Cannot start container cd18fcb9d2a9146293473a16bf983e6169a32c52f758094d0b691b8e64efa16f: [8] System error: exec: "worker": executable file not found in $PATH
[mhahn:empire(interactive-commit-messages)]$
```

ejholmes:
:+1:

Issue #795, opened 2016-04-22
Errors returned from interactive `emp run`'s are ugly


ejholmes:
```console
$ emp run bash -a does-not-exist-app
{"id":"not_found","message":"Request failed, the specified resource does not exist","url":""}
```

ejholmes:
For the record, this happens because we don't check the Content-Type header [here](https://github.com/remind101/empire/blob/master/cmd/emp/run.go#L148).

[ClientConn.Do](https://golang.org/pkg/net/http/httputil/#ClientConn.Do) returns an *http.Response. If the `Content-Type` header is `applciation/json` then we should just do normal decoding. If it's `application/vnd.empire.raw-stream`, then we should do the io.Copy.

ejholmes:
The other bug here is that, when this happens the command gets "stuck" and never returns. You can find the process and kill it (`killall emp`), but then your terminal is stuck because apparently the [defer to reset the terminal](https://github.com/remind101/empire/blob/master/cmd/emp/run.go#L160) never gets run. Then you have to `$ reset` the terminal.

/cc @bmarini

ejholmes:
Fixed in https://github.com/remind101/empire/pull/816

PR #816, opened 2016-05-03
Handle errors when trying to do an interactive emp run

mhahn:
Addresses https://github.com/remind101/empire/issues/795

The headers that come back from the request are:
```
map[Date:[Tue, 03 May 2016 21:04:53 GMT] Content-Length:[67] Content-Type:[text/plain; charset=utf-8]]
```

Since this is still a "heroku" request, it seemed fitting to just export the `CheckResp` method and call that here.


ejholmes:
Amazing. This looks great. Can you update the CHANGELOG too?

👍

PR #811, opened 2016-04-30
Instance port allocation for CloudFormation backend

ejholmes:
Addresses https://github.com/remind101/empire/pull/803#discussion_r61543193

This adds instance port allocation to the CloudFormation scheduling backend by using CloudFormation custom resources.

In a nutshell, when we provision a CloudFormation stack for an app, every container port in a task definition gets a `Custom::InstancePort` resource (e.g. `web8080InstancePort`). CloudFormation sends the provisioning request to an SNS topic, and Empire pulls these off of an SQS queue and provisions the ports (or releases them) using the existing port allocator.

This also sets up a nice little framework for adding [additional custom resources](https://github.com/remind101/empire/issues/810).

**TODO**

* [ ] Auto-subscribe sqs queue to sns queue
* [ ] Don't start custom resources server if sqs queue isn't set.
* [ ] Error out of cloudformation scheduler is sqs queue isn't set.

PR #818, opened 2016-05-05
Auto subscribe CustomResourcesQueue in dev stack.

ejholmes:
This is a follow up to https://github.com/remind101/empire/pull/811, which auto subscribes the CustomResourcesQueue to the CustomResourcesTopic in the demo/dev stack.


mhahn:
👍

Issue #810, opened 2016-04-29
Custom CloudFormation resources


ejholmes:
https://github.com/remind101/empire/pull/803 got me thinking, we could have Empire support provisioning of custom resources so you can integrate Empire into CloudFormation.

For example, we could have these custom resources:

* `Custom::App`

Then, you integrate Empire into your CloudFormation stack for provisioning resources (e.g. you create an RDS database, and you want to add it as the `DATABASE_URL` to the `api` app).

```json
{
  "AWSTemplateFormatVersion": "2010-09-09",
  "Description": "Custom::App test",

  "Parameters": {
    "CustomResourcesTopic": {
      "Type": "String",
      "Description": "The ARN of the SNS queue to create Empire Custom Resources."
    }
  },

  "Resources": {
    "apiBucket": {
      "Type": "AWS::S3::Bucket"
    },
    "api": {
      "Type": "Custom::App",
      "Properties": {
        "ServiceToken": { "Ref": "CustomResourcesTopic" },
        "Name": "api",
        "Environment": {
          "BUCKET": { "Ref": "apiBucket" }
        }
      }
    }
  }
}
```

ejholmes:
This would bring Empire up to the level of doing something like https://segment.com/blog/rebuilding-our-infrastructure/, where the entire infrastructure can be managed with CloudFormation, but still provide simple operations for doing things like migrations and scaling.

ejholmes:
Just thinking through UX for something like this, if an environment variable was set by CloudFormation, then you probably wouldn't want it to be visible and/or not allow it to be updated through the api.

ejholmes:
[Here's a commit](https://github.com/remind101/empire/commit/b1dac8dd792a4a66b732907905bb9ebe646958f7) that adds a `Custom::App` resource on top of https://github.com/remind101/empire/pull/811, which allows you to create apps, and manage environment variables through cloudformation.

You can use the example stack in the description to create an "api" app that as a `BUCKET` environment variable set to a CloudFormation generated S3 bucket.

ejholmes:
There are now `Custom::EmpireApp` and `Custom::EmpireAppEnvironment` resources in the [0.11.0 release](https://github.com/remind101/empire/releases/tag/v0.11.0).

PR #819, opened 2016-05-05
Add Custom::App CloudFormation resource.

ejholmes:
Closes https://github.com/remind101/empire/issues/810

This adds a `Custom::App` resource, so you can provision Empire applications, and set environment variables from CloudFormation stacks. This allows you to provision your entire infrastructure, including Empire apps, and link it all together.

Not quite ready, because I think there's some UX things to address first, but I'm pretty excited about what this enables.

**TODO**

* [ ] Environment variables set by CloudFormation probably shouldn't be editable (or visible) from `emp set`.
* [ ] Docs
* [ ] Can we use `Custom::Empire::App`?

### DIFF
--- a/server/cloudformation/apps.go
+++ b/server/cloudformation/apps.go
@@ -1,0 +1,116 @@
+package cloudformation
+
+import (
+	"github.com/jinzhu/gorm"
+	"github.com/remind101/empire"
+	"golang.org/x/net/context"
+)
+
+type AppsProvisioner struct {
+	empire *empire.Empire
+}
+
+func (p *AppsProvisioner) Provision(req Request) (id string, data interface{}, err error) {
+	ctx := context.Background()
+	user := &empire.User{Name: "cloudformation"}
+
+	switch req.RequestType {
+	case Create:
+		name := req.ResourceProperties["Name"].(string)
+		app, err := p.empire.AppsFind(empire.AppsQuery{
+			Name: &name,
+		})
+		if err != nil && err != gorm.RecordNotFound {
+			return "", nil, err
+		}
+
+		app, err = p.empire.Create(ctx, empire.CreateOpts{
+			User: user,
+			Name: name,
+		})
+		if err != nil {
+			return "", nil, err
+		}
+
+		if err := p.setEnvironment(ctx, user, app, req); err != nil {
+			return "", nil, err
+		}
+
+		return app.ID, map[string]string{"Id": app.ID}, nil
+	case Delete:
+		id := req.PhysicalResourceId
+		app, err := p.empire.AppsFind(empire.AppsQuery{
+			ID: &id,
+		})
+		if err != nil {
+			return id, nil, err
+		}
+
+		// TODO: Handle error
+		_ = p.empire.Destroy(ctx, empire.DestroyOpts{
+			User: user,
+			App:  app,
+		})
+
+		return id, nil, nil
+	case Update:
+		id := req.PhysicalResourceId
+
+		app, err := p.empire.AppsFind(empire.AppsQuery{
+			ID: &id,
+		})
+		if err != nil {
+			return id, nil, err
+		}
+
+		if err := p.setEnvironment(ctx, user, app, req); err != nil {
+			return id, nil, err
+		}
+
+		return id, map[string]string{"Id": app.ID}, nil
+	}
+
+	return
+}
+
+func (p *AppsProvisioner) setEnvironment(ctx context.Context, user *empire.User, app *empire.App, req Request) error {
+	vars := varsFromRequest(req)
+
+	_, err := p.empire.Set(ctx, empire.SetOpts{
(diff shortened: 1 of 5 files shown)


server/cloudformation/apps.go
```diff
@@ -0,0 +1,116 @@
...
+	"golang.org/x/net/context"
+)
+
+type AppsProvisioner struct {
+	empire *empire.Empire
+}
+
+func (p *AppsProvisioner) Provision(req Request) (id string, data interface{}, err error) {
+	ctx := context.Background()
+	user := &empire.User{Name: "cloudformation"}
```
ejholmes: Cool thing is you'll still get even notifications when CloudFormation sets environment variables.

server/cloudformation/apps.go
```diff
@@ -0,0 +1,116 @@
...
+	case Create:
+		name := req.ResourceProperties["Name"].(string)
+		app, err := p.empire.AppsFind(empire.AppsQuery{
+			Name: &name,
+		})
+		if err != nil && err != gorm.RecordNotFound {
+			return "", nil, err
+		}
+
+		app, err = p.empire.Create(ctx, empire.CreateOpts{
```
ejholmes: This needs a way to link an existing application.

server/cloudformation/cloudformation.go
```diff
@@ -86,11 +86,11 @@ type Request struct {
 	// This field contains the contents of the Properties object sent by the
 	// template developer. Its contents are defined by the custom resource
 	// provider.
-	ResourceProperties interface{} `json:"ResourceProperties"`
+	ResourceProperties map[string]interface{} `json:"ResourceProperties"`
```
ejholmes: Maybe easier to make this a json.RawMessage so resource providers can just decode into a struct.

nzoschke:
This is really neat :)

ejholmes:
Thinking about this a bit, I think you should be able to control whether an environment variable is secret (hidden from `emp env`) and whether an environment variable is locked by the context that set it (e.g. stack id).

There's certain things that we'd want to make both locked and secret (like DATABASE_URL) and certain things that we might want to just set as sensible defaults when a new environment is brought up.

server/cloudformation/cloudformation.go
```diff
@@ -187,11 +187,14 @@ type CustomResourceProvisioner struct {
...
 // sqs client configured from config.
-func NewCustomResourceProvisioner(db *sql.DB, config client.ConfigProvider) *CustomResourceProvisioner {
+func NewCustomResourceProvisioner(empire *empire.Empire, config client.ConfigProvider) *CustomResourceProvisioner {
 	return &CustomResourceProvisioner{
 		Provisioners: map[string]Provisioner{
 			"Custom::InstancePort": &InstancePortsProvisioner{
-				ports: lb.NewDBPortAllocator(db),
+				ports: lb.NewDBPortAllocator(empire.DB.DB.DB()),
+			},
+			"Custom::App": &AppsProvisioner{
```
ejholmes: Should check if it's possible to use `Custom::Empire::App` here. Not sure.

server/cloudformation/apps.go
mhahn: this should account for empire requiring messages, maybe just supplying a default one?

server/cloudformation/apps.go
```diff
@@ -0,0 +1,116 @@
...
+	return
+}
+
+func (p *AppsProvisioner) setEnvironment(ctx context.Context, user *empire.User, app *empire.App, req Request) error {
+	vars := varsFromRequest(req)
+
+	_, err := p.empire.Set(ctx, empire.SetOpts{
+		User: user,
+		App:  app,
+		Vars: vars,
```
mhahn: would be cool to pass a message here describing which CF event set the variables, ie. update or create

mhahn:
i can't get over how cool this is going to be. 95% of the variables set within my empire apps are from cloudformation via stacker. being able to define everything within cloudformation is going to be life changing.

👍 👍 💯

server/cloudformation/cloudformation.go
mhahn: doesn't sound like it: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cfn-customresource.html#aws-cfn-resource-type-name

ejholmes: Yeah. I tried it but didn't work. Maybe just make this `Custom::EmpireApp`.

server/cloudformation/apps.go
ejholmes: Good call. I'll update this with messages.

mhahn:
https://github.com/remind101/empire/pull/855 should be used instead

ejholmes:
Closing in favor of https://github.com/remind101/empire/pull/855

PR #821, opened 2016-05-07
Avoid DescribeStackResource call

ejholmes:
I discovered that for apps that have a lot of processes, you would almost always get throttled when calling DescribeStackResource (it's N+1 with the number processes). Previously, we used this call to map the process name, to the ARN of the ECS service.

To avoid hitting the rate limit, we can instead just use an output to do the mapping, which means better performance, and no rate limiting. The only downside is, you have to wait for the output to get populated before you can view the running tasks, but that's a minor annoyance.

### DIFF
--- a/scheduler/cloudformation/template.go
+++ b/scheduler/cloudformation/template.go
@@ -141,6 +141,8 @@ func (t *EmpireTemplate) Build(app *scheduler.App) (interface{}, error) {
 	resources := map[string]interface{}{}
 	outputs := map[string]interface{}{}
 
+	serviceMappings := []map[string]interface{}{}
+
 	for _, p := range app.Processes {
 		cd := t.ContainerDefinition(p)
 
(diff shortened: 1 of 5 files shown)


ejholmes:
Tested on an app with 40+ processes and this works much better (and faster).

scheduler/cloudformation/template.go
```diff
@@ -305,15 +313,21 @@ func (t *EmpireTemplate) Build(app *scheduler.App) (interface{}, error) {
...
-			"Metadata": &serviceMetadata{
-				Name: p.Type,
-			},
+			"Type":       "AWS::ECS::Service",
 			"Properties": serviceProperties,
 		}
 
 	}
 
+	outputs[servicesOutput] = map[string]interface{}{
```
phobologic: This looks like it's the only output we use - are there any others that might be useful as well?  Maybe some that have the app name, the elb ID, etc?

phobologic:
👍

scheduler/cloudformation/template.go
ejholmes: Yeah, so far this is the only output. I've kinda opted to not add any extra outputs that don't actually get used programmatically by the backend, since I'm not sure what else we need (but I could totally see exposing the load balancer endpoints back to Empire so you can display that in `emp`).

Anyways, easy to add outputs moving forward.

Issue #820, opened 2016-05-06
Document how to actually try this out


dbp:
I wanted to try using Empire, but there does not seem to be enough documentation to actually do this.

I first tried clicking the launch button on the readme, which successfully created a cluster w/ cloudformation, but I was unable to log in / do anything with it. I both tried using the 'fake/' user/password referenced in the quickstart, and then when that didn't work, followed the (closed) issue #345 (I don't know why it was closed - it is _not_ documented) to try to enable github authentication - I created a github application with the callback from the application, set the two env vars (client_id and client_secret) in the empire ECS task, updated it, confirmed that the task that was running was my task, and then... Didn't know what to do. I tried running `emp apps`, which said it was missing an API token. I tried `emp login`, but didn't know what to put in - isn't the github authentication supposed to offload that to github - why am I still entering in an email/password? The error changed once I added these env vars, so I expect something is happening, but without documentation of what to do, it seems quite impossible to proceed.

Next I destroyed that cluster, and tried the `bin/bootstrap` script. But that printed out a huge warning about this not being something you should actually use (which was different from the output described in the quickstart), and even when I filled out the details, it failed to start. And, regardless, I'm not interested in trying something out that shouldn't actually be used for real. 

I'd really love to try this project - it seems like a really great thing, but I can't do it if there is no documented way to actually:

1. Create the cluster
2. Authenticate with it
3. Deploy an application

ejholmes:
Hey @dbp.

> I first tried clicking the launch button on the readme, which successfully created a cluster w/ cloudformation, but I was unable to log in / do anything with it

Historically this has worked pretty well, so something that was merged recently may have broke the demo cloudformation stack. Sorry about that! I'll take a look at deploying a new dev environment today to see where the problem is.

>  I tried running emp apps, which said it was missing an API token. I tried emp login, but didn't know what to put in - isn't the github authentication supposed to offload that to github - why am I still entering in an email/password?

The GitHub authentication backend allows you to authenticate with Empire using your GH username/password/2fa device and provides team based authorization, so you can lock Empire down to a certain github team (or organization).

Sorry the documentation here is lacking.

> Next I destroyed that cluster, and tried the bin/bootstrap script. But that printed out a huge warning about this not being something you should actually use (which was different from the output described in the quickstart), and even when I filled out the details, it failed to start. And, regardless, I'm not interested in trying something out that shouldn't actually be used for real.

[...]

PR #824, opened 2016-05-09
Improvements to demo/dev CloudFormation stack

ejholmes:
https://github.com/remind101/empire/issues/820 brought up some good pain points. While I wasn't able to find any functional issue with the current CloudFormation stack, there's definitely some improvements that can be made.

1. This now groups and re-labels the parameters when creating a new stack in the CloudFormation ui to make things that are related, more clear: 
2. I added parameters for enabling GitHub authentication.
3. It's no longer required to provide a KeyName parameter, which means you can click through the entire stack creation without changing anything. This used to be annoying if you don't select a KeyName, because the stack would rollback.
4. Changes the quickstart install docs to use the CloudFormation UI, instead of cloning and running `./bin/bootstrap`. `./bin/bootstrap` is mostly meant for devs working on Empire, so it's usually easier to just clickthrough cloudformation if you just want to try it.
5. The IAM permissions that Empire needs are now explicitly defined, which people can use as a reference when productionizing Empire.

### DIFF
--- a/docs/cloudformation.json
+++ b/docs/cloudformation.json
@@ -2,6 +2,48 @@
   "AWSTemplateFormatVersion": "2010-09-09",
   "Description": "Example ECS cluster for Empire",
 
+  "Metadata": {
+    "AWS::CloudFormation::Interface": {
+      "ParameterGroups": [
+        {
+          "Label": { "default": "Empire" },
+          "Parameters": ["LaunchEmpire", "EmpireVersion", "EventsBackend", "RunLogsBackend"]
+        },
+        {
+          "Label": { "default": "GitHub Authentication" },
+          "Parameters": ["GitHubClientId", "GitHubClientSecret", "GitHubOrganization", "GitHubTeamId"]
+        },
+        {
+          "Label": { "default": "Docker" },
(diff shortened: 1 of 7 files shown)


docs/cloudformation.json
```diff
@@ -2,6 +2,48 @@
...
+          "Label": { "default": "Docker" },
+          "Parameters": ["DockerUser", "DockerPass", "DockerEmail", "DockerRegistry"]
+        },
+        {
+          "Label": { "default": "Cluster" },
+          "Parameters": ["DesiredCapacity", "AmiId", "AvailabilityZones", "InstanceType", "KeyName"]
+        }
+      ],
+      "ParameterLabels": {
+        "KeyName": { "default": "Key name" },
```
phobologic: Maybe `ssh Key Name`?

docs/cloudformation.json
```diff
@@ -2,6 +2,48 @@
...
+        "LaunchEmpire": { "default": "Launch" },
+        "EmpireVersion": { "default": "Version" },
```
phobologic: `Empire Daemon Version`

phobologic:
👍

PR #825, opened 2016-05-10
Cleanup docs for CloudFormation backend.

ejholmes:
As of #814, CloudFormation is now the default scheduling backend, so there's some caveats that are no longer relevant in the docs.

Also removed some of the TODO sections, since it makes browsing the docs a little hard (we should still write these!).

### DIFF
--- a/docs/known_limitations.md
+++ b/docs/known_limitations.md
@@ -12,16 +12,6 @@ Right now Empire can only serve http & https services.
 
 You can only have a single exposed process per app, the `web` app.
 
-## Unable to update ELB for an app once it is deployed
-
-Due to a [bug](https://github.com/remind101/empire/issues/498) in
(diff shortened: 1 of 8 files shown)


ejholmes:
Our README could also use some sprucing up. There's a lot of information there, some of which should probably be moved out into the docs, like the section on Architecture (maybe replace it with some bullet points of features).

docs/known_limitations.md
```diff
@@ -12,16 +12,6 @@ Right now Empire can only serve http & https services.
 
 You can only have a single exposed process per app, the `web` app.
 
-## Unable to update ELB for an app once it is deployed
```
phobologic: man, what an awesome side effect of this change

phobologic:
👍

mhahn:
👍

PR #826, opened 2016-05-10
De-dup app environment

ejholmes:
This is one of the changes extracted from https://github.com/remind101/empire/pull/791.

This moves environment and labels in the `scheduler` package from the `Process` type, to the `App` type, which is a better representation of a twelve factor app (shared environment between all processes, but each process can tag on some addition environment vars).

In the long run, this may make it easier to make CloudFormation templates smaller, if we can store most of the environment only once within the template.

### DIFF
--- a/releases.go
+++ b/releases.go
@@ -134,7 +134,7 @@ func (s *releasesService) Rollback(ctx context.Context, db *gorm.DB, opts Rollba
 
 // Release submits a release to the scheduler.
 func (s *releasesService) Release(ctx context.Context, release *Release) error {
-	a := newServiceApp(release)
+	a := newSchedulerApp(release)
 	return s.Scheduler.Submit(ctx, a)
 }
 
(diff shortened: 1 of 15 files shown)


releases.go
```diff
@@ -251,36 +251,45 @@ func releasesCreate(db *gorm.DB, release *Release) (*Release, error) {
...
 }
 
-func newServiceApp(release *Release) *scheduler.App {
+func newSchedulerApp(release *Release) *scheduler.App {
+
 	var processes []*scheduler.Process
 
 	for name, p := range release.Formation {
-		processes = append(processes, newServiceProcess(release, name, p))
+		processes = append(processes, newSchedulerProcess(release, name, p))
```
phobologic: Curious about the change from `Service` -> `Scheduler`

phobologic:
👍

releases.go
ejholmes: Ah yeah. It was named `Service` because package `scheduler` used to be named package `service`, and this just never got changed.

Issue #822, opened 2016-05-07
Emp releases should show who created the release


bmarini:
It would be really nice to see who deployed, changed config etc

PR #827, opened 2016-05-10
Store user and commit message with release description

mhahn:
Fixes: https://github.com/remind101/empire/issues/822

Sample output:
```
$ ./build/emp releases -a acme-inc
v1    May 10 11:59  Deploy remind101/acme-inc:master (fake)
v2    May 10 14:26  Set ELSE, TEST config vars (fake: 'important commit message')
v3    May 10 14:29  Set TEST config var (fake: 'other thing')
v4    May 10 14:35  Rollback to v2 (fake: 'bad variable')
v5    May 10 14:41  Deploy remind101/acme-inc:master (fake: 'deploying new version')
```

### DIFF
--- a/configs.go
+++ b/configs.go
@@ -145,7 +145,7 @@ func (s *configsService) Set(ctx context.Context, db *gorm.DB, opts SetOpts) (*C
 		App:         release.App,
 		Config:      c,
 		Slug:        release.Slug,
-		Description: configsApplyReleaseDesc(vars),
+		Description: configsApplyReleaseDesc(opts),
 	})
 	return c, err
 }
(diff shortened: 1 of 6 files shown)


phobologic:
👍

ejholmes:
I'm totally cool with doing this, although it might be better if we just store the user in a `user` column on the releases table (I think just store the username would be enough). We actually [already try to display this field in emp](https://github.com/remind101/empire/blob/master/cmd/emp/releases.go#L145), but we don't set it.

Either way, 👍

mhahn:
yea, we also attempt to display it here: https://github.com/remind101/empire/blob/master/cmd/emp/releases.go#L193

i feel like if we're going to start to store user we should do it the right way and reference a user id. that also seems like it would be needed for implementing some kind of permissions system as well.

i'm going to merge this and we can discuss adding better support for user.

PR #828, opened 2016-05-11
Fix HTTPS listener in CloudFormation backend.

ejholmes:
When adding the HTTPS listener, we want to actually use https. Doh!


mhahn:
👍

Issue #829, opened 2016-05-12
Support specifying a duration within emp log


mhahn:
This should allow you to start streaming logs historically.

Kinesis supports this with the `AT_TIMESTAMP` shard iterator.

PR #830, opened 2016-05-12
Log duration

mhahn:
Allows you to start streaming logs from a specific point in time.

ie:

`emp log -a acme-inc -d 10m` will start streaming logs from 10 minutes ago

### DIFF
--- a/server/heroku/logs.go
+++ b/server/heroku/logs.go
@@ -13,9 +13,19 @@ type PostLogs struct {
 	*empire.Empire
 }
 
+type PostLogsForm struct {
+	Duration string
+}
+
 func (h *PostLogs) ServeHTTPContext(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
 	a, err := findApp(ctx, h)
 	if err != nil {
+		return err
+	}
+
+	var form PostLogsForm
+
+	if err := Decode(r, &form); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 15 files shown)


mhahn:
Addresses https://github.com/remind101/empire/issues/829

server/heroku/logs.go
```diff
@@ -13,18 +13,28 @@ type PostLogs struct {
 	*empire.Empire
 }
 
+type PostLogsForm struct {
+	Duration string
```
ejholmes: On a similar note to https://github.com/remind101/kinesumer/pull/22/commits/11e496a3d697d12c961ea2307dc0e97ccbdbd28c#r62950197. I think it might be better if we do the `time.ParseDuration` at the edge of user input (in the emp client) and make this parameter be an int (milliseconds/seconds).

ejholmes:
Man, this is awesome.

ejholmes:
What version of the aws-sdk-go does this upgrade us to? Can't tell from the diff (also, mind opening that as a separate PR first?).

mhahn:
Opened up a separate PR (https://github.com/remind101/empire/pull/833), I updated to the latest https://github.com/aws/aws-sdk-go/releases/tag/v1.1.26

mhahn:
@ejholmes addressed your comments.

when https://github.com/remind101/kinesumer/pull/22 is merged, i'll fix the ref to that here too.

ejholmes:
:+1: after https://github.com/remind101/empire/pull/833 and https://github.com/remind101/kinesumer/pull/22

PR #831, opened 2016-05-12
Describe tasks in chunks of 100.

ejholmes:
The `ecs.DescribeTasks` call can only describe 100 tasks per call, so we break up the tasks into chunks of 100.


phobologic:
👍

PR #835, opened 2016-05-17
Simplify cert attaching.

ejholmes:
This just simplifies attaching SSL certs, so that you can provide just the name instead of the full cert ARN.


phobologic:
👍

mhahn:
👍

Issue #832, opened 2016-05-12
Events should be written to the apps log stream


mhahn:
When events are published, we should write them to the log stream

mhahn:
@ejholmes to implement this:

- we can use: https://github.com/remind101/kinesis which only works if the stream exists
- or i was thinking we could replace the "writer" code within https://github.com/remind101/logspout-kinesis/ to use the above kinesis library and move the stream creation etc. to the kinesis library.

that way when an event is published (ie. create/deploy etc) and a stream doesn't exist, the same code is being called to create and write to the stream.

i was thinking if we pulled the kinesis code out of logspout-kinesis that would help us move away from logspout in the future, but if you switch to logstash, you would be using some ruby kinesis output plugin.

i see two options for tackling this:

1. use kinesis library to write to the stream with the app id, if the stream doesn't exist, just don't write the event.
2. refactor the stream creation process out of logspout-kinesis and into the kinesis library that will get called by both logspout-kinesis and empire.

1 is the faster implementation that i think makes the most sense. i don't think it matters if you drop some events from the stream, there is no guarantee that the DropWriter isn't going to do that either and the window of the stream not existing is small.

thoughts?

phobologic:
I'm honestly open to either of them - the fact that logstash seems likely in our future now makes me wonder if this shouldn't just wait for that implementation.

PR #836, opened 2016-05-17
Support publishing events to an app's log stream

mhahn:
Addresses: https://github.com/remind101/empire/issues/832

I wasn't sure about the name for the event streamer.

Calling it `kinesis` seemed off, because I would assume that a true `kinesis` event backend would take a stream name as a config and write everything to that stream.

### DIFF
--- a/events.go
+++ b/events.go
@@ -23,7 +23,7 @@ type RunEvent struct {
 	Attached bool
 	Message  string
 
-	appID string
+	app *App
 }
 
 func (e RunEvent) Event() string {
@@ -42,8 +42,8 @@ func (e RunEvent) String() string {
 	return appendCommitMessage(msg, e.Message)
 }
 
-func (e RunEvent) AppID() string {
-	return e.appID
+func (e RunEvent) GetApp() *App {
+	return e.app
 }
 
 // RestartEvent is triggered when a user restarts an application.
@@ -53,7 +53,7 @@ type RestartEvent struct {
 	PID     string
 	Message string
 
-	appID string
+	app *App
 }
 
 func (e RestartEvent) Event() string {
@@ -70,8 +70,8 @@ func (e RestartEvent) String() string {
 	return appendCommitMessage(msg, e.Message)
 }
 
-func (e RestartEvent) AppID() string {
-	return e.appID
+func (e RestartEvent) GetApp() *App {
+	return e.app
 }
 
 // ScaleEvent is triggered when a manual scaling event happens.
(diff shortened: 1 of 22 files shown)


events/app/app.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// EventStream is an implementation of the empire.EventStream interface backed
+// by Kinesis.
+type EventStream struct {
+	empire  *empire.Empire
+	kinesis kinesisClient
+}
+
+type Event interface {
```
ejholmes: Minor, but might be nice to move this interface below https://github.com/remind101/empire/blob/master/events.go#L191-L198 as:

```go
// An AppEvent is an Event that relates to a specific app.
type AppEvent interface {
        Event
        App() *App // AppID is fine too.
}
```

events.go
```diff
@@ -215,22 +251,24 @@ var NullEventStream = EventStreamFunc(func(event Event) error {
...
+	events  chan Event
 }
 
 // AsyncEvents returns a new AsyncEventStream that will buffer upto 100 events
 // before applying backpressure.
-func AsyncEvents(e EventStream) *AsyncEventStream {
-	s := &AsyncEventStream{
-		EventStream: e,
-		events:      make(chan Event, 100),
+func AsyncEvents(s []EventStream) *AsyncEventStream {
```
ejholmes: You could keep this as an `EventStream` and just make `[]EventStream` implement the EventStream interface:

```go
// MutliEventStream is an EventStream implementation that publishes the event to multiple EventStreams, returning any errors after publishing to all streams.
type MultiEventStream []EventStream

func (streams MultiEventStream) PublishEvent(event Event) error {
	var errors []error
	for _, s := range streams {
		errors = append(errors, s.PublishEvent(event))
	}

	if len(errors) > 0 {
		// Probably a better way to generate a MultiError here.
		return fmt.Errorf("errors publishing to event streams: %v", errors)
	}

	return nil
}
```

ejholmes:
This is awesome. Other than those comments, 👍.

events.go
mhahn: woah, thats cool!

events/app/app.go
mhahn: yea i like that

events.go
mhahn: https://github.com/hashicorp/go-multierror seems to handle multierrors cleanly so i went with that instead of reimplementing something.

ejholmes:
Awesome 👍

Issue #756, opened 2016-02-10
Set deadlines on deployments.


ejholmes:
Couple of things we should do:

1. There should be some sensible default on deployment timeouts (like 20 mins or so).
2. When Empire shuts down, any in progress deployments should be canceled gracefully (probably just log something to the log stream).

ejholmes:
For the most part, cancelations and timeouts are handled at the scheduler level, within the CloudFormation backend.

Issue #837, opened 2016-05-18
Better scheduler submission queue


ejholmes:
With the switch to CloudFormation, it can be annoying if you want to make quick updates in short succession (say, for example, setting environment variables). Empire currently needs to wait for CloudFormation to reach a "steady state" (CREATE_COMPLETE/UPDATE_COMPLETE) before submitting a new template.

There may be things we can optimize to make the UX better. For example, if I do the following:

```console
emp set FOO=bar
emp set BAR=foo
emp set FOOBAR=foobar
```

The first command will create a new release, and trigger a stack update. The second command will have to wait for the first to finish. The second command will have to wait for both to finish, but in reality there's no need to submit the second release.

To be fair, this was technically still a problem with the old scheduler, if you did 5 updates in short succession.

ejholmes:
I'm gonna go ahead and jot down some different scenarios, and the expected behavior, and we can use that to help guide a technical implementation for what we want:

#### Scenario 1 - Setting environment variables while a deployment is happening

```console
$ emp deploy remind101/acme-inc
release 1
$ emp set FOO=bar
release 2
```

* We'd expect the first command to start updating the stack.
* We'd expect the second command to return immediately and the stack update to be queued behind the first.

#### Scenario 2 - Setting multiple environment variables in quick succession

```console
$ emp set FOO=bar
release 1
$ emp set BAR=foo
release 2
$ emp set FOOBAR=foobar
release 3
```

* We'd expect the first command to return immediate, and a stack update to be triggered.
* We'd expect the second command to return immediately and have it's stack update be queued behind the first.
* We'd expect the last command to return immediately, and have it's stack update overwrite the stack update for release 2, since it's now obsolete.

#### Scenario 3 - Rolling back while there's an existing stack update

```console
$ emp deploy remind101/acme-inc
release 1
$ emp set FOO=bar
release 2
$ emp rollback v1
release 3
```

* We'd expect the first command to start updating the stack.
* We'd expect the second command to return immediately and have it's stack update be queued behind the first.
* We'd expect the rollback command to immediately cancel any stack update, then start updating the stack with the new stack. This one gets tricky, because canceling a stack update will trigger a stack rollback, which we have to wait for.

ejholmes:
Seems like we want a queue, per app, that has a maximum depth of 2, where a Push operation replaces the top most item.

[...]

PR #841, opened 2016-05-20
Set explicit deadlines on operations

ejholmes:
Closes https://github.com/remind101/empire/issues/756
Slightly improves https://github.com/remind101/empire/issues/837

This sets explicit deadlines on `Deploy`, `Set`, `Rollback` and `CertsAttach` operations, which all use scheduler.Submit to submit a new release to the scheduler backend. The CloudFormation scheduler now will also use the deadline to cancel waiting for the stack to move to a stabilized state.

The effect of this is slightly better error messages. For example:

```console
$ emp set FOO=bar -a acme-inc
error: stack did not stabilize from UPDATE_IN_PROGRESS after waiting 28 seconds
```

Instead of seeing an `EOF` error returned.

There's further changes we can make to improve the behavior so that actually hitting this error is rare (like queueing releases instead of performing them within the request/making rollbacks trigger a stack rollback first/etc), but I figure this is at least a good start.

### DIFF
--- a/empire.go
+++ b/empire.go
@@ -26,6 +26,14 @@ var (
 	DefaultReporter = reporter.NewLogReporter()
 )
 
+// Timeouts for certain API actions.
+var (
+	TimeoutDeploy      = 20 * time.Minute
+	TimeoutSet         = 30 * time.Second
+	TimeoutRollback    = 30 * time.Second
+	TimeoutCertsAttach = 30 * time.Second
+)
+
 const (
 	// WebPort is the default PORT to set on web processes.
 	WebPort = 8080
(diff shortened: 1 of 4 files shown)


empire.go
```diff
@@ -273,6 +281,9 @@ func (opts SetOpts) Validate(e *Empire) error {
 // Config. If the app has a running release, a new release will be created and
 // run.
 func (e *Empire) Set(ctx context.Context, opts SetOpts) (*Config, error) {
+	ctx, cancel := context.WithTimeout(ctx, TimeoutSet)
```
ejholmes: Thinking about this a bit more, these should really be defined inside server/heroku instead of here.

mhahn: 👍

mhahn:
:+1:

ejholmes:
Gonna close this one for now since https://github.com/remind101/empire/pull/852 was merged. Still would be good to come back to this, but less of an immediate need right now.

PR #842, opened 2016-05-21
Don't wait for ECS services to stabilize

ejholmes:
Switching to CloudFormation has some less than optimal behavior with the way ECS services are updated; CloudFormation will wait for the ECS service to stabilize, when updating. This has some negative consequences:

1. The stack is locked during the entire duration, making it impossible to submit a new stack template until all of the services have stabilized.
2. If the cluster has insufficient resources, the stack could potentially stay in an UPDATE_IN_PROGRESS state for a very long time, or eventually rollback.

From Empire's perspective, we'd like CloudFormation to simply submit the updates to ECS, then complete, which would mimic the behavior of the old non-cloudformation scheduler.

This PR adds a `Custom::ECSService` resource to our custom resource handlers. This operates similarly to `AWS::EC2::Service`, except it does not wait for the service to stabilize. Doing this should allow us to submit changes to CloudFormation faster, and should resolve most of https://github.com/remind101/empire/issues/837.

For the time being, this can be enabled on a per app basis by setting `ECS_UPDATES=fast` in the apps environment. You can switch back and forth between the two without any apparent downtime. This may become the default before the next release.

### DIFF
--- a/scheduler/cloudformation/template.go
+++ b/scheduler/cloudformation/template.go
@@ -317,15 +317,7 @@ func (t *EmpireTemplate) Build(app *scheduler.App) (interface{}, error) {
 		}
 
 		service := fmt.Sprintf("%s", key)
-		if fast {
-			service = fmt.Sprintf("%sService", service)
-		}
-		serviceMappings = append(serviceMappings, map[string]interface{}{
-			"Fn::Join": []interface{}{
-				"=",
-				[]interface{}{p.Type, map[string]string{"Ref": service}},
-			},
-		})
+		ecsServiceType := "AWS::ECS::Service"
 		serviceProperties := map[string]interface{}{
 			"Cluster": t.Cluster,
 			"DesiredCount": map[string]string{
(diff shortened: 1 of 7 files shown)


scheduler/cloudformation/template.go
```diff
@@ -306,6 +317,9 @@ func (t *EmpireTemplate) Build(app *scheduler.App) (interface{}, error) {
 		}
 
 		service := fmt.Sprintf("%s", key)
+		if fast {
```
mhahn: we should be able to just do 1 check for fast, makes it a little cleaner IMO

server/cloudformation/ecs.go
```diff
@@ -0,0 +1,133 @@
...
+
+type LoadBalancer struct {
+	ContainerName    *string
+	ContainerPort    *IntValue
+	LoadBalancerName *string
+}
+
+// ECSServiceProperties represents the properties for the Custom::ECSService
+// resource.
+type ECSServiceProperties struct {
```
mhahn: nice this is great

server/cloudformation/cloudformation.go
```diff
@@ -296,3 +314,37 @@ func (c *CustomResourceProvisioner) Handle(message *sqs.Message) error {
...
+// It's common to use `Ref`'s inside templates, which means the value of some
+// properties could be a string or an integer.
+type IntValue int64
+
+func intValue(v int64) *IntValue {
+	i := IntValue(v)
+	return &i
+}
+
+func (i *IntValue) UnmarshalJSON(b []byte) error {
```
mhahn: is this used anywhere?

mhahn:
👍 crazy how powerful custom resources are

server/cloudformation/cloudformation.go
ejholmes: This gets used when we decode the resource request into a `Request` (See the DesiredCount property of ECSServiceProperties).

ejholmes:
@mhahn I added a [commit](https://github.com/remind101/empire/pull/842/commits/92db13f51c527cff97fd0f4b25c38658f959c4af) to this that adds support for re-creating the service if one of the non-updatable params are changed (Role, ServiceName, LoadBalancers, Cluster), so it has the same behavior as `AWS::ECS::Service`. Also adds a postfix to the service name when creating a service, as to avoid any collisions when a replacement is required.

I was originally thinking that the Update request would have to be responsible for cleaning up the old resource, but it turns out, if you return a new PhysicalResourceId, CloudFormation will send a DELETE request for the old resource during the UPDATE_CLEANUP state. Cool!

scheduler/cloudformation/template.go
ejholmes: Yep, good call. Updated.

PR #843, opened 2016-05-21
Break out EnvironmentResource

mhahn:
still need to test some things, but wanted to get the guts of it out there for review


mhahn:
@ejholmes if you merge master into app-resource this diff should get smaller

mhahn:
we should also probably standardize the naming, but i thought i'd let you take care of that.

my thinking is that:

server/cloudformation/apps.go -> server/cloudformation/**app.go**
InstancePortsProvisioner -> InstancePortResource
CustomResourceProvisioner -> CustomResources

etc. but you probably have more of an opinion

server/cloudformation/environment.go
```diff
@@ -0,0 +1,160 @@
...
+
+	return err
+}
+
+type Variable struct {
+	Key   string
+	Value *string
+}
+
+func parseVariable(index int, input interface{}) (*Variable, error) {
```
ejholmes: We should just change the type of `ResourceProperties` and `OldResourceProperties` types to json.Message, then we can just unmarshal into a struct. I already did this in https://github.com/remind101/empire/pull/842, but can break it out into a smaller PR that we can merge first.

server/cloudformation/environment_test.go
```diff
@@ -0,0 +1,475 @@
...
+	}
+}
+
+// Run the tests with empiretest.Run, which will lock access to the database
+// since it can't be shared by parallel tests.
+func TestMain(m *testing.M) {
+	empiretest.Run(m)
+}
+
+func TestProvision_Create(t *testing.T) {
```
ejholmes: I think inside this package, we should just unit test the struct (mock out empire.Empire). Some actual integration level tests for the cloudformation resources would be nice, but we should put integration level tests in `./tests`.

server/cloudformation/environment.go
ejholmes: Actually, maybe even better would be to just do [this](https://github.com/remind101/empire/pull/842/commits/68a226ec0a3c3065e35037b29c507f465570be84)

mhahn: yea that looks great

server/cloudformation/environment_test.go
mhahn: ok cool

ejholmes:
You'll need to rebase this since https://github.com/remind101/empire/pull/842 was merged, but it should make everything here a lot easier.

mhahn:
@ejholmes i'm going to close this and open another PR against master

Issue #840, opened 2016-05-19
Destroying a stack that hasn't been provisioned fails


mhahn:
```
$ ./build/emp create test
$ ./build/emp destroy test
warning: This will destroy test and its add-ons. Please type "test" to continue:
> test
error: no stack for app found
```

phobologic:
Closed by #845

PR #845, opened 2016-05-23
Relaxed cloudformation scheduler Remove()

jonson:
I've been using empire in our pre-production environment for a while now and thought it's time to make my first contribution.  I just updated my empire servers to get the new cloudformation scheduler and I've come across issue #840. 

This PR handles two cases where there are currently un-destroyable apps.

1. App is created in empire but no corresponding stack is created in the db.
2. App is created, stack is created in db, but there's no associated stack in cloudformation (ex: someone manually deleted the cloudformation stack outside of empire).

I'm not sure if this is the best approach, but AFAIK it aligns with the previous scheduler's remove behaviour.

### DIFF
--- a/scheduler/cloudformation/cloudformation.go
+++ b/scheduler/cloudformation/cloudformation.go
@@ -349,9 +349,14 @@ func (s *Scheduler) Remove(ctx context.Context, appID string) error {
 	return tx.Commit()
 }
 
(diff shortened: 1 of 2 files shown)


ejholmes:
closes #840.

This looks great! Much appreciated.

Issue #846, opened 2016-05-24
Removing a process from the Procfile makes it difficult to update the stack


ejholmes:
When removing an entry from the Procfile, the stack can't be updated because we try to set a parameter that is no longer valid.

**Example**

```console
$ emp deploy remind101/acme-inc:master
$ emp deploy remind101/acme-inc:no-worker
420890c9e918: Already exists
a3ed95caeb02: Pull complete
1e14aab0083d: Already exists
4393485f2bb3: Already exists
421fc7ff962c: Pull complete
ce70a5759cca: Pull complete
Digest: sha256:c4ab91508f77f83b7bc26174db31cf628830157605c92f6bafb1943e28256c16
Status: Downloaded newer image for remind101/acme-inc:no-worker
error: error updating stack: ValidationError: Parameters: [workerScale] do not exist in the template
        status code: 400, request id: fd296ad6-2157-11e6-8f24-cf26b7634bc4
```

PR #848, opened 2016-05-24
Allow processes to be removed with CloudFormation backend.

ejholmes:
Fixes https://github.com/remind101/empire/issues/846

Previously, we'd merge in all the parameters from the existing stack even if we were submitting a new stack template that had different parameters. Now, we only merge in existing parameters if we using the same template (UsePreviousTemplate).


mhahn:
👍

Issue #847, opened 2016-05-24
Restarting relies on order of environment variables


ejholmes:
In the CloudFormation backend, a successful restart relies on environment variables being re-ordered. Instead, we should be able to update a param to trigger a restart of a service.

I think we can do this in two steps:

1. Change the backend to sort environment variables, then add a Docker label whose value is based on a "Restart" parameter. Changing this value will trigger a new task definition to be created, which will trigger a service restart.
2. Add a `Restart` and `RestartProcess` method to the Scheduler interface that updates these parameters, instead of perform a full `Submit`.

ejholmes:
I've noticed that tests in ./scheduler/cloudformation/template_test.go fail intermittently because of this issue too.

ejholmes:
This was fixed in https://github.com/remind101/empire/pull/850

PR #850, opened 2016-05-24
Better restart with CloudFormation backend

ejholmes:
Step 1 of https://github.com/remind101/empire/issues/847#issue-156405642.

This makes it so we're not dependent on random sorting of environment variables to restart an application and preps the backend for an easy implementation of `Restart` and `RestartProcess` methods (which means we'll finally support `emp restart <process>`, hooray!).

Implementation wise, this works by setting an `cloudformation.restart_key` docker label on the task definition, with it's value as a composition of the app and process restart keys. If you change the global restart key, all task definitions are re-created. If you change a process restart key, the process task definition is updated (which ultimately causes a service restart).

### DIFF
--- a/scheduler/cloudformation/cloudformation.go
+++ b/scheduler/cloudformation/cloudformation.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"time"
 
+	"code.google.com/p/go-uuid/uuid"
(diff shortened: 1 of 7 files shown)


phobologic:
👍

mhahn:
👍

PR #851, opened 2016-05-24
Don't set EMPIRE_CREATED_AT env var.

ejholmes:
Empire used to use this env var internally to determine when a task was started, but we've since implemented that using the native [StartedAt, CreatedAt, StoppedAt attributes](https://github.com/remind101/empire/blob/master/scheduler/ecs/ecs.go#L276-L284) in https://github.com/remind101/empire/pull/683 (released in 0.10.0).

This should be safe to remove since it was only used internally, and wasn't documented anywhere, but it shouldn't be merged until https://github.com/remind101/empire/pull/850 is merged (restarts would always set a new value for this).


phobologic:
👍

mhahn:
👍

PR #852, opened 2016-05-25
Queue stack updates behind each other.

ejholmes:
This locks stack updates with an advisory lock, and makes updates asynchronous when they have to wait on an existing stack update to complete.

**TODO**

* [x]  More tests for concurrency
* [x] Timeouts
* [x] Validate stack template.
* [x] Handle error returned when advisory lock is canceled.
* [ ] Handle `context.Done()` (cancel the update if it's pending).


ejholmes:
Wonder if this might be easier to conceptualize if we implemented the sync.Locker interface backed by the table, then it just looks like we're using a mutex:

```go
q.Lock() // Obtains a lock on the "active" stack update
defer q.Unlock() // Unlocks the "active" stack update.

// Update the stack
```

But things get a little tricky when we want to obsolete a stack update.

phobologic:
I'm not sure about the idea of obsoleting the old update.  Will a release still be created in the database, just not rolled out?

ejholmes:
@phobologic yep. My thought process is that, if we're creating a bunch of releases in short succession, there's not much advantage to deploying each one sequentially when the desired state is the last release.

But, it does add some complexity. Maybe it would be better to just start with a more naive approach that doesn't do obsoletion, which should be ok for most cases (especially with the `Custom::ECSService` resource).

phobologic:
👍 for starting simple, cool.

ejholmes:
Ok, this needs more tests, but I'm actually pretty happy with this implementation now, and still gets us stack update obsoletion without much complexity (literally just 1 extra query).

Implementation wise, whenever we execute a stack update, we obtain an advisory lock. Before we obtain the lock, we just cancel any pending advisory locks that haven't yet been granted.

scheduler/cloudformation/cloudformation.go
```diff
@@ -280,13 +288,106 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
...
+//
+// * If there are no updates currently in progress, the stack will be updated .
+// * If there is a currently active stack update, the update will be queued behind it.
 //
-// TODO: Timeout?
-func (s *Scheduler) updateStack(stack *cloudformation.Stack, input *cloudformation.UpdateStackInput, wait bool) (*cloudformation.UpdateStackOutput, error) {
-	status := *stack.StackStatus
-	stackName := input.StackName
+// An error will be sent on the `done` channel when the stack update completes.
+func (s *Scheduler) enqueueStackUpdate(input *cloudformation.UpdateStackInput, locked chan struct{}, done chan error) error {
```
ejholmes: Maybe we can generalize the locking a bit so that it can be reused for other stack operations (e.g. destroy).

ejholmes:
One thing that might be nice as a follow up to this is to surface more information about whether a release is pending, fullfilled or errored. It would be nice to see something like this in `emp releases`:

```
v77 success  May 26 09:21  Set BAR config var (ejholmes)
v78 success  May 26 09:22  Set FOOBAR config var (ejholmes)
v79 pending  May 26 10:00  Deploy remind101/acme-inc:master (ejholmes)
```

Something for another day.

scheduler/cloudformation/cloudformation.go
```diff
@@ -316,44 +422,29 @@ func (s *Scheduler) updateStack(stack *cloudformation.Stack, input *cloudformati
 		}
 	}
 
-	// If there's currently an update happening, wait for it to
-	// complete.
-	if strings.Contains(status, "IN_PROGRESS") {
```
ejholmes: We should probably keep this around to ensure that the stack is in an updateable state before we start. Without this, creating a new stack then immediately deploying would cause an error.

scheduler/cloudformation/cloudformation.go
```diff
@@ -280,13 +288,111 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
...
+
+	// Start up a goroutine that will wait for this stack update to
+	// complete, and release the lock when it completes.
+	go s.waitUntilStackUpdateComplete(tx, *input.StackName, done)
+
+	return nil
+}
+
+func (s *Scheduler) waitUntilStackUpdateComplete(tx *sql.Tx, stackName string, done chan error) {
+	err := s.cloudformation.WaitUntilStackUpdateComplete(&cloudformation.DescribeStacksInput{
```
ejholmes: This should have some reasonable timeout (like 10 minutes or so), just to ensure that this can't hold the lock forever.

scheduler/cloudformation/cloudformation.go
```diff
@@ -280,13 +305,128 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
 	return nil
 }
 
-// updateStack performs a stack update, but waits for the stack to enter a
-// stable state before starting.
+// updateStack enqueues the stack update, and waits up to 10 seconds for the
```
mhahn: this waits 2 seconds now right?

scheduler/cloudformation/cloudformation.go
```diff
@@ -280,13 +305,128 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
...
+// stack updated to be submitted.
```
mhahn: /updated/update

scheduler/cloudformation/cloudformation.go
```diff
@@ -280,13 +305,128 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
...
+	_, err = tx.Exec(fmt.Sprintf("SET LOCAL lock_timeout = %d", timeout))
+	if err != nil {
+		return fmt.Errorf("error setting lock timeout: %v", err)
+	}
+
+	// We have one lock per CloudFormation stack.
+	key := crc32.ChecksumIEEE([]byte(fmt.Sprintf("stack_%s", *input.StackName)))
+
+	// If there are any other pending updates for this stack, fuck em.
+	_, err = tx.Exec(`SELECT pg_cancel_backend(pending.pid) FROM (SELECT pid FROM pg_locks WHERE locktype = 'advisory' AND granted = 'f' AND objid = $1) as pending`, key)
```
mhahn: this is cool

scheduler/cloudformation/cloudformation.go
```diff
@@ -280,13 +305,128 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
...
+	if err != nil {
+		return fmt.Errorf("error canceling pending locks: %v", err)
+	}
+
+	context := fmt.Sprintf("stack %s", *input.StackName)
+	if input.TemplateURL != nil {
+		context = fmt.Sprintf("%s (%s)", context, *input.TemplateURL)
+	}
+
+	_, err = tx.Exec(fmt.Sprintf("SELECT pg_advisory_lock($1) /* %s */", context), key)
```
mhahn: good call with the context comment

ejholmes:
Ok, I think this is good to go. I extracted the locking semantics into a `pkg/pg/lock` package that we can probably re-use elsewhere. It just provides some primitives to wrap `pg_advisory_lock`, and adds a bunch of concurrency related tests.

ejholmes:
Also added a commit to generalize the locking so that CreateStack also obtains the lock (which means, `emp create && emp deploy remind101/acme-inc && emp set FOO=bar` behaves like you'd expect it to, and doesn't return an error).

pkg/pg/lock/lock.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	_, err := l.tx.Exec(fmt.Sprintf("SELECT pg_advisory_lock($1) /* %s */", l.Context), l.key)
+	if err != nil {
+		// If there's an error trying to obtain the lock, probably the
+		// safest thing to do is commit the transaction and make this
+		// lock invalid.
+		l.commit()
+
+		// This will happen when a newer stack update obsoletes
```
mhahn: nit, this comment is out of context now

scheduler/cloudformation/cloudformation.go
```diff
@@ -280,13 +294,118 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
...
+	go func() {
+		p := func() error { return s.executeStackUpdate(input) }
+		waiter := s.cloudformation.WaitUntilStackUpdateComplete
+		errCh <- s.performStackOperation(*input.StackName, p, waiter, locked, done)
+	}()
+
+	var err error
+	select {
+	case <-time.After(lockWait):
+		// FIXME: At this point, we don't want to affect UX by waiting
```
mhahn: i think we should be publishing an event here, that way you get it via SNS if you have that setup, at the very least it should go to the app stream?

mhahn: could be a separate PR

scheduler/cloudformation/cloudformation.go
```diff
@@ -280,13 +294,118 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
...
+		return fmt.Errorf("error canceling pending stack operation: %v", err)
+	}
+
+	if err := l.Lock(); err != nil {
+		// This will happen when a newer stack update obsoletes
+		// this one. We simply return nil.
+		//
+		// TODO: Should we return an error here?
+		if err == pglock.Canceled {
+			return nil
```
mhahn: i don't think this should be an error, but having it just return nil feels weird too.

i was thinking you could maybe return `(bool, error)` where bool is whether or not the execution took place, but i'm not sure where you would use it. it seems noisy to push these out as notifications, but could be useful to log?

mhahn:
looks great, some minor comments, but 👍

scheduler/cloudformation/cloudformation.go
ejholmes: Yeah, definitely. I was thinking we could change the interface of `scheduler.Submit` to take a `done chan error` arg. Empire could then use this to submit quickly, but receive on that channel to be notified when it was actually fullfilled (e.g. the stack update completed).

Then we could expose this in the `emp releases` command (as well as events), so you can tell if a release is pending, completed, canceled, etc.

But yeah, probably a separate PR.

scheduler/cloudformation/cloudformation.go
ejholmes: Yep, I think this ties into https://github.com/remind101/empire/pull/852#discussion_r65461995

scheduler/cloudformation/cloudformation.go
mhahn: yea that sounds great

PR #854, opened 2016-05-26
Add the app release as a stack output.

ejholmes:
This just makes it easier to see what release of an app the cloudformation stack is on.

Also adds the empire version that was used to submit the cloudformation stack.


phobologic:
👍 Awesome, this will be so nice.

mhahn:
👍

PR #855, opened 2016-05-27
Support for App & Environment resources

mhahn:


server/cloudformation/apps.go
```diff
@@ -0,0 +1,88 @@
...
+// AppResource is a Provisioner that will manage an Empire application
+type AppResource struct {
+	empire appClient
+}
+
+func (p *AppResource) Provision(req Request) (id string, data interface{}, err error) {
+	ctx := context.Background()
+	user := NewUser()
+
+	properties := req.ResourceProperties.(*AppProperties)
```
mhahn: should we be checking if this fails?

server/cloudformation/environment.go
```diff
@@ -0,0 +1,147 @@
...
+type Variable struct {
+	Name  *string
+	Value *string
+}
+
+// EnvironmentProperties represents the properties for the
+// Custom::EmpireAppEnvironment
+type EnvironmentProperties struct {
+	AppId     *string
+	Variables []Variable
```
mhahn: this was way better than before, 👍 on setting that up @ejholmes

server/cloudformation/apps.go
ejholmes: This should never fail, unless it's a programmer error, so I think it's fine to just let it panic in that case.

server/cloudformation/apps.go
```diff
@@ -0,0 +1,88 @@
...
+		app, err = p.empire.Create(ctx, empire.CreateOpts{
+			User:    user,
+			Name:    *name,
+			Message: "Creating app via Cloudformation",
+		})
+		if err != nil {
+			return "", nil, err
+		}
+
+		return app.ID, map[string]string{"Id": app.ID}, nil
```
ejholmes: My original assumption was that we couldn't use `Ref` with the physical Id, so I added the Id as a returned property that we could use with `GetAtt`, but I don't think we need to do that. Can just `return app.ID, nil, nil` here.

server/cloudformation/cloudformation.go
```diff
@@ -158,6 +158,12 @@ func NewResponseFromRequest(req Request) Response {
 	}
 }
 
+// NewUser returns an empire.User that should be used by sources when making
+// requests to empire.
+func NewUser() *empire.User {
+	return &empire.User{Name: "cloudformation"}
```
ejholmes: Minor, might read nicer in events if we uppercase this `CloudFormation`.

server/cloudformation/cloudformation.go
```diff
@@ -158,6 +158,12 @@ func NewResponseFromRequest(req Request) Response {
 	}
 }
 
+// NewUser returns an empire.User that should be used by sources when making
+// requests to empire.
+func NewUser() *empire.User {
```
ejholmes: Also minor, might as well just unexport this `newUser`.

server/cloudformation/test.go
```diff
@@ -0,0 +1,33 @@
+package cloudformation
```
ejholmes: May wanna rename this `<prefix>_test.go` otherwise the imports will get included in the binary and the types would be available at runtime.

server/cloudformation/cloudformation.go
```diff
@@ -189,11 +195,17 @@ type CustomResourceProvisioner struct {
...
 	return &CustomResourceProvisioner{
 		Provisioners: map[string]Provisioner{
 			"Custom::InstancePort": &InstancePortsProvisioner{
-				ports: lb.NewDBPortAllocator(db),
+				ports: lb.NewDBPortAllocator(empire.DB.DB.DB()),
+			},
+			"Custom::EmpireApp": &AppResource{
+				empire: empire,
+			},
+			"Custom:EmpireAppEnvironment": &EnvironmentResource{
```
ejholmes: Missing a `:` here.

Maybe we should name the structs the same as the resource name, with `Resource` appended, just so we have a common convention:

```go
"Custom::EmpireApp": EmpireAppResource
"Custom::EmpireAppEnvironment": EmpireAppEnvironmentResource
```

Right now, EnvironmentResource will conflict with the one I have in https://github.com/remind101/empire/pull/859

server/cloudformation/environment.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func (p *EnvironmentResource) Provision(req Request) (id string, data interface{}, err error) {
+	ctx := context.Background()
+	user := NewUser()
+	properties := req.ResourceProperties.(*EnvironmentProperties)
+
+	switch req.RequestType {
+	case Create:
+		if *properties.AppId == "" {
```
ejholmes: I think we'd also want to check if it was nil?

```go
if properties.AppId == nil || *properties.AppId == ""
```

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 * You can now specify the CPU and memory constraints for attached one-off tasks with the `-s` flag to `emp run` [#809](https://github.com/remind101/empire/pull/809)
 * You can now provide a duration to `emp log` with the `-d` flag to start streaming logs from a specific point in time ie (5m, 10m, 1h) [#829](https://github.com/remind101/empire/issues/829)
 * If log streaming is enabled, Empire will attempt to write events to the kinesis stream for the application [#832](https://github.com/remind101/empire/issues/832)
+* You can now provision Empire applications and set environment variables from CloudFormation stacks using a `Custom::App` resource [#819](https://github.com/remind101/empire/pull/819)
```
ejholmes: Need to change this to "using the `Custom::EmpireApp` and `Custom::EmpireAppEnvironment` resources".

It'd be good if we could write some docs about how to use these.

server/cloudformation/environment.go
```diff
@@ -0,0 +1,145 @@
...
+			return "", nil, fmt.Errorf("missing parameter: AppId")
+		}
+		id = *properties.AppId
```
ejholmes: I haven't tested this, but I wonder if there would be a problem returning the same physical id as the app.

ejholmes:
This is awesome 👍. I'm so excited to start using this.

CHANGELOG.md
mhahn: yea i think we should do a new release and then write some docs on all of these new features

server/cloudformation/environment.go
mhahn: from the docs:

> This value should be an **identifier unique to the custom resource vendor**, and can be up to 1Kb in size. The value must be a non-empty string.

i believe both of these are treated as different resource vendors, so the PhysicalResourceId should just need to be unique within that scope.

server/cloudformation/cloudformation.go
mhahn: ok, i also renamed the file to `app_environment.go` to handle the conflict with the file name too.

mhahn:
@ejholmes this good to merge?

mhahn:
Implemented this with stacker here: https://github.com/remind101/stacker_blueprints/pull/24

The Custom::EmpireAppEnvironment is harder to implement cleanly with the way stacker currently handles parameters.

Ideally we want something like:

```
  - name: acmeAppEnv
    class_path: stacker_blueprints.empire.app.Environment
    parameters:
      ServiceToken: empireDaemon::CustomResourcesTopic
      AppId: acmeApp::AppId
      Variables:
        - Name: FOO
          Value: BAR
        - Name: TEST
          Value: HERE
        - Name: DATABASE_URL
          Value: db::DatabaseURL
```

but that syntax doesn't work with stacker `PARAMETERS` and it doesn't work with `LOCAL_PARAMETERS` because they don't resolve outputs like `db::DatabaseURL`. I'm going to add support for `LOCAL_PARAMETERS` resolving outputs so we can support that.

server/cloudformation/cloudformation.go
```diff
@@ -241,6 +249,12 @@ func requiresReplacement(n, o properties) (bool, error) {
 	return a != b, nil
 }
 
+// newUser returns an empire.User that should be used by resources when making
+// requests to empire.
+func newUser() *empire.User {
+	return &empire.User{Name: "Cloudformation"}
```
ejholmes: Super anal, but I think aws uses `CloudFormation` (capital F): https://aws.amazon.com/cloudformation/

ejholmes:
👍

Issue #849, opened 2016-05-24
Add a Custom::ECSTaskDefinition resource


ejholmes:
There's a few advantages I see to doing this:

1. We could use KMS to encrypt the environment variables, then pass the cipher text to the `Custom::ECSTaskDefinition`. This means we'd no longer have plain text environment variables in the stack template (good and bad to this, obviously) and it means that we'd have less duplication, allowing us  to reduce the overall size of the template, which we've come pretty close to hitting on some services.
2. Doing this would make it easier to implement a restart key for https://github.com/remind101/empire/issues/847, so that we create a new task definition whenever the key changes.

As an example, it might look like this in the template:

```json
{
  "Type": "Custom::ECSTaskDefinition",
  "Properties": {
    "Environment": {
      "App": "<ciphertext>",
      "Process": "<ciphertext>"
    },
    "Release": "<release version>",
    "RestartKey": "<some key>"
  }
}
```

ejholmes:
Empire supports this experimentally now by setting the `ECS_TASK_DEFINITION=custom` env var: https://github.com/remind101/empire/pull/935

PR #859, opened 2016-06-02
Custom::ECSTaskDefinition resource.

ejholmes:
Fixes https://github.com/remind101/empire/issues/849

This is a spike into a fix for large templates for apps with lots of environment variables, and lots of processes defined in the Procfile. This drops the template size generated by the `TestEmpireTemplate_Large` test from 275kb, to 50kb.

Basically, we add two knew custom resources:

* **Custom::Environment**: A really simple resource that just takes some environment variables, and stores them, returning a physical id to reference that set of environment variables.
* **Custom::ECSTaskDefinition**: Pretty much identical to the AWS::ECS::TaskDefinition resource, but the `Environment` key takes a list of physical id's returned from a `Custom::Environment`.

Doing this means we don't have to duplicate the environment variables for each task definition. This also opens up the possibility of supporting encryption of environment variables, so that they aren't stored in plain text in the template.

**TODO**

* [ ] Storage? postgres? s3? dynamo? (Leaning towards postgres)
* [ ] Sort environment variables when using the custom task definition.
* [ ] Tests for new custom resources.
* [ ] Don't postfix the task definition?

### DIFF
--- a/server/cloudformation/environment.go
+++ b/server/cloudformation/environment.go
@@ -1,0 +1,41 @@
+package cloudformation
+
+import "fmt"
+
+// EnvironmentProperties are the properties provided to the
+// Custom::Environment custom resource.
+type EnvironmentProperties struct {
+	Environment map[string]string
+}
+
+// EnvironmentResource is a custom resource that takes some encrypted
+// environment variables, stores them, then returns a unique identifier to
+// represent the environment.
+type EnvironmentResource struct {
+}
+
+func (p *EnvironmentResource) Properties() interface{} {
+	return &EnvironmentProperties{}
+}
+
+func (p *EnvironmentResource) Provision(req Request) (string, interface{}, error) {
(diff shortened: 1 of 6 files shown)


server/cloudformation/environment.go
```diff
@@ -0,0 +1,41 @@
...
+	case Update:
+		id, err := p.store(properties.Environment)
+		return id, nil, err
+	default:
+		return "", nil, fmt.Errorf("%s is not supported", req.RequestType)
+	}
+}
+
+func (p *EnvironmentResource) store(env map[string]string) (string, error) {
+	return "env", nil
```
ejholmes: Still need to implement this. I'm thinking of just using a `cloudformation_environment` table in postgres to store the environment.

ejholmes:
Gonna close this for now. There may come a time when we need it, but I'd like to avoid the maintenance overhead for now.

Issue #861, opened 2016-06-06
Per process restart keys creates too many parameters for apps with lots of processes


phobologic:
While trying to unset a variable, or restart a backend with a *lot* of environment variables (127) and a *lot* of processes (40) one of the devs got this error:

```
error: error validating CloudFormation template: ValidationError:
    Template format error: Parameter count 86 is greater than max allowed 60
```

Rolling back to the old CF backend (remind101/empire:b68a836dc236a81b02ebaa48d5de5ce764752a01 # https://github.com/remind101/empire/commits/b68a836dc236a81b02ebaa48d5de5ce764752a01) seems to have fixed the issue.

Also, not sure what the best way is to dig into these issues - I looked at the last template uploaded to S3 and it looked like it only contained 43 parameters, so not sure what was going on there.

PR #862, opened 2016-06-06
Remove per process restart keys

ejholmes:
Fixes https://github.com/remind101/empire/issues/861

We only added the per process restart keys with the idea of supporting `emp restart <process>`, but this creates too many parameters for some applications that have a lot of processes.


phobologic:
👍

Issue #864, opened 2016-06-06
emp log returns error if -d isn't provided


ejholmes:
```
$ ./build/emp log -a r101-api
time: invalid duration
Usage: emp log [-a <app or remote>] [-d]
Use 'emp help log' for more information.
```

Issue #863, opened 2016-06-06
io.EOF returned from `emp log`


ejholmes:
https://github.com/remind101/empire/pull/830 introduced a regression where older versions of the CLI won't post a body, so https://github.com/remind101/empire/commit/030b7b59cab9e9b951878a4fe416bb8c3d413cff#diff-0c7d5dfcf1094dad0a17e93076c9f7c8R28 returns io.EOF.

When I upgraded the CLI, everything worked fine.

PR #865, opened 2016-06-06
Annotate log streaming errors.

ejholmes:
Added this while debugging issues with #863 and #864, which were ultimately caused elsewhere, but we should still be doing more of this to make errors more useful. `error: io.EOF` makes me sad.


mhahn:
👍

PR #866, opened 2016-06-07
We should only parse duration if it has been provided

mhahn:
Fixes https://github.com/remind101/empire/issues/864


mhahn:
These tests are passing locally, seems to be an unstable test?

ejholmes:
Yeah, I've seen that test fail occasionally. Just restarted it.

Otherwise, looks good :+1:

Issue #812, opened 2016-04-30
Switch to pkg/errors?


ejholmes:
http://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully is a good read on the issues with error handling in Go, and hits on a lot of pain points that we've had with error handling inside Empire.

We currently rely on `pkg/reporter` calls at the source of errors to get reasonable stack traces, but this sucks for multiple reasons:

1. If the error continues down the stack and there's another call to reporter.Report, that error is getting reported twice.
2. It feels wrong to be putting error reporting anywhere other than at the very top of the stack. This results in some errors getting reported, that are gracefully handled further down the stack (e.g. when we use the error returned from UpdateService to check if the service already exists).
3. Currently requires adding context.Context support, which most Go packages don't support.

If we use pkg/errors, then any errors bubbled up can be reported and still have a sane stack trace that we can actually use to debug the issue.

phobologic:
👍 +somany

ejholmes:
I'm gonna close this. `pkg/errors` has it's own set of problems, but we've been annotating errors with more information so they're easier to track down, which I think solves this for the most part.

PR #867, opened 2016-06-07
Backwards compatability with the old emp cli

mhahn:
If we get an empty body (from the old emp cli) just use the default form
values.

Fixes: https://github.com/remind101/empire/issues/863

### DIFF
--- a/server/heroku/logs.go
+++ b/server/heroku/logs.go
@@ -26,7 +26,7 @@ func (h *PostLogs) ServeHTTPContext(ctx context.Context, w http.ResponseWriter, 
 
 	var form PostLogsForm
 	if err := Decode(r, &form); err != nil {
-		if err != "EOF" {
+		if err.Error() != "EOF" {
 			return fmt.Errorf("error decoding request: %v", err)
 		}
 	}
--- a/server/heroku/logs.go
+++ b/server/heroku/logs.go
@@ -2,6 +2,7 @@ package heroku
 
 import (
 	"fmt"
+	"io"
 	"net/http"
 	"time"
 
@@ -26,7 +27,9 @@ func (h *PostLogs) ServeHTTPContext(ctx context.Context, w http.ResponseWriter, 
 
 	var form PostLogsForm
 	if err := Decode(r, &form); err != nil {
-		if err.Error() != "EOF" {
+		// We ignore the EOF error for backwards compatability with the "emp"
+		// command that doesn't support providing a duration.
+		if err != io.EOF {
 			return fmt.Errorf("error decoding request: %v", err)
 		}
 	}
--- a/server/heroku/logs.go
+++ b/server/heroku/logs.go
@@ -1,8 +1,6 @@
 package heroku
 
 import (
-	"fmt"
-	"io"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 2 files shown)


server/heroku/logs.go
```diff
@@ -20,13 +21,14 @@ type PostLogsForm struct {
 func (h *PostLogs) ServeHTTPContext(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
 	a, err := findApp(ctx, h)
 	if err != nil {
-		return err
+		return fmt.Errorf("error finding app: %v", err)
```
ejholmes: Don't think we can do this here right now because of the way [we currently handle gorm.RecordNotFound](https://github.com/remind101/empire/blob/master/server/heroku/errors.go#L66-L68).

mhahn: ah cool

ejholmes: But, we could change the `findApp` method to handle gorm.RecordNotFound, and return a more descriptive `ErrorResource` from there. e.g:

```go
func findApp(ctx context.Context, e interface {
        AppsFind(empire.AppsQuery) (*empire.App, error)
}) (*empire.App, error) {
        vars := httpx.Vars(ctx)
        name := vars["app"]
        reporter.AddContext(ctx, "app", name)

        a, err := e.AppsFind(empire.AppsQuery{Name: &name})
        if err == gorm.RecordNotFound {
                return errNotFound(fmt.Sprintf("could not find app with name %s", name))
        } else if err != nil {
                return fmt.Errorf("error finding app: %v")
        }
        return a, nil
}
```

mhahn: Thats what confuses me about the current error handling, some have been annotated and it's "safe" to just return `err` while others are `EOF` and suck.

Maybe it's worth biting the bullet and getting after: https://github.com/remind101/empire/issues/812 with a clear standard on how to handle errors?

server/heroku/logs.go
```diff
@@ -24,9 +25,10 @@ func (h *PostLogs) ServeHTTPContext(ctx context.Context, w http.ResponseWriter,
 	}
 
 	var form PostLogsForm
-
 	if err := Decode(r, &form); err != nil {
-		return err
+		if err.Error() != "EOF" {
```
ejholmes: You can change this to:

```go
if err != io.EOF
```

I'd add a comment there as well just to mention why we're ignoring io.EOF.

server/heroku/logs.go
```diff
@@ -24,9 +25,10 @@ func (h *PostLogs) ServeHTTPContext(ctx context.Context, w http.ResponseWriter,
...
+			return fmt.Errorf("error decoding request: %v", err)
```
ejholmes: It'd be nice if we did this globally across the handlers, but then you can't check for io.EOF. We could split out the Decode method so that we have a lower level DecodeRequest method that can ignore io.EOF.

e.g.:

```go
func DecodeRequest(r *http.Request, v interface{}, ignoreEOF) error {
        if err := json.NewDecoder(r.Body).Decode(v); err != nil {
                if err == io.EOF && ignoreEOF {
                        return nil
                }

                return fmt.Errorf("error decoding request body: %v", err)
        }

        return nil
}

func Decode(r *http.Request, v interface{}) error {
        return DecodeRequest(r, b, false)
}
```

Then, in this endpoint:

```go
ignoreEOF := true
if err := DecodeRequest(r, &form, ignoreEOF); err != nil {
        return err
}
```

mhahn: yea that makes sense

mhahn: i think then we're also continuing confusion around: https://github.com/remind101/empire/pull/867#discussion_r65992662, ie. hoping for underlying annotation

mhahn:
@ejholmes refactored out `DecodeRequest`

ejholmes:
Awesome 👍

PR #868, opened 2016-06-07
Make CloudFormation scheduler opt-in

ejholmes:
@phobologic @mhahn what we previously talked about, this makes the default scheduling backend the old ECS scheduler, so CloudFormation doesn't block a new release.

My thoughts are:

* 0.10.1 adds CloudFormation backend, but it is not enabled by default (people can opt into it with `--scheduler=cloudformation-migration`)
* 0.11.0 CloudFormation backend becomes the default for new apps. People should migrate existing apps to the CloudFormation backend.
* 1.0.0 CloudFormation is the default, and the old scheduler is removed.

If this looks good, I'll update https://github.com/remind101/empire/wiki/Scheduler-Deprecation-Plan and https://github.com/remind101/empire/wiki/Upgrading-to-0.11.0

### DIFF
--- a/cmd/empire/factories.go
+++ b/cmd/empire/factories.go
@@ -93,9 +93,17 @@ func newScheduler(db *empire.DB, c *cli.Context) (scheduler.Scheduler, error) {
 		return nil, err
 	}
 
-	s, err := newMigrationScheduler(db, c)
-	if err != nil {
-		return nil, err
+	var s scheduler.Scheduler
+	if c.String(FlagDefaultScheduler) == "cloudformation" {
+		s, err = newMigrationScheduler(db, c)
+		if err != nil {
+			return nil, err
+		}
+	} else {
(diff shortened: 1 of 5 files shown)


cmd/empire/factories.go
```diff
@@ -93,9 +93,17 @@ func newScheduler(db *empire.DB, c *cli.Context) (scheduler.Scheduler, error) {
 		return nil, err
 	}
 
-	s, err := newMigrationScheduler(db, c)
-	if err != nil {
-		return nil, err
+	var s scheduler.Scheduler
+	if c.String(FlagDefaultScheduler) == "cloudformation" {
```
ejholmes: Tweaked this a bit so that the default (opt out) bypasses cloudformation.MigrationScheduler as well.

phobologic:
👍 Awesome.

PR #872, opened 2016-06-11
Middleware cleanup.

ejholmes:
This fixes an annoying issue in development, where tests that exercised the http server of Empire would silence panics.

This does a couple of things:

1. Production level middleware (logging, panic recovery, etc), is only wrapped around the server in `cmd/empire`, so `server.New` returns a clean httpx.Handler that we can test against.
2. The middleware was cleaned up, so we just create a root context.Context in `package main` that has a reporter and logger embedded, then inject this wherever it's needed (http server, and cloudformation resource provisioner).
3. The empire.Empire type no longer has `Logger` and `Reporter` fields. These were only there for the purpose of passing it to server.New, but were otherwise unused by empire core.

End result is less noisy tests, panics that actually panic, and no dependency on `pkg/httpx/middleware` (which will most likely become irrelevant with Go 1.7) hooray!

### DIFF
--- a/server/heroku/heroku.go
+++ b/server/heroku/heroku.go
@@ -13,7 +13,6 @@ import (
 	"github.com/remind101/empire/pkg/heroku"
 	"github.com/remind101/empire/server/auth"
 	"github.com/remind101/pkg/httpx"
-	"github.com/remind101/pkg/httpx/middleware"
 )
 
 // The Accept header that controls the api version. See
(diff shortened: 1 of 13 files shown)


server/heroku/heroku.go
```diff
@@ -74,13 +74,16 @@ func New(e *empire.Empire, authenticator auth.Authenticator) httpx.Handler {
...
-	}
-
 	api := Authenticate(r, authenticator)
 
-	return middleware.HandleError(api, errorHandler)
+	return httpx.HandlerFunc(func(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+		err := api.ServeHTTPContext(ctx, w, r)
+		if err != nil {
+			Error(w, err, http.StatusInternalServerError)
+			reporter.Report(ctx, err)
```
ejholmes: Thinking of also making this change. Reporting an error here won't give a very useful stack trace, but it's still good to see that an error was generated, and what the error message was.

mhahn:
👍

Issue #697, opened 2015-12-14
Behavior of `emp restart` is confusing


ejholmes:
Currently `emp restart` has the following behavior:

* `emp restart` will do a rolling restart of all processes. The implementation is weird since we perform a new release, when we should have a `Restart` method on the Scheduler interface.
* `emp restart <ptype>` returns an error `InvalidParameterException: taskId shorter than 36`.
* `emp restart <ecs taskid>` kills the ECS task.

With that said, the behavior we probably want is:

* `emp restart` to behave as above, but fix implementation details.
* `emp restart <ptype>` should do a rolling restart of the ECS service. Implementation should be pretty easy if we fix above.
* `emp restart <ecs taskid>` I'm not sure we actually want to keep this. It's almost never used and can be potentially dangerous since it doesn't do a rolling restart.

/cc @phobologic since you were talking about fixing this.

phobologic:
Yeah, that's actually where I was thinking we'd go with this.  I had the same concerns about the taskID method, then got pulled off on other stuff.

PR #873, opened 2016-06-12
Issue 844

mhahn:
Need to test this with a real app, but this is what you and i discussed @ejholmes, removing "Scale" and just using "Release".


mhahn:
Tested that this now submits 1 update to CF.

mhahn:
Logs look like:
```
server_1  | mhahn scaled `web` on acme-inc from 2(1X) to 1(1X)
server_1  | mhahn scaled `worker` on acme-inc from 2(1X) to 1(1X)
```

we could also put it on one line, something like
```mhahn scaled `web` from 2(1X) to 1(1X) and `worker` from 2(1X) to 1(1X) on acme-inc```

ejholmes:
This looks great to me 👍 .

One thing that I was thinking, right now in the CloudFormation backend, [we always set the RestartKey when submitting a new release](https://github.com/remind101/empire/blob/master/scheduler/cloudformation/cloudformation.go#L249). The end goal is to have a `Restart` method on the Scheduler interface, which would be the only place where we'd ever change that value.

I don't think it's a big deal right now, it just means that even with the CloudFormation backend, a new submit will always result in all new task definitions getting created and a service restarting. Once we remove that, then submitting an identical release should result in no updates.

mhahn:
I'm going to start on: https://github.com/remind101/empire/issues/697, I'll address the above within that

PR #876, opened 2016-06-14
Don't set DNS parameter unless explicitly told to.

ejholmes:
This can be handy in some situations where you want to disable DNS, but still be able to make stack updates through Empire. With this change, Empire will only explicitly set the DNS parameter during step1 of the migration. Otherwise, it will use the previous parameter value.


ejholmes:
Specifically in our case, I have an existing cloudformation stack that I want to switch to, so I want to be able to switch the backend in Empire to cloudformation, deploy a new version so the stack gets updated and scaled up, then manually swap DNS over. This change makes that easy to do.

mhahn:
👍

PR #877, opened 2016-06-14
Switch default ECS service resource to Custom::ECSService.

ejholmes:
This has been working pretty well on the apps that we've enabled it on, and the overall user experience will be better (especially in cases where rollback is necessary).

I think this should be merged before we release 0.11.0, but I don't plan on merging this until we've manually switched our applications over to Custom::ECSService, since it potentially incurs some downtime when the old ECS services are destroyed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 **Improvements**
 
 * The Custom::ECSService custom resource now waits for newly created ECS services to stabilize [#878](https://github.com/remind101/empire/pull/878)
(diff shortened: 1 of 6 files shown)


ejholmes:
This can probably be safely merged now that https://github.com/remind101/empire/pull/878 was merged, but I'll try doing it manually on a couple apps in production to make sure there's no downtime.

mhahn:
For the record, also moved prod apps to this with 0 downtime

mhahn:
👍

PR #878, opened 2016-06-14
Wait for newly created services to stabilize.

ejholmes:
I think it might be good to wait for services to stabilize _only_ when they're being created. Couple reasons:

1. It'll make [this change](https://github.com/remind101/empire/pull/877) easier (no downtime).
2. In cases where a replacement is required, it will result in no downtime.
3. It shouldn't effect the 99% case where a service is just getting updated with a new task definition.


server/cloudformation/ecs.go
```diff
@@ -117,7 +118,17 @@ func (p *ECSServiceResource) create(properties *ECSServiceProperties) (string, e
 		return "", fmt.Errorf("error creating service: %v", err)
 	}
 
-	return *resp.Service.ServiceArn, nil
+	arn := resp.Service.ServiceArn
+	if err := p.ecs.WaitUntilServicesStable(&ecs.DescribeServicesInput{
```
ejholmes: May want to also set a hard timeout on this. Like 5-10 minutes.

ejholmes: Turns out, the waiter will check up to 40 times, with a 15 second delay between checks, so this will wait up to 10 minutes, which seems about right.

https://github.com/aws/aws-sdk-go/blob/2e1a49c71fb7d8f6b7e8cca2259ed9b68d9b97ce/service/ecs/waiters.go#L41-L42

mhahn:
👍

ejholmes:
Since this can potentially block the sqs polling loop, provisioning should happen in a goroutine.

ejholmes:
Ok, made some changes to this after testing in my environment:

1. Updated go-jmespath so that WaitUntilServicesStable actually works.
2. Provisioning now happens in a goroutine, and provisioners are given 20 minutes to do their work. Should never actually hit that, but it'll keep goroutines from growing unbounded.

ejholmes:
This is working well in staging. Gonna merge.

PR #882, opened 2016-06-15
Issue 697

mhahn:
Fixes https://github.com/remind101/empire/issues/697

- We only bump the `RestartKey` within the cloudformation scheduler on `emp restart`
- If the app can support a restart key per process (less than 25 processes), `emp restart <process type>` will do a rolling restart of the specified process type.
- `emp restart <task-id>` works as it did before, killing the process. We had talked about pulling this out into a `kill` method, but since the task definition can still exist and bring up another process in its place it really is a restart so I just left it alone.

Todos:
- [ ] If a restart key per process is not enabled, we should throw an error if you specify a process type
- [ ] Update the event message if process type is specified

### DIFF
--- a/scheduler/cloudformation/cloudformation.go
+++ b/scheduler/cloudformation/cloudformation.go
@@ -169,6 +169,10 @@ type SubmitOptions struct {
 	NoDNS bool
 }
 
+func (s *Scheduler) Restart(ctx context.Context, app *scheduler.App) error {
+	return s.Submit(ctx, app)
+}
+
 // SubmitWithOptions submits (or updates) the CloudFormation stack for the app.
 func (s *Scheduler) SubmitWithOptions(ctx context.Context, app *scheduler.App, opts SubmitOptions) error {
 	if opts.Done == nil {
--- a/scheduler/cloudformation/cloudformation.go
+++ b/scheduler/cloudformation/cloudformation.go
@@ -168,12 +168,16 @@ type SubmitOptions struct {
 	// migrating to this scheduler
 	NoDNS bool
 
-	// When true, will generate a new RestartKey.
-	Restart bool
+	// When provided, will generate a new RestartKey for the target.
+	RestartTarget string
 }
 
-func (s *Scheduler) Restart(ctx context.Context, app *scheduler.App) error {
-	return s.SubmitWithOptions(ctx, app, SubmitOptions{Restart: true})
+func (s *Scheduler) Restart(ctx context.Context, app *scheduler.App, p string) error {
+	target := p
+	if target == "" {
+		target = "all"
+	}
+	return s.SubmitWithOptions(ctx, app, SubmitOptions{RestartTarget: target})
 }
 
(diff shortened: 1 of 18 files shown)


scheduler/cloudformation/cloudformation.go
```diff
@@ -167,6 +167,17 @@ type SubmitOptions struct {
 	// When true, does not make any changes to DNS. This is only used when
 	// migrating to this scheduler
 	NoDNS bool
+
+	// When provided, will generate a new RestartKey for the target.
+	RestartTarget string
+}
+
+func (s *Scheduler) Restart(ctx context.Context, app *scheduler.App, p string) error {
```
ejholmes: I think the endgoal, once we can remove the old ECS scheduler, would be to change this method signature to `Restart(context.Context, app string, process string) error`, since we won't need to submit a new template to cloudformation, and can just update a parameter.

For that reason, I think we should still make this just update parameters, instead of doing a full Submit and generating a new template.

ejholmes: Also, to keep things simple to start, and because of the parameter limit on cloudformation, we may wanna just start with a `Restart(context.Context, *scheduler.App)` method and then add process restarts later.

scheduler/cloudformation/cloudformation.go
```diff
@@ -235,34 +246,70 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
...
+	var stackParameters []*cloudformation.Parameter
+	var create bool
+	resp, err := s.cloudformation.DescribeStacks(&cloudformation.DescribeStacksInput{
+		StackName: aws.String(stackName),
+	})
+	if err, ok := err.(awserr.Error); ok && err.Message() == fmt.Sprintf("Stack with id %s does not exist", stackName) {
+		create = true
+	} else if err == nil {
+		create = false
+		stackParameters = resp.Stacks[0].Parameters
```
ejholmes: This is probably a dangerous place to do this. If this is an update, it might get queued behind an in progress update, at which point these parameters may have been invalidated from that update. The parameters returned from describing a stack should only really be considered valid while the process holds the [stackOperation lock](https://github.com/mhahn/empire/blob/eb40679ae5416bc9d2c4e636f3485b185c44d0aa/scheduler/cloudformation/cloudformation.go#L393-L397).

When the stack update eventually gets executed, [then it will describe the stack and merge in the existing stack parameters there](https://github.com/mhahn/empire/blob/eb40679ae5416bc9d2c4e636f3485b185c44d0aa/scheduler/cloudformation/cloudformation.go#L455-L486).

mhahn: ah cool. i was initially going to go with UseExistingTemplate, but was worried about having to ensure we submit all the parameters.

mhahn:
Closing in favor of: https://github.com/remind101/empire/pull/883. I left out the process restart, we can come back to that after.

PR #883, opened 2016-06-16
Issue 697 v2

mhahn:
Addresses the issue of creating a new restart key each time we submit the cloudformation template.

### DIFF
--- a/scheduler/cloudformation/cloudformation.go
+++ b/scheduler/cloudformation/cloudformation.go
@@ -169,6 +169,10 @@ type SubmitOptions struct {
 	NoDNS bool
 }
 
+func (s *Scheduler) Restart(ctx context.Context, app *scheduler.App) error {
+	return s.Submit(ctx, app)
+}
+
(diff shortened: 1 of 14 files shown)


scheduler/cloudformation/cloudformation.go
```diff
@@ -189,6 +189,38 @@ func (s *Scheduler) SubmitWithOptions(ctx context.Context, app *scheduler.App, o
...
+		return err
+	}
+
+	return s.updateStack(&cloudformation.UpdateStackInput{
+		StackName:           aws.String(stackName),
+		UsePreviousTemplate: aws.Bool(true),
+		Parameters: []*cloudformation.Parameter{
+			{
+				ParameterKey:   aws.String(restartParameter),
+				ParameterValue: aws.String(newUUID()),
```
phobologic: What if we used a timestamp here?  Maybe as well as a UUID?  Could be useful when debugging?

ejholmes:
Awesome. Looks great to me 👍

mhahn:
closing for the second time in favor of an updated branch 🙈

PR #884, opened 2016-06-17
Prevent possible duplicate ECS services.

ejholmes:
After https://github.com/remind101/empire/pull/878 I discovered an issue where duplicate ECS services got created from the same CloudFormation update. This does 2 things to prevent that from happening:

1. I'd recommend setting a 30 minute visibility timeout on the sqs queue. We currently give custom provisioners 20 minutes to complete their work, so this should be more than enough time.
2. We set the ClientToken property to ensure 100% idempotency of the CreateService call.


ejholmes:
Ok, after testing locally, I discovered that the ClientToken is linked to the service name, which means adding the ClientToken had no effect, since we were generating a random, non-deterministic postfix when creating a service.

To fix this, I changed it so that the postfix is generating deterministically based on the properties of the ECS service that would require a replacement, if they were to change. Functionally, this works well and prevents duplicate ECS services from being created, but I'm not a fan of the length and format of the postfix (based64 encoded sha1):



If anyone has any suggestions on how to shorten this, but keep it sufficiently unique and avoid collisions, I'm all ears. Maybe truncating the sha would be fine?

ejholmes:
Maybe it's fine the way it is. My concerns are mostly cosmetic =\.

It does push things closer to the limit of the size of the service name (255) but you still have 228 characters, which is a pretty massive app + process name.

ejholmes:
Ok, really happy with this now. The postfix is now the base62 encoded value of the [Sum64](https://golang.org/pkg/hash/#Hash64) of the hash function, which means the maximum size of the string is 11 characters, and it's alphanumeric.



I also pulled in https://github.com/mitchellh/hashstructure, which is a really nice library for generating a hash of a go struct.

pkg/base62/base62.go
```diff
@@ -0,0 +1,29 @@
+// Package base62 implements conversion to and from base62. Useful for url shorteners.
+package base62
```
ejholmes: This was stripped from an internal project.

server/cloudformation/ecs.go
```diff
@@ -32,19 +32,25 @@ type LoadBalancer struct {
 type ECSServiceProperties struct {
 	ServiceName    *string
 	Cluster        *string
-	DesiredCount   *IntValue
-	LoadBalancers  []LoadBalancer
+	DesiredCount   *IntValue      `hash:"ignore"`
+	LoadBalancers  []LoadBalancer `hash:"set"`
 	Role           *string
-	TaskDefinition *string
+	TaskDefinition *string `hash:"ignore"`
```
ejholmes: This is pretty nifty. Fields that can be updated, we just add this tag.

server/cloudformation/ecs.go
```diff
@@ -32,19 +32,25 @@ type LoadBalancer struct {
 type ECSServiceProperties struct {
 	ServiceName    *string
 	Cluster        *string
-	DesiredCount   *IntValue
-	LoadBalancers  []LoadBalancer
+	DesiredCount   *IntValue      `hash:"ignore"`
+	LoadBalancers  []LoadBalancer `hash:"set"`
```
ejholmes: This ensures that the ordering of the load balancers doesn't cause a replacement.

server/cloudformation/ecs.go
```diff
@@ -172,39 +190,13 @@ func (p *ECSServiceResource) delete(ctx context.Context, service, cluster *strin
...
-func postfix() string {
-	n := 12
-	b := make([]rune, n)
-	for i := range b {
-		b[i] = letters[rand.Intn(len(letters))]
+// It's important that the postfix we append to the service name
+// is deterministic based on the non updateable fields of the
+// service, otherwise ClientToken has no effect on idempotency.
+func postfix(p *ECSServiceProperties) (string, error) {
+	h, err := p.Hash()
```
ejholmes: Something to think of, it would be sufficient to just hash the request id instead of the properties.

ejholmes: Thinking about this more, hashing the request id would be better. If we hash the props, you could potentially get into a scenario where a service gets re-created with the same name as a service that was previously deleted. We've seen a lot of weird behavior when that happens (complaining about the load balancer not existing, saying it's in a draining state, etc), so best to avoid it.

ejholmes:
Ok, simplified this a bunch so we just generate a hash of the request id, then use that as a postfix for the service name.

Think this is good to review now.

server/cloudformation/ecs_test.go
```diff
@@ -292,7 +300,8 @@ func TestRequiresReplacement(t *testing.T) {
 		},
 	}
 
-	for _, tt := range tests {
+	for i, tt := range tests {
+		t.Log(i)
```
mhahn: leftover?

mhahn:
👍

docs/cloudformation.json
```diff
@@ -870,7 +870,10 @@
     },
 
     "CustomResourcesQueue": {
-      "Type": "AWS::SQS::Queue"
+      "Type": "AWS::SQS::Queue",
+      "Properties": {
+        "VisibilityTimeout": "1800"
```
phobologic: This seems really long - you handle de-duping requests here as well, right?  Otherwise we could see things being re-submitted 30 minutes later after a crash or something.  In the case where de-duping *is* handled, what's the idea behind making this so long?  Would 5 minutes be sufficient?

ejholmes: Yeah good point. Probably the best thing to do would be to leave the default, and then use the [ChangeMessageVisibility](http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ChangeMessageVisibility.html) api call to increase it when needed.

I'll revert the current change for now, since the de-duping is sufficient to fix the current issue.

PR #887, opened 2016-06-20
Debug migration scheduler

mhahn:
I was running into errors with the new migration scheduler.

Part of it was the error message being swallowed. I also added some error annotations.

### DIFF
--- a/cmd/empire/factories.go
+++ b/cmd/empire/factories.go
@@ -123,7 +123,7 @@ func newScheduler(db *empire.DB, c *cli.Context) (scheduler.Scheduler, error) {
 	}
 
 	if err != nil {
-		return nil, fmt.Errorf("error creating scheduler: %v", err)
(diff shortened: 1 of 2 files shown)


cmd/empire/factories.go
```diff
@@ -122,6 +122,10 @@ func newScheduler(db *empire.DB, c *cli.Context) (scheduler.Scheduler, error) {
 		return nil, fmt.Errorf("unknown scheduler: %s", c.String(FlagScheduler))
 	}
 
+	if err != nil {
+		return nil, fmt.Errorf("error creating scheduler: %v", err)
```
ejholmes: Minor, maybe `s/creating/initializing/`

ejholmes: Also, maybe include the name of the scheduler:

```go
fmt.Errorf("failed to initialize %s scheduler: %v", c.String(FlagScheduler), err)
```

ejholmes:
nice find 👍

PR #889, opened 2016-06-22
Better Godocs

ejholmes:
This is just some general house keeping, mostly geared at improving the [generated package documentation](https://godoc.org/github.com/remind101/empire)

This looks big, but it's primarily cosmetic; there shouldn't be any functional/behavior changes. In a nutshell:

* Various types that were previously unnecessarily exported in `package empire` are now private (e.g. `Scope`, `Migrations`, etc).
* `package empire` has some improved top level documentation that describes the core internal Empire API.
* Some inconsistently named structs were renamed (DeploymentsCreateOpts -> DeployOpts).
* Some exported types/functions got some example tests added, which show up in godocs.

If you wanna see the new godocs, you can run `godoc -http=":6060"` and visit <http://localhost:6060/pkg/github.com/remind101/empire>.

### DIFF
--- a/access_tokens.go
+++ b/access_tokens.go
@@ -22,13 +22,13 @@ func (t *AccessToken) IsValid() error {
 }
 
 type accessTokensService struct {
-	Secret []byte // Secret used to sign jwt tokens.
+	*Empire
 }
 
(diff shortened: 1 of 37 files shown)


mhahn:
👍 looks great.

PR #892, opened 2016-06-23
Ensure that stack operation lock is always unlocked

ejholmes:
I discovered that there are certain cases where the stack operation lock would not get unlocked. If, for example, an error was returned from calling UpdateStack, we'd never reach the call to l.Unlock().

This change makes it so that `performStackOperation` is synchronous throughout the stack operation, which means we can just `defer l.Unlock()` immediately after locking.

### DIFF
--- a/scheduler/cloudformation/cloudformation.go
+++ b/scheduler/cloudformation/cloudformation.go
@@ -131,6 +131,10 @@ type Scheduler struct {
 	// Any additional tags to add to stacks.
 	Tags []*cloudformation.Tag
 
+	// When true, alls to Submit won't return until the stack has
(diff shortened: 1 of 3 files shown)


scheduler/cloudformation/cloudformation.go
```diff
@@ -131,6 +131,10 @@ type Scheduler struct {
 	// Any additional tags to add to stacks.
 	Tags []*cloudformation.Tag
 
+	// When true, alls to Submit won't return until the stack has
```
mhahn: s/alls/calls

mhahn:
👍 looks great

Issue #891, opened 2016-06-22
Fail to boot if DB schema is incompatible


ejholmes:
https://github.com/remind101/empire/issues/890 For background.

This would help in upgrade scenarios, so that a previous version of Empire wouldn't boot against a newer schema, and vice versa.

ejholmes:
I think this could just go in the [db health check](https://github.com/remind101/empire/blob/master/db.go#L78)

PR #893, opened 2016-06-23
Add schema version check to db health check.

ejholmes:
Closes https://github.com/remind101/empire/issues/891

This adds a check to the db health check that checks that the schema version matches what the running version of Empire expects. This means:

1. If you're in the process of upgrading to a new version of Empire, once the migrations are complete, older versions of Empire will start failing health checks and get taken out of the load balancer (assuming you have Empire behind a load balancer and are using the `/health` endpoint).
2. If you try to start `empire server` with an incompatible database schema, you'll get an error message:

    ```console
$ ./build/empire server
2016/06/23 12:19:23 Streaming logs are disabled
2016/06/23 12:19:23 Using Stdout run logs backend:
2016/06/23 12:19:23 expected database schema to be at version 17, but was 18. You can resolve this error by running the migrations with `empire migrate` or with the `--automigrate` flag
````


phobologic:
awesome 👍

mhahn:
👍

PR #898, opened 2016-06-24
Use existing parameter values, even for new templates

ejholmes:
https://github.com/remind101/empire/pull/848 introduced a change where we would only merge in the existing parameter values from the stack, if we're updating the stack using the existing stack template.

https://github.com/remind101/empire/pull/876 introduced a change where we actually wan't to always use the previous value , even if we're updating to a new stack template

Without this change if you manually set the DNS parameter on the stack to false, then deploy a new version, it will still get changed back to true, because CloudFormation will use the default value, because we didn't provide the previous value for the parameter.

The net effect  of this change is that we always merge in the existing parameter values from the stack, but only if they're defined in the new stack template.

### DIFF
--- a/scheduler/cloudformation/cloudformation.go
+++ b/scheduler/cloudformation/cloudformation.go
@@ -205,32 +205,10 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App, 
 		return err
 	}
 
-	buf := new(bytes.Buffer)
-	if err := s.Template.Execute(buf, app); err != nil {
+	t, err := s.createTemplate(ctx, app)
+	if err != nil {
 		return err
 	}
-
-	key := fmt.Sprintf("%s/%s/%x", app.Name, app.ID, sha1.Sum(buf.Bytes()))
-
-	_, err = s.s3.PutObject(&s3.PutObjectInput{
-		Bucket:      aws.String(s.Bucket),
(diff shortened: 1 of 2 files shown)


scheduler/cloudformation/cloudformation.go
```diff
@@ -416,41 +454,23 @@ func (s *Scheduler) waitUntilStackOperationComplete(lock *pglock.AdvisoryLock, w
...
-					ParameterKey:     aws.String(k),
-					UsePreviousValue: aws.Bool(true),
-				})
-			}
-		}
+	i := &cloudformation.UpdateStackInput{
+		StackName:  input.StackName,
+		Parameters: updateParameters(input.Parameters, stack, input.Template),
+	}
+	if input.Template != nil {
```
mhahn: when is this ever `nil`?

mhahn: seems like: https://github.com/remind101/empire/pull/898/files#diff-67e3f5de813e6b6e24cb66d3278cff99R208 will only return a `nil` template on error, otherwise you should always have this?

mhahn:
👍

scheduler/cloudformation/cloudformation.go
ejholmes: This will be nil when we go ahead with https://github.com/remind101/empire/pull/883.

PR #900, opened 2016-06-24
Fix stack update deadlock.

ejholmes:
https://github.com/remind101/empire/pull/892 introduced a bug where, when we would hit the `lockWait` timeout, the call to update the stack would deadlock (trying to send on a channel that had no consumers), preventing the lock from ever getting unlocked. The fix was to simply use a buffered `submitted` channel. Channels are fun...

Basically didn't have any coverage for this case, so I added some tests for the different timeout scenarios.


mhahn:
👍

PR #901, opened 2016-06-24
Enable race detector (and fix flaky test)

ejholmes:
I could have sworn we enabled this a long time ago, but I guess not. Anyway, it caught the concurrent deployment test, which has been flaky in the past.


mhahn:
👍

PR #902, opened 2016-06-24
Cache task definitions

ejholmes:
This is a small performance optimization that takes into account the fact that task definitions can essentially be treated as immutable, when keying off the ARN.

Describing task definitions is a pretty expensive call, both in time (it's N+1, and slow) and in throttling. This should help quite a bit with `emp ps` performance, once the cache is filled.

**TODO**

* [x] Tests
* [x] Don't cache if the TaskDefinition is not an ARN.


phobologic:
Cool idea.

mhahn:
:

Issue #899, opened 2016-06-24
Procfile extraction doesn't work on Docker 1.12


ejholmes:
Since I upgraded my local Docker, deploying remind101/acme-inc results in the Procfile not being found:

```
Procfile not found: No such container: c9ffa2b732440446955086c44bb6256607c2b78e6d33c37c882588db13e32f29
```

ejholmes:
This happens [here](https://github.com/remind101/empire/blob/79a76ce7c86706f388df8211b0d64882dc7700af/extractor.go#L155). Dunno why it's able to inspect the container fine, but then returns this error when we try to copy.

ejholmes:
Woops, this got closed accidentally.

PR #910, opened 2016-06-30
Upgrade Docker client

ejholmes:
This is some prep work to fix https://github.com/remind101/empire/issues/899 and add Docker 1.12 compatibility.

This upgrades the fsouza/go-dockerclient dependency to the latest version, and also its Docker related dependencies (stuff from docker/docker/pkg) to Docker 1.11.2 versions.

Commands that exercise the Docker client (emp deploy, emp run) seem to work fine in my dev environment, but worth running this on staging for a bit to see if any issues crop up (it would be nice if we had some integration tests against different versions of Docker).

I separated this into two commits.

1. First commit is committing the new deps to vendor. For reference [here](https://gist.github.com/ejholmes/dbd990d67433392ef0955da4de7a46ae) is the name of the files that changed.
2. Second commit includes some minor changes to accommodate API changes in the new deps, and the new vendor/vendor.json.

Probably best to only review  no. 2 unless you want your browser to explode.


phobologic:
👍

ejholmes:
Seems to be working well in our staging environment so far. I'll merge for now.

PR #914, opened 2016-07-02
Stop

ejholmes:
It's pretty confusing that you use the `emp restart` command to stop a single task/process. This adds an `emp stop` command to help make this more intuitive.

Also does a little cleanup to move the "friendly" name into server/heroku and adds the ID of the task to the API response.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * The Custom::ECSService custom resource now waits for newly created ECS services to stabilize [#878](https://github.com/remind101/empire/pull/878)
 * The CloudFormation backend now uses the Custom::ECSService resource instead of AWS::ECS::Service, by default [#877](https://github.com/remind101/empire/pull/877)
 * The database schema version is now checked at boot, as well as in the http health checks. [#893](https://github.com/remind101/empire/pull/893)
+* An `emp stop` command has been added. [#914](https://github.com/remind101/empire/pull/914)
 
 **Bugs**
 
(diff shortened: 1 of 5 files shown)


phobologic:
This just restarts the process, right?  I think the use of 'stop' is a little confusing - people might think that it leads to the process being stopped forever (which, technically is true, but the fact that ECS will bring up a new one in it's place makes the term stop a bit harder to understand).

ejholmes:
@phobologic granted ECS will spin up a new container to replace it, it's essentially equivalent to issuing a `docker stop` on the container. If it's, for example, a detached process, then it's not actually going to get restarted, which is why I think `emp restart` is confusing when given a task id.

mhahn:
i think this would be better if we only let you stop processes that ECS won't start up again.

the current restart logic is confusing, because it won't restart the detached process. i could see this being confusing if i try and `stop` a web task but it keeps coming back up.

reading the docs of the stop function i would think there is a bug.

phobologic:
Yeah, I see both sides of this.  Maybe if it printed out that the process was going to be restarted because it wasn't a `run` process?

ejholmes:
Maybe we just need more docs around the behavior of restarting and stopping processes:

1. `emp restart -a <app>` triggers a rolling restart of all ECS services. One-off processes and scheduled tasks are not restarted.
2. `emp restart <process> -a <app>` triggers a rolling restart of the ECS service for the given process (not supported yet).
3. `emp stop <task id> -a <app>` stops an ECS task or container. If the task was part of an ECS service, the task maybe be replaced.

We could put this in the `emp help` docs.

mhahn:
I agree that more docs are a good first step. If this becomes an issue, we can make it more complicated

ejholmes:
Gonna close this one for the time being.

PR #916, opened 2016-07-06
Deployment notifications

mhahn:
Still a WIP but I wanted to publish this for discussion.

Sample output
```
master: Pulling from remind101/acme-inc

420890c9e918: Already exists
a3ed95caeb02: Already exists
1e14aab0083d: Already exists
4393485f2bb3: Already exists
a2a051cf14f7: Already exists
1f9177248efa: Already exists
Digest: sha256:d05682347b56191a950f55ab06f11a6c6e34041921af32281b01f4a6a9a0f75f
Status: Created new release v64 for acme-inc
Status: Updating stack "acme-inc"
Status: Stack updated
Status: Deployment in progress for web
Status: Deployment in progress for worker
Status: Deployment stable for web
Status: Deployment stable for worker
Status: Closing event stream for release v64
```

There are issues with these tests: github.com/remind101/empire/tests/api. Relying on the Done channel to be closed via the EventsChan receiving an `Event{Close: true}` is really brittle.

### DIFF
--- a/deployments.go
+++ b/deployments.go
@@ -57,7 +57,7 @@ func (s *deployerService) deploy(ctx context.Context, db *gorm.DB, opts DeployOp
 		Config:      config,
 		Slug:        slug,
 		Description: desc,
-	})
+	}, opts.Events)
 
 	return r, err
 }
@@ -77,6 +77,34 @@ func (s *deployerService) Deploy(ctx context.Context, db *gorm.DB, opts DeployOp
 	if err := json.NewEncoder(opts.Output).Encode(&msg); err != nil {
 		return r, err
 	}
-
+	go s.streamEvents(opts)
 	return r, err
+}
+
+func (s *deployerService) streamEvents(opts DeployOpts) {
+	for event := range opts.Events {
+		var msg string
+		if event.Error != nil {
+			msg = fmt.Sprintf("Error: %s", event.Error.Error())
+		} else {
+			if event.Message != "" {
+				msg = fmt.Sprintf("Status: %s", event.Message)
+			}
+		}
+
+		if msg != "" {
+			m := jsonmessage.JSONMessage{Status: msg}
+			if err := json.NewEncoder(opts.Output).Encode(&m); err != nil {
+				close(opts.Events)
+				opts.Done <- err
+				return
+			}
+		}
+
+		if event.Error != nil || event.Close {
+			close(opts.Events)
+			opts.Done <- event.Error
+			return
+		}
+	}
 }
--- a/scheduler/cloudformation/cloudformation.go
+++ b/scheduler/cloudformation/cloudformation.go
@@ -342,6 +342,7 @@ func (s *Scheduler) waitForDeploymentToStabilize(ctx context.Context, app *sched
 		serviceIDs = append(serviceIDs, aws.String(id))
 	}
 
+	dedupe := make(map[string]bool)
 	check := func() bool {
 		status, err := s.ecs.DescribeServices(&ecs.DescribeServicesInput{
 			Cluster:  aws.String(s.Cluster),
@@ -382,8 +383,11 @@ func (s *Scheduler) waitForDeploymentToStabilize(ctx context.Context, app *sched
 			} else {
 				status = "superseded by newer release"
 			}
-			events <- scheduler.Event{
-				Message: fmt.Sprintf("Deployment %s for %s", status, name),
+			message := fmt.Sprintf("Deployment %s for %s", status, name)
+			d := fmt.Sprintf("%s:%s", *service.ServiceArn, message)
+			if _, ok := dedupe[d]; !ok {
+				events <- scheduler.Event{Message: message}
+				dedupe[d] = true
 			}
 			complete := !primary || primary && stable
 			if complete {
--- a/scheduler/cloudformation/cloudformation.go
+++ b/scheduler/cloudformation/cloudformation.go
@@ -99,6 +99,7 @@ type ecsClient interface {
 	RegisterTaskDefinition(*ecs.RegisterTaskDefinitionInput) (*ecs.RegisterTaskDefinitionOutput, error)
 	StopTask(*ecs.StopTaskInput) (*ecs.StopTaskOutput, error)
 	UpdateService(*ecs.UpdateServiceInput) (*ecs.UpdateServiceOutput, error)
+	DescribeServices(*ecs.DescribeServicesInput) (*ecs.DescribeServicesOutput, error)
 }
 
 // s3Client duck types the s3.S3 interface that we use.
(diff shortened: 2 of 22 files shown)


scheduler/cloudformation/cloudformation.go
```diff
@@ -274,9 +277,138 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
 		}
 	}
 
+	go s.waitForDeploymentToStabilize(ctx, app, stackName, events, done)
 	return nil
 }
 
+func (s *Scheduler) waitForDeploymentToStabilize(ctx context.Context, app *scheduler.App, name string, events scheduler.EventChan, done chan error) {
+	if events == nil {
```
mhahn: if this hasn't been provided, we should probably just exit

scheduler/cloudformation/cloudformation.go
```diff
@@ -274,9 +277,138 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
 		}
 	}
 
+	go s.waitForDeploymentToStabilize(ctx, app, stackName, events, done)
```
mhahn: this needs to handle if `wait` is specified

scheduler/cloudformation/cloudformation.go
```diff
@@ -274,9 +277,138 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
...
+	if events == nil {
+		events = make(scheduler.EventChan)
+	}
+
+	events <- scheduler.Event{Message: fmt.Sprintf("Updating stack \"%s\"", name)}
+	if err := <-done; err != nil {
+		events <- scheduler.Event{Error: err}
+		return
+	}
+	events <- scheduler.Event{Message: "Stack updated"}
```
mhahn: Since we don't return an error if the lock is canceled this outputs "Stack updated" if it hasn't actually updated but has just been superseded

deployments.go
```diff
@@ -77,6 +77,34 @@ func (s *deployerService) Deploy(ctx context.Context, db *gorm.DB, opts DeployOp
...
+		} else {
+			if event.Message != "" {
+				msg = fmt.Sprintf("Status: %s", event.Message)
+			}
+		}
+
+		if msg != "" {
+			m := jsonmessage.JSONMessage{Status: msg}
+			if err := json.NewEncoder(opts.Output).Encode(&m); err != nil {
+				close(opts.Events)
```
ejholmes: Closing a channel that has outside consumers is potentially dangerous. If the consumer happens to send on this channel, after it's been closed, Go will panic: https://play.golang.org/p/YvdnnAzBSw

If we use a channel based API, it should probably be up to the consumer to close the channel, when it's done producing.

deployments.go
```diff
@@ -77,6 +77,34 @@ func (s *deployerService) Deploy(ctx context.Context, db *gorm.DB, opts DeployOp
 	if err := json.NewEncoder(opts.Output).Encode(&msg); err != nil {
 		return r, err
 	}
-
+	go s.streamEvents(opts)
```
ejholmes: I think the io.Writer will be closed after this method returns, so we'd probably want to block here until the events channel is closed.

scheduler/cloudformation/cloudformation.go
ejholmes: Since this method does things like describing the stack, it should only be operating while the stack operation lock is held, otherwise multiple concurrent `Submits` will produce the wrong results.

ejholmes: I think the correct place to do this, would be immediately after [this call](https://github.com/mhahn/empire/blob/c1ea348005d0c4eb832858810a87cf7926295118/scheduler/cloudformation/cloudformation.go#L572), since the stack operation lock is held, and the stack has been updated/created successfully at that point.

scheduler/scheduler.go
```diff
@@ -118,7 +118,7 @@ type Scheduler interface {
 	Runner
 
 	// Submit submits an app, creating it or updating it as necessary.
-	Submit(context.Context, *App) error
+	Submit(context.Context, *App, EventChan) error
```
ejholmes: The more I think about it, the more I think an interface here would be better, and prevent easy mistakes with channels from happening.

My first thought, would be an interface like this:

```go
type Events interface {
        // Publish an event to the event stream.
        Publish(Event)

        // Signals that no more events will be published.
        Done(error)
}
```

We can then implement a channel based implementation of this, so that publishers don't have to worry about blocking:

```go
import "sync"

type events struct {
	sync.Mutex
	done bool
	err  error
	ch   chan Event
}

func newEvents() *events {
	return &events{ch: make(chan Event, 100)}
}

func (e *events) Publish(event Event) {
	e.Lock()
	defer e.Unlock()

	if done {
		// Or just warn.
		panic("Publish called on finalized event stream")
	}

	e.publish(event)
}

func (e *events) publish(event Event) {
	select {
	case e.ch <- event:
	default:
		// Drop
	}
}

func (e *events) Subscribe() <-chan Event {
	return e.ch
}

func (e *events) Done(err error) {
	e.Lock()
	defer e.Unlock()

	e.done = true
	e.err = err
}

func (e *events) Err() error {
	return e.err
}
```

So then, submitting and consuming the event stream would look like this:

```go
events := scheduler.NewEvents()
s.Submit(ctx, app, events)

for event := range events {
        // Write jsonmessage to stream.
}

if err := events.Err(); err != nil {
        // Write error to stream.
}
```

ejholmes:
This is looking good so far. Do mind splitting this into 2 PR's?

1. Adds the event stream.
2. Adds waiting for deployments to stabilize.

They're mostly orthogonal to each other and it'll make it easier to review.

deployments.go
mhahn: right, that makes sense. i was under the impression whoever was consuming from the channel should be the one to close it, but this will make things cleaner

deployments.go
mhahn: i initially had the blocking here, but was having issues with concurrent requests, so i moved it here: https://github.com/mhahn/empire/blob/c1ea348005d0c4eb832858810a87cf7926295118/server/heroku/deployments.go#L33. let me verify if that was the issue after i make some of these other changes.

scheduler/scheduler.go
mhahn: yea, i'd also like to include "levels" on the events so we have a way of publishing some of these (like the stabilization event) to the empire event stream.

i'll take a crack at this today

deployments.go
mhahn: why would the io.Writer be closed after this method returns? you're referring to this right: https://github.com/mhahn/empire/blob/c1ea348005d0c4eb832858810a87cf7926295118/server/heroku/deployments.go#L60

mhahn:
This is handled by other PRs

PR #917, opened 2016-07-08
Initial implementation of StatusStream

mhahn:
Split out setting up events from: https://github.com/remind101/empire/pull/916

Calling these "Scheduler Events" was too confusing given that we also have "Empire Events" that publish to various sources. I went with calling this a "StatusStream" that can be created and passed to an action (like Deploy) and can be published and read from throughout the lifecycle of the action.

I ended up creating a separate package `status` because otherwise we would have had to import the `scheduler` within `deployments.go` which felt off to me.


deployments.go
```diff
@@ -78,5 +80,22 @@ func (s *deployerService) Deploy(ctx context.Context, db *gorm.DB, opts DeployOp
 		return r, err
 	}
 
+	if s, ok := opts.Updates.(status.SubscribableStream); ok {
+		for update := range s.Subscribe() {
+			msg := fmt.Sprintf("Status: %s", update.String())
```
ejholmes: I think this will already get prefixed with `Status: ` when the jsonmessage is written.

ejholmes: Scratch that, maybe not.

deployments.go
```diff
@@ -78,5 +80,22 @@ func (s *deployerService) Deploy(ctx context.Context, db *gorm.DB, opts DeployOp
 		return r, err
 	}
 
+	if s, ok := opts.Updates.(status.SubscribableStream); ok {
+		for update := range s.Subscribe() {
```
ejholmes: I think, like you were saying before, this will cause concurrent deployments to lock up because we open a transaction [here](https://github.com/mhahn/empire/blob/f3afbd7e24e98c275d932581fc1c9afe31c6cb32/empire.go#L574).

In reality, we only need to make [deploymentsService.deploy](https://github.com/mhahn/empire/blob/f3afbd7e24e98c275d932581fc1c9afe31c6cb32/deployments.go#L21) automic, so we could move opening a transaction to [here](https://github.com/mhahn/empire/blob/f3afbd7e24e98c275d932581fc1c9afe31c6cb32/deployments.go#L71)

empire.go
```diff
@@ -541,6 +542,9 @@ type DeployOpts struct {
 
 	// Commit message
 	Message string
+
+	// Updates is a stream to publish status updates to during deploy
+	Updates status.StatusStream
```
ejholmes: Since we're only using this to write status updates to the io.Writer, we might as well just remove this option and move status.NewStatusStream() to [here](https://github.com/mhahn/empire/blob/f3afbd7e24e98c275d932581fc1c9afe31c6cb32/deployments.go#L71)

status/status.go
```diff
@@ -0,0 +1,91 @@
+package status
+
+import "sync"
+
+// Status represents a status update from an action within Empire.
+type Status interface {
```
ejholmes: Probably not much advantage to making this an interface, so we might as well just go with a struct like:

```go
type Status struct {
        // A friendly human readable message about the status change.
        Message string
}

// String implements the fmt.Stringer interface.
func (s *Status) String() string {
        return s.Message
}
```

It may make sense to eventually add a `Status` field to that struct, which we could use to update a field on the `Release` and show the status in `emp releases` too.

status/status.go
```diff
@@ -0,0 +1,91 @@
...
+	// Publish publishes an update to the status stream
+	Publish(Status)
+
+	// Done finalizes the status stream
+	Done(error)
+}
+
+type SubscribableStream interface {
+	Subscribe() <-chan Status
+	Error() error
```
ejholmes: Minor, but from what I've seen, the standard convention for adding a method that returns an error like this is to call it `Err()` (see [bufio#Scanner.Err](https://golang.org/pkg/bufio/#Scanner.Err), [sql#Rows.Err](https://golang.org/pkg/database/sql/#Rows.Err), etc).

ejholmes:
> I ended up creating a separate package status because otherwise we would have had to import the scheduler within deployments.go which felt off to me.

`remind101/empire` already depends on `remind101/empire/scheduler`, so it would be fine to put `StatusStream` within `remind101/empire/scheduler`. I actually think that would be better until there's a reason to extract it.

ejholmes:
This is looking great so far.

mhahn:
i'm fine moving within `remind101/empire/scheduler`. just to confirm, that is just moving `remind101/empire/status/status.go` to `remind101/empire/scheduler/status.go` or should we just include everything within `remind101/empire/scheduler/scheduler.go`?

status/status.go
mhahn: ok cool

ejholmes:
I think everything in `remind101/empire/scheduler/scheduler.go` would be best.

status/status.go
mhahn: yea, i initially had this but was looking at `events` and they implemented an interface. go makes it easy to do things before you need to :)

ejholmes: Yeah haha. If we wanted to tag on additional structured attributes, then an interface with different implementations would make sense. But for what we'll use this for right now, a struct is probably simpler.

mhahn:
ok cool

status/status.go
```diff
@@ -0,0 +1,91 @@
...
+// an Empire action.
+type StatusStream interface {
+	// Publish publishes an update to the status stream
+	Publish(Status)
+
+	// Done finalizes the status stream
+	Done(error)
+}
+
+type SubscribableStream interface {
```
ejholmes: Just food for thought. Another possibility here would be to forgo channels entirely and make the interface:

```go
// StatusStream is an interface for publishing status updates while executing
// an Empire action.
type StatusStream interface {
        // Publish publishes an update to the status stream
        Publish(Status) error

        // Done finalizes the status stream
        Done(error)

        // Wait returns a channel that that receives once Done() is called.
        // Consumers should call the Err() method to determine if an error
        // ocurred.
        Wait() <-chan struct{}

        // Returns the error from calling Done().
        Err() error
}
```

Then we just pass in an implementation that writes to the logstream:

```go
type jsonmessageStatusStream struct {
	sync.Mutex
	done chan struct{}
	err  error
	w    io.Writer
}

func (s *jsonmessageStream) Publish(message scheduler.Status) {
	select {
	case <-done:
		panic("Publish called on finalized stream")
	default:
	}

	json.NewEncoder(s.w).Encode(jsonmessage.Message{Status: status.Message})
}

func (s *jsonmessageStream) Done(err error) {
	close(s.done)
	s.err = err
}

func (s *jsonmessageStream) Err() error {
	return s.err
}

func (s *jsonmessageStream) Wait() <-chan struct{} {
	return s.done
}
```

Main advantage would be that we don't need to worry about buffering the channel, calls to `Publish` will immediately write to the stream and it's easy to wrap with middleware (e.g. writing to the apps kinesis stream, etc).

mhahn: i like that.

another thing i'm coming across is having to do: https://github.com/remind101/empire/pull/917/files#diff-48d76aef8c283046a79db32824933290R523.

not the end of the world, but definitely annoying.

i was thinking of adding an `active` flag within the stream, if it never gets activated, then the subscriber never blocks. there could either be a specific `activate` method or as soon as you `publish` the stream would get activated. but that defeats the purpose of being able to write to the stream at any point within a go routine and having the client subscribe.

also, i'm not a fan of throwing: `panic("Publish called on finalized stream")`. given these are just status updates, it feels aggressive to throw a panic. what would be the proper way to warn instead? i couldn't find any examples in the code.

ejholmes: > also, i'm not a fan of throwing: panic("Publish called on finalized stream"). given these are just status updates, it feels aggressive to throw a panic. what would be the proper way to warn instead? i couldn't find any examples in the code.

Agreed. If we change the interface to `Publish(ctx, Status)`, then you can do a `logger.Warn(ctx, "message")`.

mhahn: ah ok cool

mhahn:
ok, i implemented the `jsonmessageStatusStream` as well as the other pieces of feedback.

the only part i'm not happy with is having to do: https://github.com/remind101/empire/pull/917/files#diff-48d76aef8c283046a79db32824933290R523

deployments.go
```diff
@@ -57,18 +59,21 @@ func (s *deployerService) deploy(ctx context.Context, db *gorm.DB, opts DeployOp
...
 }
 
 // Deploy is a thin wrapper around deploy to that adds the error to the
 // jsonmessage stream.
-func (s *deployerService) Deploy(ctx context.Context, db *gorm.DB, opts DeployOpts) (*Release, error) {
+func (s *deployerService) Deploy(ctx context.Context, opts DeployOpts) (*Release, error) {
 	var msg jsonmessage.JSONMessage
 
-	r, err := s.deploy(ctx, db, opts)
+	tx := s.db.Begin()
```
ejholmes: I think we'd also wanna rollback here. Might be easier to wrap the transaction logic in a separate method:

```go
func (s *deployerService) deployInTransaction(ctx context.Context, stream scheduler.StatusStream, opts De
        tx := s.db.Begin()
        r, err := s.deploy(ctx, tx, stream, opts)
        if err != nil {
                tx.Rollback()
                return r, err
        }

        if err := tx.Commit().Error; err != nil {
                return r, err
        }

        return r, nil
}
```

scheduler/scheduler.go
```diff
@@ -152,3 +157,69 @@ func merge(envs ...map[string]string) map[string]string {
...
+type jsonmessageStatusStream struct {
+	sync.Mutex
+	done chan struct{}
+	err  error
+	w    io.Writer
+}
+
+// NewJSONMessageStream returns a new instance of the default status stream.
+func NewJSONMessageStream(w io.Writer) StatusStream {
+	return &jsonmessageStatusStream{w: w, done: make(chan struct{}, 1)}
```
ejholmes: No need to buffer this channel, since we never actually send on it.

releases.go
```diff
@@ -146,13 +146,13 @@ func (s *releasesService) Rollback(ctx context.Context, db *gorm.DB, opts Rollba
...
-	})
+	}, nil)
 }
 
 // Release submits a release to the scheduler.
-func (s *releasesService) Release(ctx context.Context, release *Release) error {
+func (s *releasesService) Release(ctx context.Context, release *Release, ss scheduler.StatusStream) error {
 	a := newSchedulerApp(release)
-	return s.Scheduler.Submit(ctx, a)
+	return s.Scheduler.Submit(ctx, a, ss)
```
ejholmes: If `ss` is nil, maybe want to create a null implementation and pass that through, so implementations of scheduler.Scheduler don't need to check for nil.

ejholmes:
I just [hacked together](https://github.com/remind101/empire/commit/a34b0e60977839d50f539549ad11e65c84c48e79) a couple changes in the CloudFormation backend to use this. So awesome seeing more context in the deployment stream:

https://asciinema.org/a/b6vnccpg2u2uykf9wyyb1ijej

This looks gtm 👍. We can make changes to `Scheduler` implementations to use this in a separate PR (or if you wanna just merge those changes into this branch, that's cool too).

phobologic:
raw

scheduler/scheduler.go
```diff
@@ -152,3 +157,69 @@ func merge(envs ...map[string]string) map[string]string {
...
+}
+
+// String implements the fmt.Stringer interface.
+func (s *Status) String() string {
+	return s.Message
+}
+
+// StatusStream is an interface for publishing status updates while a scheduler
+// is executing.
+type StatusStream interface {
```
ejholmes: I was thinking some more about this last night and I think there's one more small change we might wanna do that will make this much simpler. We can give semantic meaning to when the StatusStream is nil, and change the signature of `scheduler.Submit` to:

```go
// When StatusStream is nil, Submit should return as quickly as possible, usually when the new version has been received, and validated. If StatusStream is not nil, it's recommended that the method not return until the deployment has fully completed.
scheduler.Submit(context.Context, *App, StatusStream) error
```

This would mean, that we can remove the `Done`, `Wait`, and `Err` methods on the `StatusStream` interface so it's just `Publish`. For the CloudFormation backend, we'd just change [this line](https://github.com/remind101/empire/blob/bd469460a1fc75b387bc49cdc686e05e896bb9d4/scheduler/cloudformation/cloudformation.go#L271)  to be `if ss != nil`.

The primary advantage will be in cases like the [migration scheduler](https://github.com/remind101/empire/blob/bd469460a1fc75b387bc49cdc686e05e896bb9d4/scheduler/cloudformation/migration.go#L152-L163) where we'd call Submit on two different schedulers, that might both call `Done`.

deployments.go
mhahn: i was rolling back here: https://github.com/remind101/empire/pull/917/files#diff-ff47559c0383dd1959cdbe50cd2c036dR76 which is roughly equivalent, but +1 on the separation, i was thinking the same

scheduler/scheduler.go
mhahn: i like this. i think it helps differentiate `EventStream` as async empire notifications and `StatusStream` as synchronous notifications during the execution of an action.

mhahn:
@ejholmes i made one more change this weekend: https://github.com/remind101/empire/pull/917/commits/9843828726f53829bc56e2be0c9225aba0ddeab1

this fixes concurrent requests

ejholmes:
Awesome 👍 (sorry for the churn :)).

mhahn:
@ejholmes https://github.com/remind101/empire/pull/917/commits/c6441b7c6d0cac2d1e525e9e09e71d43804068e7 is actually the commit I was talking about, I forgot I had done it in a different branch.

Without this the db transaction surrounds both the creation and the submission of the release, which breaks concurrency.

mhahn:
@ejholmes ok i think this is good to go. i added an `-s` flag so we'll only setup the status stream if someone enables it. This means people can opt in to the new UX.

mhahn:
Deploy now looks like:

```
$ ./build/emp deploy remind101/acme-inc:master -s
master: Pulling from remind101/acme-inc

420890c9e918: Already exists
a3ed95caeb02: Already exists
1e14aab0083d: Already exists
4393485f2bb3: Already exists
a2a051cf14f7: Already exists
1f9177248efa: Already exists
Digest: sha256:d05682347b56191a950f55ab06f11a6c6e34041921af32281b01f4a6a9a0f75f
Status: Created cloudformation template: https://empire-0b96756d-templatebucket-fzr5jqii49z4.s3.amazonaws.com/acme-inc/e904d241-a249-4b13-aa77-9dc0a3639617/57c5f02918c0fd1714d4689075924b42523c4fb6
Status: Stack update submitted
Status: Waiting for stack update to complete
Status: Stack update complete
Status: Finished processing events for release v9 of acme-inc
```

When a deploy has been superseded:

```
$ ./build/emp deploy remind101/acme-inc:master -s
master: Pulling from remind101/acme-inc

420890c9e918: Already exists
a3ed95caeb02: Already exists
1e14aab0083d: Already exists
4393485f2bb3: Already exists
a2a051cf14f7: Already exists
1f9177248efa: Already exists
Digest: sha256:d05682347b56191a950f55ab06f11a6c6e34041921af32281b01f4a6a9a0f75f
Status: Created cloudformation template: https://empire-0b96756d-templatebucket-fzr5jqii49z4.s3.amazonaws.com/acme-inc/86874600-74f0-4ee4-9517-51c13a3e7240/3c156706ec606aa39375c6dfd7182f2f8a1d0ee4
Status: Waiting for existing stack operation to complete
Status: Operation superseded by newer release
Status: Finished processing events for release v4 of acme-inc
```

While waiting for another deploy to finish:

```
$ ./build/emp deploy remind101/acme-inc:master -s
master: Pulling from remind101/acme-inc

420890c9e918: Already exists
a3ed95caeb02: Already exists
1e14aab0083d: Already exists
4393485f2bb3: Already exists
a2a051cf14f7: Already exists
1f9177248efa: Already exists
Digest: sha256:d05682347b56191a950f55ab06f11a6c6e34041921af32281b01f4a6a9a0f75f
Status: Created cloudformation template: https://empire-0b96756d-templatebucket-fzr5jqii49z4.s3.amazonaws.com/acme-inc/86874600-74f0-4ee4-9517-51c13a3e7240/d5e7525e4fb98ee49bd3aceceed7e726a9029e64
Status: Waiting for existing stack operation to complete
Status: Stack update submitted
Status: Waiting for stack update to complete
Status: Stack update complete
Status: Finished processing events for release v5 of acme-inc
```

Create looks like:

```
$ ./build/emp deploy remind101/acme-inc:master -s
master: Pulling from remind101/acme-inc

420890c9e918: Already exists
a3ed95caeb02: Already exists
1e14aab0083d: Already exists
4393485f2bb3: Already exists
a2a051cf14f7: Already exists
1f9177248efa: Already exists
Digest: sha256:d05682347b56191a950f55ab06f11a6c6e34041921af32281b01f4a6a9a0f75f
Status: Created cloudformation template: https://empire-0b96756d-templatebucket-fzr5jqii49z4.s3.amazonaws.com/acme-inc/04a75f65-c7c4-4e36-943b-0a09614db1b4/67f96646de01688a45cc9d8a5f25ff9cb1ce479c
Status: Creating stack
Status: Stack created
Status: Finished processing events for release v2 of acme-inc
```

I wanted to get the actual cloudformation url, but i didn't see that we had easy access to the stack arn.

scheduler/cloudformation/cloudformation.go
```diff
@@ -208,6 +208,10 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
 		return err
 	}
 
+	if err := scheduler.Publish(ss, fmt.Sprintf("Created cloudformation template: %v", *t.URL)); err != nil {
```
ejholmes: Part of me feels like we should just ignore errors from `Publish`. If for example, the `emp client` disconnected from the network, this would result in `err == io.EOF` and would cause the deployment to fail.

tests/empire/empire_test.go
```diff
@@ -555,7 +555,18 @@ type mockScheduler struct {
 	mock.Mock
 }
 
+type processesByType []*scheduler.Process
```
ejholmes: If you rebase this, this change should already be in master.

scheduler/cloudformation/cloudformation.go
mhahn: yea i was thinking the same thing initially but also didn't want to just swallow errors. maybe we just log these?

ejholmes:
We should also add `Stream: true` to https://github.com/remind101/empire/blob/0f2cc1aa401cd98e0b5f202a9c6bedb965713a84/server/github/deployer.go#L59-L63.

releases.go
```diff
@@ -122,12 +132,7 @@ func (s *releasesService) Create(ctx context.Context, db *gorm.DB, r *Release) (
 	}
 
 	r, err := releasesCreate(db, r)
-	if err != nil {
-		return r, err
-	}
-
-	// Schedule the new release onto the cluster.
-	return r, s.Release(ctx, r)
+	return r, err
```
ejholmes: Minor, you can just `return releasesCreate(db, r)`.

scheduler/cloudformation/cloudformation.go
ejholmes: sounds good

scheduler/cloudformation/cloudformation.go
```diff
@@ -327,8 +326,17 @@ type createStackInput struct {
 // createStack creates a new CloudFormation stack with the given input. This
 // function returns as soon as the stack creation has been submitted. It does
 // not wait for the stack creation to complete.
-func (s *Scheduler) createStack(input *createStackInput, done chan error) error {
-	waiter := s.cloudformation.WaitUntilStackCreateComplete
+func (s *Scheduler) createStack(input *createStackInput, done chan error, ss scheduler.StatusStream) error {
+	waiter := func(input *cloudformation.DescribeStacksInput) error {
+		if err := scheduler.Publish(ss, "Creating stack"); err != nil {
```
ejholmes: Just a thought, it would be really cool to stream the stack events back. Probably not in this PR though :).

deployments.go
```diff
@@ -58,25 +59,43 @@ func (s *deployerService) deploy(ctx context.Context, db *gorm.DB, opts DeployOp
...
+	if opts.Stream {
+		stream = scheduler.NewJSONMessageStream(opts.Output)
+	}
+
+	r, err := s.deployInTransaction(ctx, stream, opts)
 	if err != nil {
 		msg = newJSONMessageError(err)
-	} else {
+	} else if stream == nil {
 		msg = jsonmessage.JSONMessage{Status: fmt.Sprintf("Status: Created new release v%d for %s", r.Version, r.App.Name)}
```
ejholmes: I feel like this message should always be present, regardless of whether streaming is enabled or not.

deployments.go
```diff
@@ -58,25 +59,43 @@ func (s *deployerService) deploy(ctx context.Context, db *gorm.DB, opts DeployOp
...
+	} else {
+		scheduler.Publish(stream, fmt.Sprintf("Finished processing events for release v%d of %s", r.Version, r.App.Name))
```
ejholmes: I think we should just set `msg =` like we do above. Right now, `msg` is nil, which results in an empty line getting added to the stream after this call:

```console
$ ./build/emp deploy remind101/acme-inc:master -s
master: Pulling from remind101/acme-inc
420890c9e918: Already exists
a3ed95caeb02: Already exists
1e14aab0083d: Already exists
4393485f2bb3: Already exists
a2a051cf14f7: Already exists
1f9177248efa: Already exists
Digest: sha256:d05682347b56191a950f55ab06f11a6c6e34041921af32281b01f4a6a9a0f75f
Status: Created cloudformation template: https://empire-77792028-templatebucket-195oucd149ybu.s3.amazonaws.com/acme-inc/72c3d272-2be3-49fa-a175-32daf22e6d57/9ef308a9614931cac961791097dbd97d5438280e
Status: Stack update submitted
Status: Waiting for stack update to complete
Status: Stack update complete
Status: Finished processing events for release v3 of acme-inc

$
```

ejholmes:
Couple of minor things, which we can follow up with later. This is awesome 👍.

deployments.go
mhahn: it looked weird when the stack had been superseded. 

You'd see

```
Status: Stack operation superseded by newer release
Status: Created new release v4 for acme-inc
```

which seemed off, but i don't have a strong preference.

ejholmes: > Or we can put the "created new release" at a different point.

yeah, I should have been more specific. I think we should still log that line, but after the release has been committed and before we submit it to the scheduler.

ejholmes:
Oh, can you update the changelog too?

scheduler/cloudformation/cloudformation.go
mhahn: yea that would be really cool, i set that up on stacker and it was helpful. created an issue to track: https://github.com/remind101/empire/issues/922

deployments.go
mhahn: ok, new messaging looks like:

no stream:


```
$ ./build/emp deploy remind101/acme-inc:master
master: Pulling from remind101/acme-inc

420890c9e918: Already exists
a3ed95caeb02: Already exists
1e14aab0083d: Already exists
4393485f2bb3: Already exists
a2a051cf14f7: Already exists
1f9177248efa: Already exists
Digest: sha256:d05682347b56191a950f55ab06f11a6c6e34041921af32281b01f4a6a9a0f75f
Status: Created new release v10 for acme-inc
Status: Finished processing events for release v10 of acme-inc
```

streaming:

```
$ ./build/emp deploy remind101/acme-inc:master -s
master: Pulling from remind101/acme-inc

420890c9e918: Already exists
a3ed95caeb02: Already exists
1e14aab0083d: Already exists
4393485f2bb3: Already exists
a2a051cf14f7: Already exists
1f9177248efa: Already exists
Digest: sha256:d05682347b56191a950f55ab06f11a6c6e34041921af32281b01f4a6a9a0f75f
Status: Created new release v11 for acme-inc
Status: Created cloudformation template: https://empire-0b96756d-templatebucket-fzr5jqii49z4.s3.amazonaws.com/acme-inc/04a75f65-c7c4-4e36-943b-0a09614db1b4/5faf9e742e482f0e66c768d1f134986a93795615
Status: Waiting for existing stack operation to complete
Status: Stack update submitted
Status: Waiting for stack update to complete
Status: Stack update complete
Status: Finished processing events for release v11 of acme-inc
```

I added the "Finished..." message to each since after we create the release, we submit to the scheduler, so without it, it might feel like there is an unknown lag.

ejholmes: Awesome

PR #921, opened 2016-07-12
Fixup docs on scheduled tasks

ejholmes:
Changes the example cron expression to one that actually works, and points to the docs on cron expressions in CloudWatch events.


phobologic:
👍

mhahn:
:+1:

ejholmes:
Added a little more to the docs, with information about why you might want to use the built-in support instead of in an process cron.

PR #923, opened 2016-07-13
Deploy stabilization

mhahn:
When a stream is provided to the CF scheduler, we'll wait for the deploy to stabilize, streaming status updates as the deploy completes.


server/cloudformation/ecs.go
```diff
@@ -196,3 +212,16 @@ func requiresReplacement(new, old *ECSServiceProperties) bool {
 
 	return false
 }
+
+func getPrimaryDeployment(service *ecs.Service) (deployment *ecs.Deployment, err error) {
```
ejholmes: You can just change this function to:

```go
func primaryDeployment(service *ecs.Service) (*ecs.Deployment) {
        for _, d := range service.Deployments {
                if d.Status != nil && *d.Status == "PRIMARY" {
                        return d
                }
        }
        return nil
}
```

And use the nil value to indicate that there's no primary deployment, then return the error elsewhere.

scheduler/cloudformation/template.go
```diff
@@ -167,6 +168,7 @@ func (t *EmpireTemplate) Build(app *scheduler.App) (*troposphere.Template, error
 		default:
 			service := t.addService(tmpl, app, p)
 			serviceMappings = append(serviceMappings, Join("=", p.Type, Ref(service)))
+			deploymentMappings = append(deploymentMappings, Join("=", p.Type, GetAtt(service, "DeploymentId")))
```
ejholmes: We currently have support for using the standard `AWS::ECS::Service` resource, which wouldn't have the `DeploymentId` attribute. We should either make this conditional, or remove support for the standard ecs service, which I think would be fine.

scheduler/cloudformation/cloudformation.go
```diff
@@ -268,11 +271,129 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
 		if err := <-done; err != nil {
 			return err
 		}
+		done = make(chan error, 1)
+		s.waitForDeploymentToStabilize(ctx, stackName, ss, done)
```
ejholmes: This has the potential to produce incorrect results, because DescribeStack is called while this deployment isn't holding the stack operation lock. Maybe we should make `performStackOperation` return a `DescribeStackOutput`.

scheduler/cloudformation/cloudformation.go
```diff
@@ -268,11 +271,129 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
...
+		if err = <-done; err != nil {
+			return err
+		}
 	}
 
 	return nil
 }
 
+func (s *Scheduler) waitForDeploymentToStabilize(ctx context.Context, name string, ss scheduler.StatusStream, done chan error) {
```
ejholmes: I think we can just remove the `done chan error` parameter here, and return an error.

ejholmes: Also, minor, maybe `waitForDeploymentsToStabilize` (plural `Deployments`) since we're waiting on multiple deployments.

ejholmes: We might be able to simplify this method by returning a channel where we send stabilization events on. Something like this:

```go
// waitForDeploymentsToStabilize returns a channel that you can range over to
// wait for multiple ECS deployments to stabilize. The channel is closed
// when one of the following occurs:
//
//	1. All deployments enter a "stable" state.
//	2. Any deployment enters an inactive state.
func (s *Scheduler) waitForDeploymentsToStabilize(deployments []ecsDeployment) <-chan deploymentStatus {
	ch := make(chan deploymentStatus)

	go func(deployments []ecsDeployment) {
		for len(deployments) > 0 {
			// Describe the services

			for i, d := range deployments {
				if stable {
					ch <- deploymentStatus{d, "stable"}
					// Remove the deployment, since it's
					// stable. No need to keep checking it.
					deployments = append(deployments[:i], deployments[i+1]...)
				} else if primary {
					// do nothing
				} else {
					ch <- deploymentStatus{d, "inactive"}
					close(ch)
					return
				}
			}
		}

		// All done
		close(ch)
	}(deployments[:])

	return ch
}
```

Then we can generate the list of deployments, and publish to the status stream as deployments change state.

```go
deploymentStatuses := s.waitForDeploymentsToStabilize(deployments)
for status := range deploymentStatuses {
	scheduler.Publish("Service %s became %s", status.service, status)
}

// Done
```

scheduler/cloudformation/cloudformation.go
```diff
@@ -512,6 +633,16 @@ func (s *Scheduler) stack(stackName *string) (*cloudformation.Stack, error) {
 	return resp.Stacks[0], nil
 }
 
+// output returns the cloudformation.Output that matches the given key.
+func (s *Scheduler) output(stack *cloudformation.Stack, key string) (output *cloudformation.Output) {
```
ejholmes: This doesn't depend on `Scheduler`, so we can just move it into a `func output(...)`

ejholmes:
Figure you probably weren't ready for this to get reviewed, since tests were failing, but added some comments anyway. This is gonna be great.

mhahn:
Thanks. Yea I need to setup mocks for describing the outputs in the tests.

scheduler/cloudformation/template.go
mhahn: yea i would vote for simplifying the footprint of what we support

scheduler/cloudformation/cloudformation.go
mhahn: yea, i wish waiter returned the result too so we didn't have to make a duplicate request

scheduler/cloudformation/cloudformation.go
```diff
@@ -414,16 +510,19 @@ func (s *Scheduler) updateStack(ctx context.Context, input *updateStackInput, do
 //   until the other stack operation has completed.
 // * If there is another pending stack operation, it will be replaced by the new
 //   update.
-func (s *Scheduler) performStackOperation(ctx context.Context, stackName string, fn func() error, waiter func(*cloudformation.DescribeStacksInput) error, ss scheduler.StatusStream) error {
+func (s *Scheduler) performStackOperation(ctx context.Context, stackName string, fn func() error, waiter func(*cloudformation.DescribeStacksInput) error, ss scheduler.StatusStream) stackOperationOutput {
```
ejholmes: I think we should probably still make the return value of this `(*cloudformation.Stack, error)`, then wrap it in a `stackOperationOutput` when we send it on the channel.

scheduler/cloudformation/cloudformation.go
```diff
@@ -257,22 +261,109 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
...
+			var serviceIDs []*string
+			arnToDeployment := make(map[string]*deploymentIndex)
+			for i, d := range deployments {
+				serviceIDs = append(serviceIDs, aws.String(d.serviceID))
+				arnToDeployment[d.serviceARN] = &deploymentIndex{i, d}
+			}
+
+			status, err := s.ecs.DescribeServices(&ecs.DescribeServicesInput{
+				Cluster:  aws.String(s.Cluster),
+				Services: serviceIDs,
```
ejholmes: I think you can just use the `serviceARN` here and remove the `serviceID` field.

scheduler/cloudformation/cloudformation.go
```diff
@@ -257,22 +261,109 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
...
-			return err
+		if err := s.waitUntilStable(ctx, output, ss); err != nil {
+			logger.Warn(ctx, fmt.Sprintf("error waiting for submit to stabilize: %v", err))
 		}
 	}
+	return nil
+}
 
+func (s *Scheduler) waitUntilStable(ctx context.Context, o chan stackOperationOutput, ss scheduler.StatusStream) error {
+	output := <-o
```
ejholmes: Feels kinda strange to do this here. Should probably receive it on line 272, check the error there, then pass the `*cloudformation.Stack` to `waitUntilStable`.

ejholmes: So, right now, the error from performStackOperation would only be warned. We want that one to bubble up.

scheduler/cloudformation/cloudformation.go
```diff
@@ -257,22 +261,109 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
...
+func (d *deploymentStatus) String() string {
+	return d.status
+}
+
+type deploymentIndex struct {
+	index      int
+	deployment *ecsDeployment
+}
+
+func (s *Scheduler) waitForDeploymentsToStabilize(ctx context.Context, deployments []*ecsDeployment) <-chan deploymentStatus {
```
ejholmes: Should we make `deployments` a `map[string]*ecsDeployment` (where string would be the service ARN)? I think that might make some of this simpler, since you wouldn't need to build the `arnToDeployment` map, and removing a stable service is just:

```go
delete(deployments, serviceARN)
```

scheduler/cloudformation/cloudformation.go
```diff
@@ -257,22 +261,109 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
...
+				arnToDeployment[d.serviceARN] = &deploymentIndex{i, d}
+			}
+
+			status, err := s.ecs.DescribeServices(&ecs.DescribeServicesInput{
+				Cluster:  aws.String(s.Cluster),
+				Services: serviceIDs,
+			})
+			if err != nil {
+				logger.Warn(ctx, fmt.Sprintf("error describing services: %v", err))
+				close(ch)
```
ejholmes: We may wanna encapsulate channel closing and logging of errors, by making this function a closure that returns an error, then running it inside the goroutine. e.g.:

```go
ch := make(chan deploymentStatus)

wait := func(deployments ...) error {
}

go func(deployments ...) {
        if err := wait(deployments); err != nil {
                logger.Warn()
        }
        close(ch)
}(deployments[:])
return ch
```

scheduler/cloudformation/cloudformation.go
ejholmes: And you wouldn't need the `deploymentIndex` type I think.

scheduler/cloudformation/cloudformation.go
```diff
@@ -941,3 +1039,61 @@ func (e *templateValidationError) Error() string {
...
+	for _, o := range stack.Outputs {
+		if *o.OutputKey == key {
+			output = o
+		}
+	}
+	return
+}
+
+type ecsDeployment struct {
+	service    string
```
ejholmes: I would change this to `process` or `processName`, so `service` doesn't get confused with the ECS service.

scheduler/cloudformation/cloudformation.go
mhahn: yea my logic with that was you wouldn't get those if you hadn't specified a statusStream (since we don't wait) so they should just be warnings. but i agree it should be an error

scheduler/cloudformation/cloudformation.go
mhahn: good call, i can't remember why i thought i needed serviceID for that call

mhahn:
@ejholmes thanks for all the feedback on this. should be good to go.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * The Custom::ECSService custom resource now waits for newly created ECS services to stabilize [#878](https://github.com/remind101/empire/pull/878)
 * The CloudFormation backend now uses the Custom::ECSService resource instead of AWS::ECS::Service, by default [#877](https://github.com/remind101/empire/pull/877)
 * The database schema version is now checked at boot, as well as in the http health checks. [#893](https://github.com/remind101/empire/pull/893)
+* The log level within empire can now be configured when starting the service. [#929]((https://github.com/remind101/empire/issues/929)
```
ejholmes: Do you wanna merge https://github.com/remind101/empire/pull/930 first then rebase this?

mhahn: yea

scheduler/cloudformation/cloudformation.go
```diff
@@ -258,22 +262,103 @@ func (s *Scheduler) submit(ctx context.Context, tx *sql.Tx, app *scheduler.App,
...
+}
 
+func (s *Scheduler) waitUntilStable(ctx context.Context, stack *cloudformation.Stack, ss scheduler.StatusStream) error {
+	deployments, err := deploymentsToWatch(stack)
+	if err != nil {
+		return err
+	}
+	deploymentStatuses := s.waitForDeploymentsToStabilize(ctx, deployments)
+	for status := range deploymentStatuses {
+		scheduler.Publish(ctx, ss, fmt.Sprintf("Service %s became %s", status.deployment.process, status))
```
ejholmes: Can we change this to `Process %s became %s`?

ejholmes:
Do we have any tests for the different stabilization scenarios? Otherwise, this looks good to me once it's rebased 👍

mhahn:
@ejholmes good call on the tests. caught a loop if the status becomes inactive due to another deploy. I adjusted the `wait` func to return a boolean for whether or not we should keep waiting.

ejholmes:
:+1:

PR #924, opened 2016-07-13
Adds a DeploymentStream type

ejholmes:
The fact that `DeployOpts.Output` was an io.Writer was kind of misleading, because everything that gets written to it needs to be in jsonmessage format.

This changes the type to a `DeploymentStream`, which adds some convenience on top of an io.Writer to encode jsonmessages, and give us static typing to give us a little more assurance that what is going into the stream is in jsonmessage format. This also gives us a good place to implement the `scheduler.StatusStream` interface, so we can just re-use the DeploymentStream and pass it to the scheduler, when streaming is enabled.

### DIFF
--- a/deployments.go
+++ b/deployments.go
@@ -3,6 +3,7 @@ package empire
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 
 	"github.com/docker/docker/pkg/jsonmessage"
 	"github.com/jinzhu/gorm"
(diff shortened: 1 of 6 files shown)


deployments.go
```diff
@@ -75,37 +76,82 @@ func (s *deployerService) createInTransaction(ctx context.Context, stream schedu
...
 	}
 
-	if err := write(opts, msg); err != nil {
-		return r, err
+	return r, w.Status(fmt.Sprintf("Finished processing events for release v%d of %s", r.Version, r.App.Name))
+}
+
+// DeploymentStream provides a wrapper around an io.Writer for writing
+// jsonmessage statuses, and implements the scheduler.StatusStream interface.
+// events, and writing
```
mhahn: leftover?

mhahn:
👍 sweet!

deployments.go
ejholmes: wth. I must have hit `p` in vim and didn't notice haha. good catch.

PR #925, opened 2016-07-13
Expand list of valid log drivers

brianz:
Currently the list of valid log drivers is limited to only `json-file` and `syslog`:

https://github.com/remind101/empire/blob/5facae771b03e976e6d68437f86bdb6f6ef705c1/pkg/ecsutil/ecs.go#L243-L246

This should be at the very least expanded to include all of the valid log drivers in Docker (which I know varies by version).  At the most, this check could be taken out completely and the user can pass anything they want in...I'd expect ECS to blow up if the user passed in something invalid.

I see an argument for removing the check since it'd do away with the need to keep up with the list of valid log drivers in Docker.

Thoughts on expanding the list vs removing it @ejholmes ?

### DIFF
--- a/pkg/ecsutil/ecs.go
+++ b/pkg/ecsutil/ecs.go
@@ -240,11 +240,6 @@ func (c *limitedClient) DescribeTasks(ctx context.Context, input *ecs.DescribeTa
 	}, nil
 }
 
-var validLogDrivers = map[string]bool{
-	"json-file": true,
-	"syslog":    true,
-}
-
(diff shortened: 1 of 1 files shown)


ejholmes:
> I see an argument for removing the check since it'd do away with the need to keep up with the list of valid log drivers in Docker.

Yeah. I think unless we map the list of valid versions, to the Docker version, then there's not much point in keeping the check. I'd be in favor of just making it freeform.

brianz:
Updated PR to simply remove the check entirely @ejholmes

ejholmes:
Awesome, thanks @brianz!

PR #926, opened 2016-07-14
Update aws-sdk-go to v1.2.5

ejholmes:
This allows Empire itself to obtain IAM credentials from the ecs agent. CloudFormation doesn't yet support providing an IAM role when creating a task definition, so the `EmpirePolicy` is still assigned to the instance role in the example cloudformation stack.

Note that this PR doesn't add support for specifying an IAM role for **apps** to use.


mhahn:
👍

PR #928, opened 2016-07-14
Include template size in status message

ejholmes:
This will include the template size in the message, so there's more transparency when an app may be getting close to the CloudFormation template size limit.


phobologic:
👍

mhahn:
👍

Issue #929, opened 2016-07-14
Support providing a log level to empire


mhahn:
This would let us add more debug logging via `logger.Debug(ctx, ...)` and only enable these during dev or debugging.

ejholmes:
Someone related, I'd love to do a revamp of logging. There's some places where we use the go logger (e.g. `cmd/empire`) and some places where we use `pkg/logger`. It'd also probably be good to remove the dependency on `pkg/logger` and bring it in as `remind101/empire/logger`.

PR #930, opened 2016-07-14
Support log level

mhahn:
Fixes: https://github.com/remind101/empire/issues/929


mhahn:
It would be cool to have a standard for doing these debug messages. 

ie. i like in node that you can use: https://github.com/visionmedia/debug where each file inits a debugger with some common prefix like "http"

then all the messages are prefixed:

`http: message`

which makes it easier to see where something is coming from (ie. i usually make the prefix the filename)

ejholmes:
> It would be cool to have a standard for doing these debug messages.

I think we should be able to do that pretty easily with [CallerFileHandler](https://github.com/inconshreveable/log15/blob/03f7e3c942fe248f8dd5f5ad5e714b57ef7e419a/doc.go#L100) middleware in log15, or something similar that uses the [stack from the Record](https://github.com/inconshreveable/log15/blob/03f7e3c942fe248f8dd5f5ad5e714b57ef7e419a/doc.go#L100).

👍

mhahn:
Oh nice, I say we start with `CallerFileHandler` and see if we need the stack details

mhahn:
Should we enable this for debug only?

ejholmes:
Yeah, probably only for debug 👍

PR #933, opened 2016-07-14
Remove support for using the standard AWS::ECS::Service resource type

ejholmes:
See https://github.com/remind101/empire/pull/923#discussion_r70767122

The `Custom::ECSService` has been stable and is a better fit for Empire's usage.


mhahn:
👍 thanks!

PR #935, opened 2016-07-15
Custom::ECSTaskDefinition resource

ejholmes:
Re-opening https://github.com/remind101/empire/pull/859

Now that we have more than 1 app that's running close to the CloudFormation template size limit, probably a good time to tackle this again.

This is enabled only when setting the `ECS_TASK_DEFINITION=custom` environment variable, the default will remain `AWS::ECS::TaskDefinition`.

**TODO**

* [x] More tests
* [ ] Handle `ClientException: TaskDefinition is inactive`
* [ ] Add BoolValue type.

### DIFF
--- a/scheduler/cloudformation/template.go
+++ b/scheduler/cloudformation/template.go
@@ -42,6 +42,10 @@ const (
 	defaultCNAMETTL                        = 60
 
 	runTaskFunction = "RunTaskFunction"
+
+	appEnvironment = "AppEnvironment"
+
+	restartLabel = "cloudformation.restart-key"
 )
 
 // This implements the Template interface to create a suitable CloudFormation
@@ -158,7 +162,21 @@ func (t *EmpireTemplate) Build(app *scheduler.App) (*troposphere.Template, error
 	serviceMappings := []interface{}{}
 	scheduledProcesses := map[string]string{}
 
+	if taskDefinitionResourceType(app) == "Custom::ECSTaskDefinition" {
+		tmpl.Resources[appEnvironment] = troposphere.Resource{
+			Type: "Custom::ECSEnvironment",
+			Properties: map[string]interface{}{
+				"ServiceToken": t.CustomResourcesTopic,
+				"Environment":  sortedEnvironment(app.Env),
+			},
+		}
+	}
+
 	for _, p := range app.Processes {
+		if p.Env == nil {
+			p.Env = make(map[string]string)
+		}
+
 		tmpl.Parameters[scaleParameter(p.Type)] = troposphere.Parameter{
 			Type: "String",
 		}
@@ -168,7 +186,8 @@ func (t *EmpireTemplate) Build(app *scheduler.App) (*troposphere.Template, error
 			// To save space in the template, avoid adding the
 			// resources if the process is scaled down.
 			if p.Instances > 0 {
-				scheduledProcesses[p.Type] = t.addScheduledTask(tmpl, app, p)
+				taskDefinition := t.addScheduledTask(tmpl, app, p)
+				scheduledProcesses[p.Type] = taskDefinition.Name
 			}
 		default:
 			service := t.addService(tmpl, app, p)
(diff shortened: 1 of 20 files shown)


server/cloudformation/cloudformation.go
```diff
@@ -339,18 +348,11 @@ func (c *CustomResourceProvisioner) Handle(ctx context.Context, message *sqs.Mes
 	switch err {
 	case nil:
 		resp.Status = StatusSuccess
-		logger.Info(ctx, "cloudformation.provision",
-			"request", req,
-			"response", resp,
-		)
+		logger.Info(ctx, "cloudformation.provision")
```
ejholmes: I removed logging the full request and response, since there might be plain text secrets in it. But we should at least log some information about the request/response.

server/cloudformation/ecs.go
```diff
@@ -173,6 +179,235 @@ func (p *ECSServiceResource) delete(ctx context.Context, service, cluster *strin
...
+
+		for _, m := range c.PortMappings {
+			portMappings = append(portMappings, &ecs.PortMapping{
+				ContainerPort: m.ContainerPort.Value(),
+				HostPort:      m.HostPort.Value(),
+			})
+		}
+
+		if c.Essential != nil {
+			essential = aws.Bool(*c.Essential == "true")
```
ejholmes: I think this could also be a real bool. Need a `BoolValue` type.

server/cloudformation/ecs.go
```diff
@@ -173,6 +179,235 @@ func (p *ECSServiceResource) delete(ctx context.Context, service, cluster *strin
...
+	if err != nil {
+		return "", fmt.Errorf("error creating task definition: %v", err)
+	}
+	return *resp.TaskDefinition.TaskDefinitionArn, nil
+}
+
+func (p *ECSTaskDefinitionResource) delete(arn string) error {
+	// We're ignoring errors here because we really don't care if this
+	// fails.
+	p.ecs.DeregisterTaskDefinition(&ecs.DeregisterTaskDefinitionInput{
```
ejholmes: Should `reporter.Report` the error here.

scheduler/cloudformation/template.go
```diff
@@ -248,6 +250,28 @@ func (t *EmpireTemplate) addService(tmpl *troposphere.Template, app *scheduler.A
...
 	// after updating.
 	ecsServiceType := "Custom::ECSService"
+	taskDefinitionType := "AWS::ECS::TaskDefinition"
+
+	if app.Env["ECS_TASK_DEFINITION"] == "custom" {
+		taskDefinitionType = "Custom::ECSTaskDefinition"
+	}
+
+	if taskDefinitionType == "Custom::ECSTaskDefinition" {
+		var env []interface{}
```
ejholmes: These should get sorted in alphabetical order so order doesn't cause a change.

ejholmes:
For an app that used to generate a ~420kb template, this drops the template size down to ~80k. It does make stack updates slower, but I think that could be improved by removing the process specific `Custom::ECSEnvironment`'s and merging them into the task definition.

ejholmes:
To improve performance, we should be able to do the following:

1. Start up multiple workers calling `ReceiveMessage`.
2. Remove process specific `Custom::ECSEnvironment` and just add those environment variables to the task definition directly.
3. Handle updates of `Custom::ECSEnvironment` so it doesn't create a new physical resource id if `Properties` and `OldProperties` are identical.
4. Remove the self imposed [10 second tick](https://github.com/remind101/empire/blob/34b4ee52a5ebda6b0635b0430320a5a00a1a6b84/server/cloudformation/cloudformation.go#L93)

ejholmes:
https://github.com/remind101/empire/pull/940 will improve the performance of custom resources significantly, but it does open things up for being rate limited easily, which will cause a stack to rollback.

We should probably pass an ecs client that has more forgiving retrying.

scheduler/cloudformation/template.go
```diff
@@ -364,14 +405,38 @@ func (t *EmpireTemplate) addService(tmpl *troposphere.Template, app *scheduler.A
...
+	}
+	if taskDefinitionType == "Custom::ECSTaskDefinition" {
+		taskDefinition = fmt.Sprintf("%sTD", key)
+
+		processEnvironment := fmt.Sprintf("%sEnvironment", key)
+		tmpl.Resources[processEnvironment] = troposphere.Resource{
+			Type: "Custom::ECSEnvironment",
+			Properties: map[string]interface{}{
+				"ServiceToken": t.CustomResourcesTopic,
+				"Environment":  sortedEnvironment(p.Env),
```
ejholmes: This needs to include the `PORT` environment variable.

scheduler/cloudformation/template.go
```diff
@@ -187,19 +189,47 @@ func (t *EmpireTemplate) Build(app *scheduler.App) (*troposphere.Template, error
 }
 
 func (t *EmpireTemplate) addScheduledTask(tmpl *troposphere.Template, app *scheduler.App, p *scheduler.Process) (taskDefinition string) {
+	taskDefinitionType := taskDefinitionType(app)
```
mhahn: not a fan of this variable shadowing

scheduler/cloudformation/template.go
```diff
@@ -187,19 +189,47 @@ func (t *EmpireTemplate) Build(app *scheduler.App) (*troposphere.Template, error
...
+	cd := t.ContainerDefinition(app, p)
 	// The task definition that will be used to run the ECS task.
-	containerDefinition := containerDefinition(t.ContainerDefinition(app, p))
 	taskDefinition = fmt.Sprintf("%sTaskDefinition", key)
-	tmpl.Resources[taskDefinition] = troposphere.Resource{
-		Type: "AWS::ECS::TaskDefinition",
-		Properties: map[string]interface{}{
-			"ContainerDefinitions": []interface{}{
-				containerDefinition,
+	containerDefinition := containerDefinition(cd)
```
mhahn: same here

scheduler/cloudformation/template.go
```diff
@@ -187,19 +189,47 @@ func (t *EmpireTemplate) Build(app *scheduler.App) (*troposphere.Template, error
...
+
+	taskDefinitionProperties := map[string]interface{}{
+		"Volumes": []interface{}{},
+	}
+	if taskDefinitionType == "Custom::ECSTaskDefinition" {
```
mhahn: it looks like this block is the same as: https://github.com/remind101/empire/pull/935/files#diff-ae0d1a0ca421a9a4d5510666ef5b1c5cR411, is it worth abstracting into a function?

server/cloudformation/ecs.go
```diff
@@ -174,6 +180,235 @@ func (p *ECSServiceResource) delete(ctx context.Context, service, cluster *strin
...
+		id, err := p.register(properties, req.Hash())
+		return id, nil, err
+	default:
+		return "", nil, fmt.Errorf("%s is not supported", req.RequestType)
+	}
+}
+
+func (p *ECSTaskDefinitionResource) mergedEnvironment(ids ...string) ([]*ecs.KeyValuePair, error) {
+	var env []*ecs.KeyValuePair
+	for _, id := range ids {
```
mhahn: if we were to have more than 2 of these ever i'd say we should fetch them all at once

server/cloudformation/ecs.go
```diff
@@ -174,6 +180,235 @@ func (p *ECSServiceResource) delete(ctx context.Context, service, cluster *strin
...
+	return nil
+}
+
+// ECSEnvironmentProperties are the properties provided to the
+// Custom::ECSEnvironment custom resource.
+type ECSEnvironmentProperties struct {
+	Environment []*ecs.KeyValuePair
+}
+
+// ECSEnvironmentResource is a custom resource that takes some encrypted
```
mhahn: does this do anything with encryption yet?

server/cloudformation/ecs.go
```diff
@@ -174,6 +180,235 @@ func (p *ECSServiceResource) delete(ctx context.Context, service, cluster *strin
...
+		return id, nil, err
+	case customresources.Update:
+		id, err := p.register(properties, req.Hash())
+		return id, nil, err
+	default:
+		return "", nil, fmt.Errorf("%s is not supported", req.RequestType)
+	}
+}
+
+func (p *ECSTaskDefinitionResource) mergedEnvironment(ids ...string) ([]*ecs.KeyValuePair, error) {
```
mhahn: this seems more like `resolvedEnvironment` to me, you're resolving a list of keys into the actual values vs. merging anything

mhahn: i guess you're merging the values from multiple arrays, so this makes more sense

mhahn:
looks good, some minor comments above

ejholmes:
Ok @mhahn, think I addressed all the comments.

mhahn:
👍

Issue #918, opened 2016-07-09
Custom CloudFormation resources should gracefully handle stack creation rollbacks


ejholmes:
CloudFormations behavior when rolling back a stack creation is kinda weird. Event if the custom resource was never created, it will send a Delete request with an autogenerated PhysicalResourceId. Right now, the InstancePort and ECSService custom resources fail when this happens.

ejholmes:
To add more information, this happens when a resource **creation** fails, at which point we send an empty string as the PhysicalResourceId.

Turns out, CloudFormation doesn't like it when `PhysicalResourceId` is not provided, so it generates one automatically for it, then sets the Status Reason to `Invalid PhysicalResourceId` instead of showing the `Reason` that we provided from the error:

![](https://s3.amazonaws.com/ejholmes.github.com/twca0.png)

Now, when the stack starts rolling back, CloudFormation sends a Delete request with it's automatically generated physical resource id, and the current handles don't know what to do with that, so they fail, return an empty physical resource id and CloudFormation complains again, eventually resulting in a ROLLBACK_FAILED.

A simple repro for this is to change [this line](https://github.com/remind101/empire/blob/master/server/cloudformation/ports.go#L27) to:

```diff
+ return "", nil, errors.New("foo")
```

Then deploy to a new app that will cause a new stack to get created:

```console
$ emp create foo
$ emp deploy remind101/acme-inc:master -a foo
```

I'll have a PR to fix these issues in a sec.

PR #938, opened 2016-07-16
Gracefully handle creation failures of custom resources

ejholmes:
Fixes https://github.com/remind101/empire/issues/918

See https://github.com/remind101/empire/issues/918 for an explanation of what was causing the issue.

This does a few things:

1. Fixes the issue in https://github.com/remind101/empire/issues/918
2. Pulls out some of the custom resources framework in `server/cloudformation`, to `pkg/cloudformation/customresources`.

End result is that, if creating a new resource fails, the correct error will appear in the CloudFormation events, and request to delete it during the rollback will succeed.


phobologic:
👍

PR #939, opened 2016-07-18
Set deployment message from GitHub deployments.

ejholmes:
Before this change, requiring messages would cause GitHub deployments to fail. This will add the description from the GitHub deployment as the message, if it's provided, otherwise it'll default to a message including the GitHub deployment id.


mhahn:
👍

PR #940, opened 2016-07-19
Improve performance of provisioning of custom cloudformation resources

ejholmes:
This does a few things, talked about in https://github.com/remind101/empire/pull/935#issuecomment-233507474

1. Removes the self imposed 10 second tick.
2. Starts up multiple worker goroutines pulling messages off of sqs.
3. Makes it so that updates to ECSEnvironment and ECSTaskDefinition don't create a new physical resource when the properties are identical as before

For an app that used to take over 10 minutes to perform the updates, this reduces it down to less than 2 minutes.

### DIFF
--- a/server/cloudformation/cloudformation.go
+++ b/server/cloudformation/cloudformation.go
@@ -79,13 +79,13 @@ func NewCustomResourceProvisioner(db *sql.DB, config client.ConfigProvider) *Cus
 	})
 
 	store := &dbEnvironmentStore{db}
-	p.add("Custom::ECSEnvironment", &ECSEnvironmentResource{
+	p.add("Custom::ECSEnvironment", newECSEnvironmentProvisioner(&ECSEnvironmentResource{
 		environmentStore: store,
-	})
-	p.add("Custom::ECSTaskDefinition", &ECSTaskDefinitionResource{
+	}))
+	p.add("Custom::ECSTaskDefinition", newECSTaskDefinitionProvisioner(&ECSTaskDefinitionResource{
 		ecs:              ecs.New(config),
 		environmentStore: store,
-	})
+	}))
 
 	return p
 }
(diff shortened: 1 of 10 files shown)


server/cloudformation/cloudformation.go
```diff
@@ -99,9 +99,13 @@ func (c *CustomResourceProvisioner) add(resourceName string, p customresources.P
 
 // Start starts pulling requests from the queue and provisioning them.
 func (c *CustomResourceProvisioner) Start() {
-	t := time.Tick(10 * time.Second)
+	for i := 0; i < 10; i++ {
```
ejholmes: This should be configurable.

server/cloudformation/cloudformation.go
```diff
@@ -224,3 +228,67 @@ func (c *CustomResourceProvisioner) provision(ctx context.Context, m Message, re
...
+	Create func(context.Context, customresources.Request) (string, interface{}, error)
+	Update func(context.Context, customresources.Request) (interface{}, error)
+	Delete func(context.Context, customresources.Request) error
+}
+
+func (p *provisioner) Properties() interface{} {
+	return p.properties()
+}
+
+func (p *provisioner) Provision(ctx context.Context, req customresources.Request) (string, interface{}, error) {
```
ejholmes: This basically just abstracts the commonality between all of the resource provisioners. Only the new `ECSEnvironment` and `ECSTaskDefinition` resources are using this so far. Eventually, we should move the `ECSService` and `InstancePort` resources over to this as well.

mhahn:
👍

ejholmes:
I'll remove 1 and 2 above, since it opens things up for rate limiting, and then merge this into https://github.com/remind101/empire/pull/935.

I'll do a separate PR for that, since there's a lot of little considerations that need to happen, and affects more than just the custom ecs task definition resource.

PR #941, opened 2016-07-20
Move ECSService and InstancePort resources to higher level interface.

ejholmes:
Follow up from https://github.com/remind101/empire/pull/935

This moves the ECSService and InstancePort custom resources to use the higher level interface.


mhahn:
👍

PR #942, opened 2016-07-20
Improve performance of custom resource provisioning

ejholmes:
This does the following to improve performance of custom resource provisioning:

1. Multiple goroutines pulling messages off of sqs are started.
2. The self imposed 10 second tick was removed.

Because of this, there's an increased probability of hitting rate limits. To counteract that, the following was done:

1. The ECS client that the `ECS*` custom resources use has a more lenient request retryer, retrying up to 10 times (default of 3).
2. Because throttling retries with exponential backoff could make a request take longer than the default queue visibility timeout, the sqs dispatcher now periodically extends the visibility timeout of received messages.

The ECS client will retry throttled requests, up to maximum wait time of ~13 minutes (previously, ~3 seconds). The backoff of a request, that retried up to 10 times, would look like this:

```
        ecs_test.go:41: delay(0): 873ms
        ecs_test.go:41: delay(1): 1.364s
        ecs_test.go:41: delay(2): 2.252s
        ecs_test.go:41: delay(3): 4.896s
        ecs_test.go:41: delay(4): 12.352s
        ecs_test.go:41: delay(5): 30.56s
        ecs_test.go:41: delay(6): 35.648s
        ecs_test.go:41: delay(7): 1m43.808s
        ecs_test.go:41: delay(8): 3m21.216s
        ecs_test.go:41: delay(9): 2m12.352s
        ecs_test.go:43: total(min): 6m23.5s
        ecs_test.go:44: total(max): 12m47s
        ecs_test.go:45: total(real): 8m45.321s
```

### DIFF
--- a/server/cloudformation/queue.go
+++ b/server/cloudformation/queue.go
@@ -1,6 +1,8 @@
 package cloudformation
 
 import (
+	"fmt"
+	"sync"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -11,8 +13,8 @@ import (
 )
 
 var (
-	DefaultVisibilityHeartbeat = 1 * time.Minute
-	DefaultNumWorkers          = 10
+	defaultVisibilityHeartbeat = 1 * time.Minute
+	defaultNumWorkers          = 10
 )
 
 // sqsClient duck types the sqs.SQS interface.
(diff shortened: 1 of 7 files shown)


server/cloudformation/queue.go
```diff
@@ -0,0 +1,127 @@
...
+			go func(m *sqs.Message) {
+				if err := q.handle(ctx, handle, m); err != nil {
+					reporter.Report(ctx, err)
+				}
+			}(m)
+		}
+	}
+}
+
+func (q *SQSDispatcher) handle(ctx context.Context, handle func(context.Context, *sqs.Message) error, message *sqs.Message) (err error) {
```
ejholmes: This could use some tests.

server/cloudformation/queue.go
```diff
@@ -0,0 +1,127 @@
...
+func newSQSDispatcher(config client.ConfigProvider) *SQSDispatcher {
+	return &SQSDispatcher{
+		VisibilityHeartbeat: DefaultVisibilityHeartbeat,
+		NumWorkers:          DefaultNumWorkers,
+		sqs:                 sqs.New(config),
+	}
+}
+
+// Start starts multiple goroutines pulling messages off of the queue.
+func (q *SQSDispatcher) Start(handle func(context.Context, *sqs.Message) error) {
```
ejholmes: This should really block until all of the started goroutines are stopped.

server/cloudformation/queue.go
```diff
@@ -0,0 +1,161 @@
...
+			t, err = q.extendMessageVisibilityTimeout(message.ReceiptHandle)
+			if err != nil {
+				return
+			}
+		}
+	}
+}
+
+// extendMessageVisibilityTimeout extends the messages visibility timeout
+// extends the timeout by VisibilityHeartbeat, and returns a channel that will
```
mhahn: typo

mhahn:
👍

Issue #945, opened 2016-07-21
Long wait times between updates in the deployment stream can cause idle connection timeouts?


ejholmes:
When deploying with the `-s` flag, if there are no updates for a while, you'll see something like this:

```
Status: Stack update submitted
Status: Waiting for stack update to complete
error: unexpected EOF
```

My guess is this is limited to ELB, because we don't send any heartbeats in the `/deploys` endpoint. I'll dig a little deeper and try to get a simple repro.

ejholmes:
Increasing the idle connection timeout on the load balancer to 3600 (1 hour) fixed the issue. I think this should probably be our recommendation, and remove the [heartbeats from `emp run`](https://github.com/remind101/empire/blob/master/server/heroku/processes.go#L111). If no data is sent within an hour, then closing the connection is a reasonable thing to do (and it will cause the docker container to SIGKILL).

PR #947, opened 2016-07-21
Example CloudFormation improvements

ejholmes:
Closes https://github.com/remind101/empire/issues/945

These are just some minor improvements to the demo/dev stack, that I made while debugging https://github.com/remind101/empire/issues/945.

1. The idle connection timeout on the ELB created for Empire is now set to 1 hour.
2. The scheduling backend is configurable through a parameter.
3. The ECS AMI was updated to most recent version (Docker 1.11.1, ECS agent 1.11)
4. Logs from the Empire daemon and and Empire applications are sent to CloudWatch logs for easier debugging.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,7 +1,0 @@
-#!/bin/bash
-
-if [ ! -z "$WAIT_TIME" ]; then
-  sleep "$WAIT_TIME"
-fi
-
-exec /go/bin/empire $@
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+if [ ! -z "$WAIT_TIME" ]; then
+  sleep "$WAIT_TIME"
+fi
+
+exec /go/bin/empire $@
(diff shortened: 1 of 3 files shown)


entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+if [ ! -z "$WAIT_TIME" ]; then
```
ejholmes: For whatever reason, updating to the new ECS AMI started making Empire start a lot faster than the postgres container, which means it couldn't connect and would just get stuck in a restart loop. I've seen it happen previously, but seems to happen consistently with the new AMI.

I'm not super happy adding this. I think the two options we have are to do this, or remove the postgres container and replace it with an RDS instance. The only reason we haven't done that is that RDS takes a lot of time to create.

phobologic: Maybe put a comment in the script to let people know why it exists, otherwise I'm guessing we'll end up with PRs to remove this :)

phobologic:
👍

mhahn:
@ejholmes would you have the RDS instance exposed publicly?

ejholmes:
@mhahn It'd be created within the VPC.

ejholmes:
I just added a commit that switches over to a small RDS instance in the VPC. I think this might be better in the long run, since it'll also solve the ephemeral nature of using postgres in a container.

mhahn:
👍

PR #950, opened 2016-07-25
Send heartbeats to Tugboat

ejholmes:
This fixes an issue in the Tugboat integration where, if Tugboat was hosted on something like Heroku, the log stream to Tugboat could be closed if the deployment stream had a gap where no logs were written. This writes the null character every 10 seconds, to keep the connection open, which [Tugboat already strips out](https://github.com/remind101/tugboat/blob/9655efe1fc24323ec3155fe43e7446dc2264e00c/logs.go#L114).


phobologic:
👍

PR #952, opened 2016-07-26
Issue 697 v3

mhahn:
Related https://github.com/remind101/empire/issues/697

Uses timestamp instead of uuid. I had to create a new PR since there were a lot of conflicts with master.

### DIFF
--- a/scheduler/cloudformation/cloudformation.go
+++ b/scheduler/cloudformation/cloudformation.go
@@ -198,7 +198,7 @@ func (s *Scheduler) Restart(ctx context.Context, app *scheduler.App, ss schedule
 	if err != nil {
 		return err
 	}
-	output := make(chan stackOperationOutput)
+	output := make(chan stackOperationOutput, 1)
 	return s.updateStack(ctx, &updateStackInput{
 		StackName: aws.String(stackName),
 		Parameters: []*cloudformation.Parameter{
--- a/scheduler/cloudformation/cloudformation.go
+++ b/scheduler/cloudformation/cloudformation.go
@@ -11,10 +11,10 @@ import (
 	"hash/crc32"
 	"html/template"
 	"io"
+	"strconv"
 	"strings"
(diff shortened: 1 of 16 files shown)


scheduler/cloudformation/cloudformation.go
```diff
@@ -192,6 +193,23 @@ func (s *Scheduler) SubmitWithOptions(ctx context.Context, app *scheduler.App, s
 	return tx.Commit()
 }
 
+func (s *Scheduler) Restart(ctx context.Context, app *scheduler.App, ss scheduler.StatusStream) error {
+	stackName, err := s.stackName(app.ID)
+	if err != nil {
+		return err
+	}
+	output := make(chan stackOperationOutput)
```
ejholmes: I think this should be buffered so that the goroutine that sends on this doesn't block forever (runaway goroutines).

```go
make(chan stackOperationOutput, 1)
```

scheduler/cloudformation/cloudformation.go
```diff
@@ -192,6 +193,23 @@ func (s *Scheduler) SubmitWithOptions(ctx context.Context, app *scheduler.App, s
...
+	return s.updateStack(ctx, &updateStackInput{
```
ejholmes: We may want to do the same thing that we do in `Submit` and wait if `ss != nil`. Otherwise, a consumer that provides a StatusStream, doesn't really know when the Restart has finished.

mhahn: yea, i agree. i think for set/restart etc. we should wait, i'll do that in another PR

ejholmes:
I removed `Fixes` from the PR comment, since this just gets us part of the way to fixing the issue in https://github.com/remind101/empire/issues/697.

ejholmes:
Awesome, looks gtm 👍

mhahn:
ah ok cool.

PR #953, opened 2016-07-26
Metrics

ejholmes:
This sets up a small stats framework, and adds some middleware to all AWS clients that will increment some counters for aws requests and responses. This will allow us to get better visibility into throttling, or other issues.

We can also start using the `stats.*` methods to generate metrics for the other things, like number of deploys, how long deploys take, etc.

The stats package is just a small layer of abstraction over a statsd'ish client, and I've only provided a datadog implementation for now.

**Metrics we should have**

* [x] `aws.request.*` For outbound AWS requests
* [x] `heroku.request.*` For inbound Heroku API requests.
* [x] `runtime.*` Golang runtime stats.
* [ ] `heroku.auth.error` for authorization errors.
* [ ] `core.*` for core API calls.
* [ ] `scheduler.cloudformation.template_size` For tracking generated template sizes


phobologic:
Would be cool if we had an `errors` metric that we could tag with different types of errors/urls, etc.

ejholmes:
Good call, I can add that to the embedded error reporter 👍

cmd/empire/factories.go
```diff
@@ -418,13 +380,34 @@ func newHBReporter(key, env string) (reporter.Reporter, error) {
...
+
+func newStats(c *Context) (stats.Stats, error) {
+	if addr := c.String(FlagDogStatsd); addr != "" {
+		return newDataDogStats(addr)
+	}
+
+	return stats.Null, nil
+}
+
+func newDataDogStats(addr string) (stats.Stats, error) {
```
phobologic: It might be worth at least having a datadog & plain statsd implementation at first.  I think the only difference would be removing tags?

ejholmes: Cool, yeah, I can add a plain statsd backend as well.

cmd/empire/factories.go
```diff
@@ -445,3 +428,48 @@ func newAuthProvider(c *cli.Context) (dockerauth.AuthProvider, error) {
...
+	s.Handlers.ValidateResponse.PushBackNamed(request.NamedHandler{
+		Name: "empire.ResponseMetrics",
+		Fn: func(r *request.Request) {
+			tags := requestTags(r)
+			if r.Error != nil {
+				if err, ok := r.Error.(awserr.Error); ok {
+					tags = append(tags, fmt.Sprintf("error:%s", err.Code()))
+					stats.Inc(fmt.Sprintf("aws.request.error"), 1, 1.0, tags)
+					stats.Inc(fmt.Sprintf("aws.request.%s.error", r.ClientInfo.ServiceName), 1, 1.0, tags)
+					stats.Inc(fmt.Sprintf("aws.request.%s.%s.error", r.ClientInfo.ServiceName, r.Operation.Name), 1, 1.0, tags)
```
phobologic: nm on the error stat :) sorry for jumping the gun

phobologic:
So awesome 👍

ejholmes:
Added a commit that adds a statsd backend as well, and also changed the flag to `--stats`, so you can configure it like this:

```console
empire server --stats=statsd://localhost:8125
```

Or

```console
empire server --stats=dogstatsd://localhost:8126
```

ejholmes:
Also added some sample of the go runtime metrics every 30 seconds as well.

PR #956, opened 2016-07-27
Wait 10 seconds for sqs.ReceiveMessage

ejholmes:
This should significantly reduce the number of sqs.ReceiveMessage calls we make.

/cc @phobologic


phobologic:
👍

PR #957, opened 2016-07-27
More metrics

ejholmes:
Adds a metric for tracking cloudformation template sizes, and also adds metrics for tracking provisioning time of cloudformation custom resources.


phobologic:
👍

PR #959, opened 2016-07-28
remove ps args for app names

### LINKED ISSUE #880: Behavior of `emp ps` doesn't work as stated
`emp ps` says that you can pass it the name of the process to list and it will only list that process within the app.

This looks to be broken since we prepend the version number to the process name: https://github.com/remind101/empire/blob/master/tasks.go#L51.

ie.

```
$ ./build/emp ps -a acme-inc worker
$ ./build/emp ps -a acme-inc v1
v1.web.dfb422e5-5187-4a73-9ff1-987c66486451     1X  RUNNING  24h  "acme-inc server"
v1.worker.80f8d14e-f0c2-4bfe-968f-fdf0a3d49bee  1X  RUNNING  24h  "acme-inc worker"
```

hobozhang:
#880 so people should use standard unix tools. 

Closes #880

### DIFF
--- a/cmd/emp/dynos.go
+++ b/cmd/emp/dynos.go
@@ -15,7 +15,7 @@ import (
 
 var cmdDynos = &Command{
 	Run:      runDynos,
-	Usage:    "ps [<name>...]",
+	Usage:    "ps",
 	Alias:    "dynos",
 	NeedsApp: true,
 	Category: "dyno",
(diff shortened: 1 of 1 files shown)


cmd/emp/dynos.go
```diff
@@ -29,50 +29,30 @@ Examples:
     run.3794  2X  up   1m  bash
     web.1     1X  up  15h  "blog /app /tmp/dst"
```
ejholmes: While we're at it. Can we update the examples here to be a more realistic?

```
    v1.run.e97e1f75e8ff  2X  RUNNING   1m  bash
    v1.web.dcc9a8c4-c0f8-4478-aa8a-f9148b362401     1X  RUNNING  15h  "blog /app /tmp/dst"
    v1.web.2bcb6e08-ef99-447f-8e7a-416d94769010     1X  RUNNING   8h  "blog /app /tmp/dst"
```

ejholmes:
Awesome! lgtm 👍

Issue #960, opened 2016-07-28
Metric for AWS request errors shows incorrect error code


ejholmes:
https://github.com/remind101/empire/blob/9f56d622418701d14d9e63f8f193a1ee91d6f4c3/cmd/empire/factories.go#L476-L489

Always shows up as `UnknownError`, instead of something like `ThrottlingError`.

Might need to be PushFront instead of PushBack.

PR #961, opened 2016-07-29
Tag aws error stats with correct error code

ejholmes:
Fixes https://github.com/remind101/empire/issues/960.

Turns out, the `Error` attribute on a `request.Request` isn't actually real until the  `UnmarshalError` handler is called. The `Retry` handlers are called immediately after `UnmarshalError`, so this seems like the best place to put it.


phobologic:
👍

PR #962, opened 2016-07-29
Fix extension of message visibility.

ejholmes:
https://github.com/remind101/empire/pull/942 introduced a bug, where if the initial call to extend the message visibility failed (e.g. AccessDenied), the error would not be bubbled up.


phobologic:
👍

PR #963, opened 2016-07-30
Time how long stack creates/updates take.

ejholmes:
This will add an `empire.scheduler.cloudformation.UpdateStack` and `empire.scheduler.cloudformation.CreateStack` metric which tracks how long it takes to update/create a stack.

Also does a little refactoring to normalize the waiters for create/update.


phobologic:
Any need for tests here, or is most of this functionality covered by other tests?  otherwise, 👍

mhahn:
this is sweet! 👍

ejholmes:
Current tests should cover both cases here.

PR #965, opened 2016-08-04
Add empire.user label to runs

ejholmes:
Sometimes, it would be helpful to know who started an attached/detached run. This will add an `empire.user` label that can be used with `docker inspect`, or with monitoring tools to determine who started the process.

/cc @phobologic


phobologic:
These are almost too easy ;) 👍 Awesome, this is going to help a lot.

Issue #966, opened 2016-08-08
Better security around `emp run`


ejholmes:
`emp run` grants a lot of power and flexibility, but can become difficult to track what people are doing.

It may be nice to provide an option to whitelist certain processes in the extended Procfile as "runnable" processes.

So instead of allowing any command to be run, you would add a command to `Procfile` and that command could then be run with `emp run`. For example:

```yaml
migrate:
  command: bundle exec rake db:migrate:up
  run: true
console:
  command: bundle exec rails console
  run: true
psql:
  command: ./bin/psql
  run: true
```

Doing this would allow you to whitelist "safe" commands, and disallow things like `emp run bash`.

Obviously, this isn't perfect, since if you allow things like `rails console`, someone can just `Kernel.system`, so this would be mostly there to prevent accidents, and provide a documented list of "operational" commands for the service.

ejholmes:
https://github.com/remind101/empire/pull/967 was merged. Gonna close this.

PR #967, opened 2016-08-09
Only allow whitelisted commands to be run

ejholmes:
This is a spike into https://github.com/remind101/empire/issues/966, which adds a configuration option to Empire to require commands to be whitelisted for `emp run` to work.

If a command is not whitelisted, `emp run` will have this behavior:

```console
$ emp run migrate
error: command not whitelisted: migrate
```

To whitelist a command, you add it to the Procfile:

```yaml
migrate: bundle exec rake db:migrate:up
```

It's recommended that you use the extended Procfile and set the `noservice: true` flag to prevent it from being scaled up (and it won't create any AWS resources).

```yaml
migrate:
  command: bundle exec rake db:migrate:up
  noservice: true
```

In this initial spike, it allows any command in the Procfile to be run. For example, if you use the new scheduled tasks, you can spin up a one off of the scheduled task easily like so:

```yaml
scheduled-job:
  command: ./bin/do-work
  cron: * * * * * *
```

```console
$ emp run scheduled-job
doing work
```

Any extra arguments will be appended to the command provided in the procfile. For example:

```yaml
migrate:
  command: ./bin/migrate
  noservice: true
```

```shell
exec bundle exec rake db:migrate:$1
```

```console
$ emp run migrate up
```

**TODO**

* Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,8 @@
 
 **Security**
 
+* Empire now has a new `commands.allowed` flag that controls the behavior of what commands are allowed with `emp run`. This can be set to `procfile` to limit `emp run` to only allow commands defined in the Procfile.
+
 ## 0.10.1 (2016-06-14)
 
 **Features**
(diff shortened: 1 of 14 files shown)


phobologic:
👍 positional args would be awesome.

Issue #968, opened 2016-08-11
Use Application Load Balancers


ejholmes:
https://aws.amazon.com/blogs/aws/powerful-aws-platform-features-now-for-containers/

Thanks to the new cloudformation backend, this should be a pretty simple swap.

PR #969, opened 2016-08-11
Switch to Application Load Balancers

ejholmes:
Closes https://github.com/remind101/empire/issues/968

This is mostly a POC at this point, since I haven't done any work to support both classic ELB's and new ALB's so that you can easily swap between both in case there are issues.

But, this does seem to work. The advantage of using ALB's is primarily that we no longer need to allocate and manage instance ports, which means multiple "web" procs can run on the same EC2 instance.

The first commit just updates the aws-sdk-go dependency, and the second includes the actual changes that switch to ALB's.


phobologic:
Looks good - really should have the ability to switch between old ELB and new ALB at least, just in case there are unforeseen issues with the new system.

ejholmes:
Alright. By default, classic ELB's will be used, but you can now (experimentally) switch to the new application load balancers by setting `LOAD_BALANCER_TYPE=alb` on the app.

Updated the changelog as well.

phobologic:
Awesome. 👍

PR #973, opened 2016-08-16
Remove pkg/httpx dependency

ejholmes:
Tests may fail right now, which is because of an unrelated change and fixed in https://github.com/remind101/empire/pull/972

Depends on https://github.com/remind101/empire/pull/971

This removes the pkg/httpx dependency, and moves all internal packages to use `context` from the stdlib, rather than from `golang.org/x/net/context`, and updates all http handlers to implement `http.Handler` rather than `httpx.Handler`.

First commit is just changes to `./vendor` (remove `pkg/httpx` and update `gorilla/mux` to support Go 1.7 context).

Second commit is the actual changes.


server/heroku/heroku.go
```diff
@@ -87,47 +91,59 @@ func New(e *empire.Empire) *Server {
...
-	h := Authenticate(s.mux, s.Authenticator)
-
-	err := h.ServeHTTPContext(ctx, w, r)
-	if err != nil {
-		Error(w, err, http.StatusInternalServerError)
-		reporter.Report(ctx, err)
-	}
-
-	return nil
+// ServeHTTPContext implements the http.Handler interface.
```
mhahn: nit: comment doesn't match

ejholmes:
Gonna close this one for now. I'll move this over incrementally later.

PR #978, opened 2016-08-20
Switch default scheduler to CloudFormation for new apps.

ejholmes:
This is prep for the 0.11.0 release, [where we'll switch the default scheduling backend for new apps to CloudFormation](https://github.com/remind101/empire/wiki/Scheduler-Deprecation-Plan#0110)

Existing applications will continue to use the legacy ECS backend, but are encouraged to migrate to CloudFormation using the [Scheduler Migration Guide](https://github.com/remind101/empire/wiki/Scheduler-Migration-Guide).


phobologic:
👍

Issue #976, opened 2016-08-18
Show host a task is running on in `emp ps`


ejholmes:
This would be really easy to do, and would be helpful for debugging.

Nosajool:
Unfortunately, the host isn't a field on the `Task` so it seems like we'd have to make another request. One way to do this is to feed the task's `ContainerInstanceArn`s to `func (*ECS) DescribeContainerInstances` Then we'd have access to the container's `Ec2InstanceId`.

http://docs.aws.amazon.com/sdk-for-go/api/service/ecs/#Task
```go
type Task struct {
    // The Amazon Resource Name (ARN) of the cluster that hosts the task.
    ClusterArn *string `locationName:"clusterArn" type:"string"`
    // The Amazon Resource Name (ARN) of the container instances that host the task.
    ContainerInstanceArn *string `locationName:"containerInstanceArn" type:"string"`
    // The containers associated with the task.
    Containers []*Container `locationName:"containers" type:"list"`
    // The Unix timestamp for when the task was created (the task entered the PENDING
    // state).
    CreatedAt *time.Time `locationName:"createdAt" type:"timestamp" timestampFormat:"unix"`
    // The desired status of the task.
    DesiredStatus *string `locationName:"desiredStatus" type:"string"`
    // The last known status of the task.
    LastStatus *string `locationName:"lastStatus" type:"string"`
    // One or more container overrides.
    Overrides *TaskOverride `locationName:"overrides" type:"structure"`
    // The Unix timestamp for when the task was started (the task transitioned from
    // the PENDING state to the RUNNING state).
    StartedAt *time.Time `locationName:"startedAt" type:"timestamp" timestampFormat:"unix"`
    // The tag specified when a task is started. If the task is started by an Amazon
    // ECS service, then the startedBy parameter contains the deployment ID of the
    // service that starts it.
    StartedBy *string `locationName:"startedBy" type:"string"`
    // The Unix timestamp for when the task was stopped (the task transitioned from
    // the RUNNING state to the STOPPED state).
    StoppedAt *time.Time `locationName:"stoppedAt" type:"timestamp" timestampFormat:"unix"`
    // The reason the task was stopped.
    StoppedReason *string `locationName:"stoppedReason" type:"string"`
    // The Amazon Resource Name (ARN) of the task.
    TaskArn *string `locationName:"taskArn" type:"string"`
    // The Amazon Resource Name (ARN) of the task definition that creates the task.
[...]

PR #983, opened 2016-08-23
Show host the task is running on in `emp ps`

Nosajool:
Closes https://github.com/remind101/empire/issues/976

```bash
$ EMPIRE_API_URL=http://$(docker-machine ip default):8080 ./build/emp ps -a acme-inc
v2.web.1a1a11a1-aaa1-1aaa-a11a-aaa111a1aa1a  1b1b11b1-bbb1-1bbb-b11b-bbb111b1bb1b  1c1c11c1-ccc1-1ccc-c11c-ccc111c1cc1c  i-1111a111  1X  RUNNING   6m  "acme-inc server"
```

Opening this since I'd like some feedback, however I don't think this is the right approach since it requires a request to `DescribeContainerInstances` **for each** process which will quickly rate limit us. Looking into if it is possible to just make one batch request for all of the containers. If not, might need to cache like in https://github.com/remind101/empire/pull/902.

Also, the output is getting pretty wide so I was wondering if maybe this stuff should be only shown if you pass the `-v` flag to `emp ps`. Then we could add other information here like @sanjayprabhu's "commit message" request. https://remind.slack.com/archives/empire/p1471910137001348

### New output
```bash
v2.web.1a1a11a1-aaa1-1aaa-a11a-aaa111a1aa1a  i-aa111aa1  1X  RUNNING  89m  "acme-inc server"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## HEAD
+
+**Improvements**
+
+* `emp ps` now displays the task's host. [#983](https://github.com/remind101/empire/pull/983)
+
 ## 0.11.0 (2016-08-22)
 
 **Features**
@@ -24,7 +30,6 @@
 * The Scheduler now has a `Restart` method which will trigger a restart of all the processes within an app. Previously, a "Restart" just re-released the app. Now schedulers like the cloudformation backend can optimize how the restart is handled. [#697](https://github.com/remind101/empire/issues/697)
 * `emp run` now publishes an event when it is ran. [#954](https://github.com/remind101/empire/pull/954)
 * `emp rollback` requires confirmation if rolling back more than 9 versions. [#975](https://github.com/remind101/empire/pull/975)
-* `emp ps` now displays the task's host. [#983](https://github.com/remind101/empire/pull/983)
 
 **Bugs**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 * The Scheduler now has a `Restart` method which will trigger a restart of all the processes within an app. Previously, a "Restart" just re-released the app. Now schedulers like the cloudformation backend can optimize how the restart is handled. [#697](https://github.com/remind101/empire/issues/697)
 * `emp run` now publishes an event when it is ran. [#954](https://github.com/remind101/empire/pull/954)
 * `emp rollback` requires confirmation if rolling back more than 9 versions. [#975](https://github.com/remind101/empire/pull/975)
+* `emp ps` now displays the task's host. [#983](https://github.com/remind101/empire/pull/983)
 
 **Bugs**
 
--- a/cmd/emp/dynos.go
+++ b/cmd/emp/dynos.go
@@ -21,7 +21,7 @@ var cmdDynos = &Command{
 	Category: "dyno",
 	Short:    "list processes",
 	Long: `
-Lists processes. Shows the name, size, id, container id, ec2 instance id, state, age, and command.
+Lists processes. Shows the name, size, host, state, age, and command.
 
 Examples:
 
@@ -58,9 +58,7 @@ func listDynos(w io.Writer) {
 func listDyno(w io.Writer, d *heroku.Dyno) {
 	listRec(w,
 		d.Name,
-		d.Id,
-		d.ContainerInstanceID,
-		d.EC2InstanceID,
+		d.Host.Id,
 		d.Size,
 		d.State,
 		prettyDuration{dynoAge(d)},
--- a/cmd/emp/dynos.go
+++ b/cmd/emp/dynos.go
@@ -21,7 +21,7 @@ var cmdDynos = &Command{
 	Category: "dyno",
 	Short:    "list processes",
 	Long: `
-Lists processes. Shows the name, size, state, age, and command.
+Lists processes. Shows the name, size, id, container id, ec2 instance id, state, age, and command.
 
 Examples:
 
@@ -58,6 +58,9 @@ func listDynos(w io.Writer) {
 func listDyno(w io.Writer, d *heroku.Dyno) {
 	listRec(w,
 		d.Name,
+		d.Id,
+		d.ContainerInstanceID,
+		d.EC2InstanceID,
 		d.Size,
 		d.State,
 		prettyDuration{dynoAge(d)},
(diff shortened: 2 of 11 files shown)


cmd/emp/dynos.go
```diff
@@ -58,6 +58,9 @@ func listDynos(w io.Writer) {
 func listDyno(w io.Writer, d *heroku.Dyno) {
 	listRec(w,
 		d.Name,
+		d.Id,
+		d.ContainerInstanceID,
```
ejholmes: In general, the container instance id isn't very useful, and it's easy to correlate in AWS if we have the EC2 instance id. I think we should leave this off for the first version.

cmd/emp/dynos.go
```diff
@@ -58,6 +58,9 @@ func listDynos(w io.Writer) {
...
+		d.EC2InstanceID,
```
ejholmes: Just to keep this de-coupled from AWS specifics, let's call this `Instance` or `Host` (think I prefer `Host`). Could also make this a struct, so we can expand on it later:

```go
type Host struct {
        Id string `json:"id"`
}

type Dyno struct {
....
        Host Host `json:"host"`
....
}
```

scheduler/cloudformation/cloudformation.go
```diff
@@ -706,12 +708,29 @@ func (s *Scheduler) Instances(ctx context.Context, app string) ([]*scheduler.Ins
 			}
 			taskDefinitions[k] = resp.TaskDefinition
 		}
+
+		k = *t.ContainerInstanceArn
+
+		if _, ok := containerInstances[k]; !ok {
+			resp, err := s.ecs.DescribeContainerInstances(&ecs.DescribeContainerInstancesInput{
+				Cluster:            t.ClusterArn,
+				ContainerInstances: []*string{t.ContainerInstanceArn},
```
ejholmes: We can optimize this better so we make less calls:

1. Iterate through the tasks, and add the container instance arn's to a map.
2. Chunk them to whatever the maximum number of container instances is for `DescribeContainerInstances`, and describe each chunk (the [docs](http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeContainerInstances.html) don't seem to specify what the limit is). Add that to another map, where the key is the container instance arn, and the value is the instance id.

scheduler/cloudformation/cloudformation.go
```diff
@@ -721,6 +740,8 @@ func (s *Scheduler) Instances(ctx context.Context, app string) ([]*scheduler.Ins
 			return instances, err
 		}
 
+		ec2InstanceID := aws.StringValue(containerInstances[*t.ContainerInstanceArn].Ec2InstanceId)
```
ejholmes: Awesome. Didn't know about about `aws.StringValue`. We can replace the `safeString` method with calls to this instead (doesn't need to be in this PR).

ejholmes:
Other than those comments, this is awesome. This will be really useful!

cmd/emp/dynos.go
Nosajool: Let me know if I should modify https://github.com/remind101/empire/blob/master/pkg/heroku/gen/schema.json instead and generate the change

scheduler/cloudformation/cloudformation.go
Nosajool:
```bash
$ EMPIRE_API_URL=http://$(docker-machine ip default):8080 ./build/emp ps -a acme-inc
error: InvalidParameterException: instanceIds can have at most 100 items.
        status code: 400, request id: 605f8aa8-6b18-11e6-b1e3-1f48c5247aba
```

scheduler/cloudformation/cloudformation.go
```diff
@@ -708,6 +710,36 @@ func (s *Scheduler) Instances(ctx context.Context, app string) ([]*scheduler.Ins
...
+		for _, chunk := range chunkStrings(containerArnPtrs, MaxDescribeContainerInstances) {
+			resp, err := s.ecs.DescribeContainerInstances(&ecs.DescribeContainerInstancesInput{
+				Cluster:            aws.String(clusterArn),
+				ContainerInstances: chunk,
+			})
+			if err != nil {
+				return nil, err
+			}
+			for _, f := range resp.Failures {
+				return nil, fmt.Errorf("error describing container instance Arn=%s err=%s", aws.StringValue(f.Arn), aws.StringValue(f.Reason))
```
ejholmes: Minor nit, can we make this error:

```go
fmt.Errorf("error describing container instance %s: %s", aws.StringValue(f.Arn), aws.StringValue(f.Reason))
```

tests/cli/restart_test.go
```diff
@@ -17,8 +17,8 @@ func TestRestart(t *testing.T) {
 		},
 		{
 			"dynos -a acme-inc",
-			`v1.web.1  1X  running   5d  "./bin/web"
-v1.web.2  1X  running   5d  "./bin/web"`,
+			`v1.web.1    1X  running   5d  "./bin/web"
```
ejholmes: Can we also update the [fake scheduler to add a fake host](https://github.com/remind101/empire/blob/master/scheduler/fake.go#L55) to the returned set of instances? Maybe just some string that looks like an ec2 instance id.

ejholmes:
Just a couple small nits, but overall this looks good to go. Just needs a rebase since https://github.com/remind101/empire/pull/985 was merged.

cmd/emp/dynos.go
ejholmes: In the past, we've just altered the Go types directly, which I think is ok. We'll _eventially_ move away from re-using the Heroku API, but adding small things like this is fine.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
 * The Scheduler now has a `Restart` method which will trigger a restart of all the processes within an app. Previously, a "Restart" just re-released the app. Now schedulers like the cloudformation backend can optimize how the restart is handled. [#697](https://github.com/remind101/empire/issues/697)
 * `emp run` now publishes an event when it is ran. [#954](https://github.com/remind101/empire/pull/954)
 * `emp rollback` requires confirmation if rolling back more than 9 versions. [#975](https://github.com/remind101/empire/pull/975)
+* `emp ps` now displays the task's host. [#983](https://github.com/remind101/empire/pull/983)
```
ejholmes: Oops, this should get added to `## HEAD` since 0.11.0 was already released.

PR #984, opened 2016-08-24
Add support for configuring the IAM role used for the application

ejholmes:
This should be considered experimental for the time being. It allows you to set the IAM role used for an application.

There's still some gotcha's with using this. The primary one being that, if you set the role, attached `emp run's` won't get the role applied, since it's handled by the ECS agent (might work if the ECS agent is running on the host? Need to test).

### DIFF
--- a/scheduler/cloudformation/template.go
+++ b/scheduler/cloudformation/template.go
@@ -29,10 +29,55 @@ var (
 )
 
 // Load balancer types
-var (
+const (
 	classicLoadBalancer     = "elb"
 	applicationLoadBalancer = "alb"
 )
+
+// Returns the type of load balancer that should be used (ELB/ALB).
+func loadBalancerType(app *scheduler.App) string {
+	check := []string{
+		"EMPIRE_X_LOAD_BALANCER_TYPE",
+		"LOAD_BALANCER_TYPE", // For backwards compatibility.
+	}
+
+	for _, n := range check {
+		if v, ok := app.Env[n]; ok {
+			return v
+		}
+	}
+
+	// Default when not set.
+	return classicLoadBalancer
+}
+
+// Returns the name of the CloudFormation resource that should be used to create
+// custom task definitions.
+func taskDefinitionResourceType(app *scheduler.App) string {
+	check := []string{
+		"EMPIRE_X_TASK_DEFINITION_TYPE",
+		"ECS_TASK_DEFINITION", // For backwards compatibility.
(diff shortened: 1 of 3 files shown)


scheduler/cloudformation/template.go
```diff
@@ -235,6 +235,13 @@ func (t *EmpireTemplate) addTaskDefinition(tmpl *troposphere.Template, app *sche
 	cd := t.ContainerDefinition(app, p)
 	containerDefinition := cloudformationContainerDefinition(cd)
 
+	// If provided in the app environment, this role will be used when
+	// running tasks.
+	var taskRole interface{}
+	if arn, ok := app.Env["TASK_ROLE_ARN"]; ok {
```
phobologic: This seems like a 'meta' environment variable, should it maybe start with `EMPIRE_`?

phobologic:
👍 Pretty awesome, we should probably have some docs detailing this - AND the limitation around `emp run` for now.

phobologic:
Also, have you verified that this actually works? I hope it does, considering what we've heard :)

scheduler/cloudformation/template.go
ejholmes: Yeah, there's a couple of other places where we decide how to build the template based on environment variables, and it would probably be better if we prefixed those with something (other ones are `ECS_TASK_DEFINITION` and `LOAD_BALANCER_TYPE`).

I'm game for deprecating those, and switching them to `EMPIRE_*` prefixed.

phobologic: Yeah, maybe we should do that here, and then for those old ones honor the old way and the new prefixed way for a release before deprecating the old one or something.  To be fair, I think both of the 'old' cases are still pretty new, right?

ejholmes:
Ok @phobologic, I think this is good to go. Did some testing locally, and updated the example cloudformation stack to run Empire using a role assigned to the task definition.

I also added a commit that adds a prefix to the env vars that the cloudformation backend looks for to enable experimental features, which are now:

* `EMPIRE_X_LOAD_BALANCER_TYPE`
* `EMPIRE_X_TASK_DEFINITION_TYPE`
* `EMPIRE_X_TASK_ROLE_ARN`

If they ever move  to a non-experimental state, they'll either be enabled by default, or we'll remove the `X` from it.

phobologic:
👍 lets get this out there, I'm excited to use it :)

PR #985, opened 2016-08-25
Replace safeString with aws.StringValue

Nosajool:
They have the same implementation

From the comment:
https://github.com/remind101/empire/pull/983#discussion_r76170409

```go
// StringValue returns the value of the string pointer passed in or
// "" if the pointer is nil.
func StringValue(v *string) string {
	if v != nil {
		return *v
	}
	return ""
}
```


ejholmes:
:+1:

PR #987, opened 2016-09-08
ACL (spike)

ejholmes:
See https://github.com/remind101/empire/wiki/ACL-Proposal

This is a spike into an ACL system using policy documents, similar to how IAM functions. This allows you to provide a set of policies when executing an action on `empire.Empire` to determine if the action should be authorized or not.

For example, if you wanted to deny the ability for anyone to create new apps, you could do this:

``` go
ctx := acl.WithPolicy(acl.Policy{
        {
                Effect: acl.Deny,
                Action: []string{"Create"},
        },
})

empire.Create(ctx, empire.CreateOpts{
        Name: "acme-inc",
})
// Output: not authorized to perform empire:Create
```

Internally, this provides a very flexible system for performing granular access control. How this gets consumed and configured in `cmd/empire` is still a question. For the GitHub authentication backend, I'm thinking that you would attach a set of policies to the github teams that you want to provide access to (See server/auth/github/acl.go)

I think this will dovetail into a 2fa integration, so that you can easily configure what commands you want to require 2fa on. 

 Feedback much appreciated!

**TODO**
- [ ] Add the `authorize(ctx, action, resource)` pre-check to all core Empire methods.
- [ ] Implement github.TeamACL.
- [x] GitHub deployments layer
- [x] CloudFormation layer.
- [ ] Docs
- [ ] Changelog


ejholmes:
One thing that's a little tricky right now, would be cases where you want to allow deployments to a specific set of apps. To do that, you might provide a policy like so:

```go
policy := Policy{
        Effect: Allow,
        Action: []string{
                "empire:Deploy",
        },
        Resource: []string{
                "acme-inc",
                "frontend",
                "backend",
        },
}
```

Now, the problem is that you can provide `empire.Deploy` with a Docker image, but an empty app name and empire will resolve it to the correct app internally. I think the desired behavior would be for empire to resolve the app name, and then determine if the policy allows the action.

server/acl/acl.go
```diff
@@ -0,0 +1,163 @@
...
+type Context struct {
+	// The Action that's being invoked.
+	Action string
+
+	// The Resource that the action is being invoked on.
+	Resource string
+}
+
+// Policy is used to define whether an action is allowed or denied.
+type Policy struct {
```
ejholmes: I'm thinking that this will be expanded with something similar to the `Condition` key in IAM policy documents, so you could do things like allow/deny requests that didn't have 2fa enabled, limit deployments to a specific Docker repo, etc.

ejholmes:
I think it would also be possible to define Empire policies within IAM, since IAM doesn't appear to do any validation on the actions defined. So, you could create IAM roles for GitHub teams, and attach the Empire policies to it, then provide those to `empire server` like `--github.team.policies=Developers:arn:aws:iam::066251891493:role/GitHubDevelopersRole`

That could get confusing though? Also may be constraining of the `Policy` type needs to be expanded.

ejholmes:
Or potentially, ship with a set of managed policies (like user, and admin) that work for most use cases, that you can map a GitHub team to:

`--github.team.policies='Developers:user,Empire Admins:admin'`

Where the policies would be defined like:

```go
user := Policy{
        Effect: Allow,
        Action: []string{
                "empire:Deploy",
                "empire:Config",
                "empire:Set",
                "empire:Rollback",
                "empire:Scale",
        },
        Resource: []string{"*"},
}

admin := Policy{
        Effect: Allow,
        Action: []string{
                "empire:*",
        },
        Resource: []string{"*"},
}
```

but I have a feeling most people will want to have more granular control over access.

server/acl/acl.go
```diff
@@ -0,0 +1,163 @@
...
+// to check if the request should be allowed.
+type Context struct {
+	// The Action that's being invoked.
+	Action string
+
+	// The Resource that the action is being invoked on.
+	Resource string
+}
+
+// Policy is used to define whether an action is allowed or denied.
```
ejholmes: Note to self. In AWS nomenclature, this is a Statement, and a collection of Statements is a Policy.

jesushernandez:
This looks very promising. I have a couple of questions after reading through the wiki and your comments:

- how are you planning to apply these policies to empire? Do they need to be applied at startup? Or they can be changed 'online'?
- do you think it'd make sense to allow providing policies on a per-app basis? (i.e. `emp create email-service --acl=@policy.json`)

Also, I also believe that providing more granular control over the API will be better. You could ship with defaults which provide 'readonly' access (maybe just allowing you to list processes, see releases, etc) and 'full' access, which includes running one-off commands, deploying etc.

Re. using IAM to create policy documents, we use terraform (v. similar to cloudformation) to provision/configure all the AWS resources required by our empire apps, so by using IAM we could get a completely reproducible environment in AWS. Otherwise we'd have to think of an alternative way of storing these policies.

ejholmes:
> how are you planning to apply these policies to empire? Do they need to be applied at startup? Or they can be changed 'online'?

My thought right now is that they would just be provided at startup.

> do you think it'd make sense to allow providing policies on a per-app basis? (i.e. emp create email-service --acl=@policy.json)

Actually, you'll be able to do that using the `Resource` attribute in the ACL, e.g.:

```go
// Only allow app1 to be restarted.
app1 := Policy{
        Effect: Allow,
        Action: []string{
                "empire:Restart",
        },
        Resource: []string{"app1"},
}

// Only allow app2 to be rolled back.
app2 := Policy{
        Effect: Allow,
        Action: []string{
                "empire:Rollback",
        },
        Resource: []string{"app2"},
}
```

> Also, I also believe that providing more granular control over the API will be better. You could ship with defaults which provide 'readonly' access (maybe just allowing you to list processes, see releases, etc) and 'full' access, which includes running one-off commands, deploying etc.

Yeah, I agree. There may be some value in shipping with some default roles that can be used easily (we'll need to define what the most valuable roles would be), but I think most people that use Empire seriously will want to have granular control.

> Re. using IAM to create policy documents, we use terraform (v. similar to cloudformation) to provision/configure all the AWS resources required by our empire apps, so by using IAM we could get a completely reproducible environment in AWS. Otherwise we'd have to think of an alternative way of storing these policies.

Right. Just storing the policies in IAM, and then having Empire translate into it's internal Policy structure seems like it'd work well from my initial poking around with custom actions/resources in IAM policies. I think that might be a good place to start, and if it ends up not working well, we can look into alternative ways of providing policies before the next release.

server/acl/acl.go
```diff
@@ -0,0 +1,163 @@
...
+
+	if actionParts[0] != matcherParts[0] {
+		return false
+	}
+
+	return stringMatch(actionParts[1], matcherParts[1])
+}
+
+func matchResource(resource string, matcher string) bool {
+	return stringMatch(resource, matcher)
```
phobologic: What about using a regexp here?  Gives you a bit more flexibility.

phobologic:
👍 I like where this is heading.  Storing the policies in IAM would be great - does IAM actually allow free form documents to be stored? I could swear I've had the editor complain about an incorrect action, but I may be confusing it.

ejholmes:
> does IAM actually allow free form documents to be stored? I could swear I've had the editor complain about an incorrect action, but I may be confusing it.

From some quick testing I did:

1. Actions are only checked that they contain a `:`. The service prefix and action postfix are not validated, so I can do things like `foo:Bar`.
2. Resource seems to be checked that it's a valid ARN. The service and resource components inside the ARN isn't checked, so this works: `arn:aws:empire:::app/*`

So, I think it'd work okay. Empire would just pull the policy document, then translate it internal an internal policy.

empire.go
```diff
@@ -602,8 +642,12 @@ func (opts DeployOpts) Validate(e *Empire) error {
 
 // Deploy deploys an image and streams the output to w.
 func (e *Empire) Deploy(ctx context.Context, opts DeployOpts) (*Release, error) {
+	if err := authorizeApp(ctx, "Deploy", opts.App); err != nil {
```
ejholmes: This is currently exploitable, if you deploy without explicitly specifying an app:

If I had a policy like so:

```go
var policy = acl.Policy{
        acl.Statement{
                Effect: acl.Deny,
                Action: []string{"Deploy"},
                Resource: []string{"app-name"},
        },
}
```

Then if I knew the app that a Docker image would get resolved to, I would still be able to deploy.

I think this authorization needs to be moved post resolution.

empire.go
```diff
@@ -708,6 +760,10 @@ func (e *Empire) StreamLogs(app *App, w io.Writer, duration time.Duration) error
 
 // CertsAttach attaches an SSL certificate to the app.
 func (e *Empire) CertsAttach(ctx context.Context, app *App, cert string) error {
+	if err := authorizeApp(ctx, "CertsAttach", app); err != nil {
```
ejholmes: One thing I don't like right now, is there are inconsistently named methods: e.g. `CertsAtttach` and `DomainsCreate` vs `StreamLogs`. I think I prefer the later (`StreamLogs`, `AttachCert`, `AddDomain`, etc) as it reads a little nicer.

An argument could be made for the reverse, since it allows for `Prefix*` in ACL's (e.g. `Logs*`, `Domains*`).

I think, whatever we decide on, we should be consistent.

Thoughts? Preferences?

ejholmes:
Ok, added more authorization checks for more actions. The following actions are supported:

```
Create
Destroy
Config
Set
Restart
Run
Rollback
Deploy
Scale
StreamLogs
CertsAttach
```

The following still need to be added, but I want to work out naming before:

```
DomainsCreate
DomainsDestroy
```

Next up will be the IAM translator, and a way to attach policies to github teams.

empire.go
ejholmes: An easier interim solution might be to to allow a nil resource. If there's an explicit deny statement, and if the provided resource is nil, then the action is not allow, regardless of the resource constraint. e.g.:

```go
var policy = acl.Policy{
        acl.Statement{
                Effect: acl.Deny,
                Action: []string{"Deploy"},
                Resource: []string{"app-name"},
        },
}

allowed := policy.Allow(acl.Context{
        Action: "Deploy",
        Resource: nil, // unknown
})
fmt.Println(allowed)
// Output: false
```

I think a more correct, and long term solution, would be to refactor so that authorizations are checked on an internal api (could just be lowercase versions of methods to keep things simple).

ejholmes: This would also currently allow you to create a new app by using `emp deploy` even if you didn't have the Create perm.

empire.go
mhahn: I have the same preference as you re: `AddDomain` etc. but the ACL argument is valid.

I would vote for easier to read actions like `AddDomain` vs: `DomainsCreate` and be explicit when defining ACL permissions.

mhahn:
👍 on the storing policies in IAM, that sounds like a great start

empire.go
phobologic: Yeah, I'm fine with either - I think our permission list will be sufficiently short that we don't need wildcards necessarily.

PR #989, opened 2016-09-10
Action Confirmations (Duo)

ejholmes:
This is a POC for the [Action Confirmations proposal](https://github.com/remind101/empire/wiki/Action-Confirmations-Proposal)

This will send the user a push notification using Duo, and block until the action is confirmed. How we'll configure what actions require this is still TBD, but might make sense to include in policies in https://github.com/remind101/empire/pull/987.

**TODO**
- [ ] Better support in GitHub Deployments.

### DIFF
--- a/action_confimations.go
+++ b/action_confimations.go
@@ -13,6 +13,8 @@ import (
 // ActionConfirmer is an interface that can be implemented to confirm that an
 // action is allowed.
 type ActionConfirmer interface {
+	fmt.Stringer
+
 	// Confirm should notify the third party of the action being performed,
 	// then block until the action has been confirmed.
 	Confirm(ctx context.Context, user *User, action Action, params map[string]string) (bool, error)
@@ -22,6 +24,10 @@ type ActionConfirmerFunc func(context.Context, *User, Action, map[string]string)
 
 func (f ActionConfirmerFunc) Confirm(ctx context.Context, user *User, action Action, params map[string]string) (bool, error) {
 	return f(ctx, user, action, params)
+}
+
+func (f ActionConfirmerFunc) String() string {
+	return "ActionConfirmerFunc"
 }
 
 // DuoActionConfirmer is an ActionConfirmer that will send the user a Duo push
@@ -49,12 +55,17 @@ func (c *DuoActionConfirmer) Confirm(ctx context.Context, user *User, action Act
 		return false, err
 	}
 
+	pushinfo := url.Values{}
+	for k, v := range params {
+		pushinfo.Set(k, v)
+	}
+
 	q := url.Values{}
 	q.Add("username", username)
 	q.Add("factor", "push")
 	q.Add("device", "auto")
 	q.Add("type", action.String())
-	//q.Add("pushinfo", )
+	q.Add("pushinfo", pushinfo.Encode())
 
 	resp, err := c.duo.Auth(q)
 	if err != nil {
@@ -64,12 +75,16 @@ func (c *DuoActionConfirmer) Confirm(ctx context.Context, user *User, action Act
 	return resp.Response.Result == "allow", nil
 }
 
-var defaultUsernameTemplate = template.Must(template.New("username").Parse(`{{.Name}}`))
+func (c *DuoActionConfirmer) String() string {
+	return "Duo"
+}
+
+var defaultDuoUsernameTemplate = template.Must(template.New("username").Parse(`{{.Name}}`))
 
 func (c *DuoActionConfirmer) username(user *User) (string, error) {
 	t := c.UsernameTemplate
 	if t == nil {
-		t = defaultUsernameTemplate
+		t = defaultDuoUsernameTemplate
 	}
 
 	b := new(bytes.Buffer)
--- a/action_confimations.go
+++ b/action_confimations.go
@@ -1,9 +1,11 @@
 package empire
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"net/url"
+	"text/template"
 
 	"github.com/remind101/empire/pkg/duo"
 )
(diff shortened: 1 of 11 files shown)


action_confimations.go
```diff
@@ -0,0 +1,48 @@
...
+	return &DuoActionConfirmer{duo: c}
+}
+
+func (c *DuoActionConfirmer) Confirm(ctx context.Context, user *User, action string, resource string, params map[string]string) (bool, error) {
+	q := url.Values{}
+	q.Add("username", user.Name)
+	q.Add("factor", "push")
+	q.Add("device", "auto")
+	q.Add("type", action)
+	q.Add("pushinfo", fmt.Sprintf("resource=%s", resource))
```
ejholmes: Anything we include here will show up on the push info screen in duo, so we could include things like the command that's being ran.

action_confimations.go
```diff
@@ -0,0 +1,48 @@
...
+	c.Key = key
+	c.Secret = secret
+	c.Host = apiHost
+
+	return &DuoActionConfirmer{duo: c}
+}
+
+func (c *DuoActionConfirmer) Confirm(ctx context.Context, user *User, action string, resource string, params map[string]string) (bool, error) {
+	q := url.Values{}
+	q.Add("username", user.Name)
```
ejholmes: Still TBD. Right now, this is making the assumption that users in your Duo account match the username in Empire, which is unlikely to be true. This will most likely need to be configurable on the User object, so you can link an Empire user to a Duo user.

empire.go
```diff
@@ -453,8 +457,56 @@ func (opts RunOpts) Validate(e *Empire) error {
...
+	case <-ctx.Done():
+		return errors.New("timed out waiting to confirm action")
+	}
+
+	return nil
+}
+
 // Run runs a one-off process for a given App and command.
 func (e *Empire) Run(ctx context.Context, opts RunOpts) error {
+	if err := e.confirm(ctx, opts.User, "Run", opts.App.Name, nil); err != nil {
```
ejholmes: This ends up looking pretty [similar](https://github.com/remind101/empire/blob/e461e9bfd577acba67fb220bb1d6b77a3868da1a/empire.go#L197-L199) to how we check if an action is authorized in the ACL PR. May make sense to combine those, so that you can configure what actions require a confirmation via security policies.

empire.go
```diff
@@ -453,8 +457,56 @@ func (opts RunOpts) Validate(e *Empire) error {
 	return e.requireMessages(opts.Message)
 }
 
+func (e *Empire) confirm(ctx context.Context, user *User, action string, resource string, params map[string]string) error {
+	if e.ActionConfirmer == nil {
+		return nil
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, time.Second*30)
```
ejholmes: Should be configurable.

empire.go
```diff
@@ -453,8 +457,56 @@ func (opts RunOpts) Validate(e *Empire) error {
...
+		)
+
+		ok, err = e.ActionConfirmer.Confirm(ctx, user, action, resource, params)
+		if err != nil {
+			errCh <- err
+			return
+		}
+
+		if !ok {
+			err = errors.New("not authorized")
```
ejholmes: Should be more descriptive.

empire.go
```diff
@@ -453,8 +457,56 @@ func (opts RunOpts) Validate(e *Empire) error {
 	return e.requireMessages(opts.Message)
 }
 
+func (e *Empire) confirm(ctx context.Context, user *User, action string, resource string, params map[string]string) error {
```
ejholmes: This will fail closed if the third party is down (Duo), which I think is the right thing to do. If we failed open, an attacker could target the third party to gain access.

ejholmes:
Something that we discussed internally:

May be nice to just start with a webhook implementation of this, and build a separate project for performing the confirmation. Couple reasons:

1. Makes it easier for people to build custom confirmations. Just build something that responds to the webhook appropriately.
2. Doesn't couple Empire to Duo. Otherwise, Empire would most likely have to be able to link Duo users to Empire users. Easier if that kind of mapping is done separately.

I think a separate project for doing these types of action confirmations would be pretty great. Similar to what Slack does [out of band notifications](https://slack.engineering/distributed-security-alerting-c89414c992d6), but better if it's just done before the action is actually executed.

ejholmes:
Ok, here's the problem I have right now: there's currently no easy way to map an Empire user, to a Duo user.

This is really just a general problem. Ideally, we might have some Identity/User directory that we can consult to say, "Hey, here's this user's github user id, who is this person in the company" or "Given this user, what's their Duo/Slack/GitHub/etc account?"

Users + their account identities could be stored in postgres, but then we'd need to build some UI to manage that. Obviously, that could be done, but I think it'd be overkill for something like Empire. Maybe you could provide Empire with a `users.yaml` file like:

```yaml
"eric.holmes@example.org":
  name: "Eric Holmes"
  github:
    id: 1234
  duo:
    username: foo@example.org
```

Where the key is really the identity of the user inside the company. So if I log in with my GitHub account, then it just maps my GitHub user id to my identity.

This then brings up another question. Currently, when using the GitHub authentication backend, we obtain an authorization from github, then encode the GitHub API token inside a JWT encoded Empire API token. In subsequent requests, we use that token to check their org/team membership. Ideally, we'd consult the Identity/User directory to get their API token, so we can check membership within other control layers (like GH deployments).

Bit of a rabbit hole...

ejholmes:
I think probably a good alternative to re-writing all of the user management may be to just use emails as the "canonical" user id. For us, that would work fine with Duo since the username always matches the employees email. When authing with GH, we can pull their `@example.org` email and make that the user id. Convention over configuration.

PR #990, opened 2016-09-13
support newlines in parsing env files

### LINKED ISSUE #988: emp env-load does not support multiline values
Sometimes an environment value has a new line. When representing this in an env file the CLI is unable to properly parse a multiline expression where: 

```
AVAR='stuf
and
morestuf'
``` 

I came across this issue when attempting to move off heroku:
```
$ heroku config --app montage-dev -s > montage-dev.env
$ emp env-load -a montage montage-dev.env
error: poorly formatted environment: variable '-----END CERTIFICATE-----'' has white spaces
```

zbyte64:
This is to fix #988 ; Uses/vendors `github.com/joho/godotenv` to parse env files. Added a test for loading an env file.

### DIFF
--- a/tests/cli/env_test.go
+++ b/tests/cli/env_test.go
@@ -3,6 +3,9 @@ package cli_test
 import (
 	"testing"
 	"time"
+	"io/ioutil"
+	"os"
+	"fmt"
 )
 
 func TestConfig(t *testing.T) {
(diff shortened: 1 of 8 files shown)


tests/cli/env_test.go
```diff
@@ -113,3 +116,21 @@ func TestConfigConsistency(t *testing.T) {
...
+	file.Write(content)
+
+	run(t, []Command{
+		DeployCommand("latest", "v1"),
+		{
+			fmt.Sprintf("env-load %s -a acme-inc", file.Name()),
+			fmt.Sprintf("Updated env vars from %s and restarted acme-inc.", file.Name()),
+		},
+	})
+	defer os.Remove(file.Name())
```
ejholmes: Can we move this defer a little higher (before the `file.Write(content)`)? Not sure this will get run if the above test fails.

ejholmes:
Minor comment, but otherwise looks great!

tests/cli/env_test.go
ejholmes: Pulled this change in as https://github.com/remind101/empire/commit/b55d8b4c4cfcfb68656c347c9891293990cecc7d.

ejholmes:
Thanks @zbyte64!

PR #997, opened 2016-09-25
Update documentation for cert-attach

jesushernandez:
Since the schedulers (cloudformation and the old one) rely on the `SSLCertificateId` property, which accepts both a IAM server certificate ARN and a ACM certificate ARN, `emp cert-attach` is valid for both types of certificates.

I have tested this and it works (attaching an ACM cert to the external ELB).

As a side question: at the moment you cannot attach a certificate once an application has been already created and deployed. Is this a feature? If it is, should we mention it in the docs?

Thanks.


ejholmes:
Looks great! Thanks @jesushernandez!

Issue #994, opened 2016-09-15
`emp deploy` doesn't prompt for a commit message if required


ejholmes:
I think this happens because we return the error, and don't write an error to the jsonmessage stream: https://github.com/remind101/empire/blob/b55d8b4c4cfcfb68656c347c9891293990cecc7d/empire.go#L606

The result is that `emp deploy` just looks like it silently fails.

PR #1000, opened 2016-10-04
Prompt for a commit message if one isn't provided, but required, during deploy

mhahn:
Fixes: https://github.com/remind101/empire/issues/994


tests/cli/cli_test.go
```diff
@@ -89,7 +92,7 @@ func run(t testing.TB, commands []Command) {
 
 		b, err := cli.Command(args...).CombinedOutput()
 		t.Log(fmt.Sprintf("\n$ %s\n%s", cmd.Command, string(b)))
-		if err != nil {
+		if !shouldError && err != nil {
```
ejholmes: What if we take advantage of `Command.Output` being an `interface{}`. Example:

```go
run(t, []Command{
        {
                "deploy remind101/acme-inc",
                errors.New("error: A message is required for this action, please run again with '-m'."),
        },
})
```

Then, we can use that to test whether a test should expect an error:

```go
if expectedErr, ok := cmd.Output; ok {
        if err != expectedErr {
                t.Fatalf("Expected %v, got %v", err, expectedErr)
        }
} else {
        if err != nil {
                t.Fatal(err)
        }
}
```

ejholmes:
^ I should have done those as a review. Woops.

tests/cli/cli_test.go
mhahn: I like this way better, I think the CLI command is exiting with "exit status 1" error vs that error message.

```
--- FAIL: TestDeploy_CommitMessageRequired (0.11s)
	cli_test.go:93:
		$ deploy remind101/acme-inc
		error: A message is required for this action, please run again with '-m'.

	cli_test.go:96: Expected exit status 1, got error: A message is required for this action, please run again with '-m'.
```

It worked before because I was just comparing the output

mhahn: I updated it to check against the error string

PR #1011, opened 2016-10-14
Decouple access tokens from core

ejholmes:
I think this will help a bit with some of the issues brought up in https://github.com/remind101/empire/pull/989#issuecomment-253697480 (mainly just paving the way to add a users/identities service, and removing the `GitHubToken` attribute from `empire.User`).

This moves the JWT access tokens from Empire core, to `server/heroku`. The fact that these ever ended up in core was a mistake, since Empire itself never needs to deal with authentication, only the control layers do, and access tokens are control layer specific.

A nice side effect of this is that, all of the authentication logic for access tokens is now contained within `server/heroku/authentication` instead of being spread out across 3 packages.


server/heroku/authentication.go
```diff
@@ -1,63 +1,72 @@
...
-func (h *Authentication) ServeHTTPContext(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+func (s *Server) Authenticate(ctx context.Context, r *http.Request) (context.Context, error) {
 	username, password, ok := r.BasicAuth()
 	if !ok {
-		return ErrUnauthorized
+		return nil, ErrUnauthorized
 	}
 
-	user, err := h.authenticator.Authenticate(username, password, r.Header.Get(HeaderTwoFactor))
+	fmt.Println(username, password, ok)
```
mhahn: i'd remove this ;)

ejholmes: Who doesn't want passwords in logs?

PR #1015, opened 2016-10-15
Canonical user id's (company email)

ejholmes:
https://github.com/remind101/empire/pull/989 is difficult to implement right now, because there's not an easy way to map an Empire user to a Duo account.

With this change, you can make the GitHub authentication backend use emails as canonical user id's, by setting an email domain. For example, if we set the email domain to `acme-inc.com`, then it will set the internal Empire user ID as `<employee>@acme-inc.com` assuming that GitHub account has a verified `@acme-inc.com` email, which we can then use with the Duo integration to map that directly to a Duo user account. Since an employee's email is their canonical user identity across most systems, this might work well, and easier than a full blown user management system. Thoughts?

### DIFF
--- a/cmd/empire/main.go
+++ b/cmd/empire/main.go
@@ -27,6 +27,7 @@ const (
 	FlagGithubOrg          = "github.organization"
 	FlagGithubApiURL       = "github.api.url"
 	FlagGithubTeam         = "github.team.id"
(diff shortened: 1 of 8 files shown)


ejholmes:
Gonna close this. The SAML backend is a better way to do this.

Issue #1016, opened 2016-10-19
SAML/OneLogin authentication backend


ejholmes:
In addition to the GitHub auth backend, a generic SAML and/or OneLogin backend should be supported. A lot of companies (including Remind) have an existing IdP, and Empire should be able to integrate with those.

PR #1017, opened 2016-10-20
SAML authentication backend

ejholmes:
This adds support for SAML authentication. Empire is Enterprise ready.

I highly recommend that you go through this commit by commit (and probably ignore the first commit):

1. First commit just adds the [crewjam/go-xmlsec](https://github.com/crewjam/go-xmlsec) dependency, which is just a go wrapper around libxmlsec for doing xmldsig verifications.
2. Second commit adds `pkg/saml`, which provides a library for implementing a SAML provider. It's essentially a fork of [crewjam/saml](https://github.com/crewjam/saml) with the code for implementing an IdP removed.
3. Third commit adds the /saml/acs endpoint for handing SAML assertions from the IdP.

Consider this mostly a POC at this point. There's still a lot of polish needed, but wanted to open this up now to get some eyes on it. I'd love to have someone that's had more experience implementing a SAML service provider to take a look and see if there's anything obviously wrong with the implementation.

**Caveats**

1. Because SAML is inherently browser based, this currently does not work with `emp login`. We can follow up with a OneLogin extension that uses the [Generate SAML Assertion endpoint](https://developers.onelogin.com/api-docs/1/saml-assertions/generate-saml-assertion) to support `emp login`. To login, you start with an IdP initiated login (e.g. clicking the app button in OneLogin) and you get presented with some copy/paste instructions to setup ~/.netrc with an API token.
  
2. Because SAML is insane, this currently introduces a dependency on libxml2 and libxmlsec1. This means you have to `brew install` some dependencies for tests to pass. I don't see any realistic way around this with Go.
3. Currently, SLO (single logout) isn't supported, nor any authorization checks post token generation. It should probably respect some TTL in the SAML assertion and expire tokens after that amount of time.

Overall, I'm pretty excited about this, since it opens up Empire to a lot of existing authentication providers, makes it easier to control access to Empire, solves the canonical employee problem, and it'll make implementing the Duo integration very simple for us.

Closes https://github.com/remind101/empire/issues/1016

**TODO**

* [x] Update dev instructions with `brew install` stuff.
* [x] Docs on configuring SAML.
* [x] Integration tests exercising SAML.
* [x] Expire tokens.
* [x] Ensure that fake auth backend isn't added to chain.
* [x] Show a better error message when request isn't authenticated (`emp login` doesn't work).
* [x] Sign relay state tokens.


cmd/empire/main.go
```diff
@@ -98,6 +106,12 @@ var Commands = []cli.Command{
 				EnvVar: "EMPIRE_SCHEDULER",
 			},
 			cli.StringFlag{
+				Name:   FlagSAMLMetadata,
+				Value:  "",
+				Usage:  "The location of the SAML metadata XML (e.g. https://app.onelogin.com/saml/metadata/1234)",
+				EnvVar: "EMPIRE_SAML_METADATA",
```
ejholmes: Right. This can take the raw content, a url, or a file path to the content. I'll update the flag usage to reflect that.

ejholmes:
Added a commit to this to address some of the todo's in the PR description:

1. `emp login` is disabled if the SAML authentication backend is used.
2. Access tokens created via SAML authentication expire before `SessionNotOnOrAfter` (for us, this is 24 hours). It's kinda lame that you would have to re-authenticate every 24 hours, but I don't see a way around this with SAML. If someone gets removed from the IdP, then we'd want to make sure that access tokens expire soon after. What do other SAML service providers do to handle long lived tokens? AWS only allows up to 12 hours, using STS.
3. `emp login` requests the UsernamePassword authentication strategy, so that you can't use it to continuously refresh tokens.

ejholmes:
Added another commit to improve the UX around logging in. If SAML authentication is enabled, then the error returned includes a link to start an SP initiated login:

```console
$ emp apps
error: Request not authenticated, API token is missing, invalid or expired. Login at https://example.org/saml/login
```

And the instructions include `emp logout` to ensure that the old invalid token gets removed from ~/.netrc.

I still think following up with a onelogin extension that uses the Generate SAML assertion API would be good, but running into various issues with their API and their support is slow...

I think this just needs some integration tests exercising SAML login, and docs on how to configure the SAML backend, then it should be good to go. I'll probably put this up in staging under a separate domain so we can test it.

ejholmes:
Added a commit with docs on configuring the SAML authentication backend.

ejholmes:
I just re-wrote the commits, so the first 3 are only changes to vendor/ and pkg/ and the rest are the changes to Empire to support SAML.

This diff might be the easiest way to review: https://github.com/remind101/empire/compare/3871b12...saml

ejholmes:
And, just added one more commit that integration tests a full SAML service provider initiated login flow.

taylorann:
While this may seem like a good idea in theory, in practice I can't carry my phone around with me everywhere because lady pockets suck. Please refactor. 💁🏻

cmd/empire/main.go
```diff
@@ -98,6 +110,30 @@ var Commands = []cli.Command{
 				EnvVar: "EMPIRE_SCHEDULER",
 			},
 			cli.StringFlag{
+				Name:   FlagServerAuth,
+				Value:  "",
+				Usage:  "The authentication backend to use to authenticate requests to the API. Can be `fake`, `github`, or `saml`.",
```
ejholmes: Previously, we've defaulted to `fake`, which was determined by the absence of `EMPIRE_GITHUB_CLIENT_ID`.

For backwards compatibility, I went with leaving this blank by default. If it's blank, then it'll use the existing behavior and look at `EMPIRE_GITHUB_CLIENT_ID`. We could break backwards compatibility here and error out if it's not explicitly provided, which I'd be ok with.

cmd/empire/main.go
```diff
@@ -82,6 +88,12 @@ var Commands = []cli.Command{
 		Usage:     "Run the empire HTTP api",
 		Flags: append([]cli.Flag{
 			cli.StringFlag{
+				Name:   FlagURL,
+				Value:  "",
+				Usage:  "That base URL where this Empire instance runs",
```
ejholmes: It's only required if the `saml` authentication backend is used. I'll update this to error out if the `saml` backend is used, but this isn't provided.

ejholmes:
I'll go ahead and merge this into master, and will follow up with any bug fixes and/or UX improvements separately.

PR #1018, opened 2016-11-01
Onelogin authentication backend

ejholmes:
This compliments the SAML authentication backend added in https://github.com/remind101/empire/pull/1017, with a Onelogin extension using the [Generate SAML assertion](https://developers.onelogin.com/api-docs/1/saml-assertions/generate-saml-assertion) API of Onelogin. This allows users to use `emp login` to perform SAML authentication, instead of using the browser flow.

With that said, this currently doesn't work for me because Duo devices are apparently broken in the Onelogin API. Onelogin has verified that it is broken, but I've yet to hear back from them one what they're doing to fix it.

### DIFF
--- a/server/auth/auth.go
+++ b/server/auth/auth.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/remind101/empire"
 )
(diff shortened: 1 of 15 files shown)


server/auth/auth.go
```diff
@@ -122,35 +123,49 @@ func (a *Auth) Authenticate(ctx context.Context, username, password, otp string,
...
 	return ctx, nil
 }
 
+// Session represents an authenticated Session.
+type Session struct {
+	// The authenticated User.
+	User *empire.User
+
+	// When this Session will expire. The zero value means no expiration.
+	ExpiresAt *time.Time
```
ejholmes: This allows us to propagate the `SessionNotOnOrAfter` assertion down, and ensures that access tokens created expire with the SAML session.

ejholmes:
I checked this out again on friday. Apparently still broken on OneLogin's side. /sigh

I bugged their support again to see what they're doing about it.

ejholmes:
OneLogin support finally got a response from their engineering team on this and got back to me. They say that support for Duo devices in the API is unsupported (despite having docs referencing Duo devices). So, I'm going to close this for now.

phobologic:
Why is there no annoyed emoji for this? Grrr.

PR #1030, opened 2016-12-14
Increase stack operation timeout

ejholmes:
10 minutes is a little too aggressive, since CloudFormation gets slow sometimes. 1 hour is a little more reasonable, this just prevents Empire from getting in a stuck state and holding onto the stack operation lock indefinitely.


ejholmes:
It's a pretty low level detail. We could make it configurable, but I doubt anyone would ever configure it. We can probably address that if it comes up.

PR #1040, opened 2017-02-13
Ignore invalid TD when rolling back

phobologic:
This is to deal with an issue we ran into in empire staging that got the
stack in a bad (unable to rollback state).

Should not be merged with master.

### DIFF
--- a/server/cloudformation/ecs.go
+++ b/server/cloudformation/ecs.go
@@ -17,6 +17,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ecs"
 	"github.com/mitchellh/hashstructure"
 	"github.com/remind101/empire/pkg/cloudformation/customresources"
+	"github.com/remind101/pkg/logger"
 	"github.com/remind101/pkg/reporter"
 )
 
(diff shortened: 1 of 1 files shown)


server/cloudformation/ecs.go
```diff
@@ -120,6 +120,12 @@ func (p *ECSServiceResource) Provision(ctx context.Context, req customresources.
 			} else {
 				err = fmt.Errorf("no primary deployment found")
 			}
+		} else {
```
ejholmes: We should probably just reverse the flow here so we check the error case first, and remove the duplication in getting the deployment id:

```go
if err != nil {
        if err, ok := err.(awserr.Error); ok && strings.Contains(err.Message(), "TaskDefinition is inactive") {
                // If the task definition has been previously deleted, just do
                // nothing.
        } else {
                return id, data, err
        }
}


d := primaryDeployment(resp.Service)
if d != nil {
        data["DeploymentId"] = *d.Id
} else {
        err = fmt.Errorf("no primary deployment found")
}
```

phobologic: Ahh, good call

ejholmes: Although, I guess in this case, `d` would be nil, since the there's no PRIMARY deployment, in which case we wouldn't want to return an error.

phobologic:
Closing, this was only meant as a temporary fix for us, we need something more robust.

PR #1043, opened 2017-03-01
Use the ECS RunTask API to run attached processes

ejholmes:
Not ready to merge.

This is a prototype for switching `emp run` to use the ECS RunTask API, instead of calling out to Docker directly. The primary motivation behind this is to allow containers started with `emp run` to use AWS Roles for ECS tasks. This is described in https://github.com/remind101/empire/wiki/RFC:-Use-RunTask-API-for-attached-runs.

There's some serious caveats with this implementation at the moment:

1. Because we use `AttachToContainer`, the container needs to be started with the `Tty` and `OpenStdin` flags set to true. Unfortunately, the ECS API and the ECS agent don't currently provide a method to pass these down. We would currently need to fork the agent.
2. I haven't yet addressed backwards compatibility. It may be a good idea to provide a flag to use the old method for running attached runs entirely through Docker.

On the plus side, this sets up some infrastructure for adding an `emp exec/attach` command for starting up a process in an existing container, or attaching to a running process.

**TODO**

* [x] Figure out how to pass down `Tty`/`OpenStdin` flags.
* [x] Add the ability to limit attached runs to a group of hosts.
* [x] Enable this feature with a flag.
* [ ] <strike>Add ability to limit normal containers to a group of hosts.</strike>
* [x] Docs
  * Expiremental
  * Document Docker TLS requirements
  * Document need for ECS agent patch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## HEAD
 
+**Features**
+
+* Empire now supports a new (experimental) feature to enable attached processes to be ran with ECS. [#1043](https://github.com/remind101/empire/pull/1043)
+
 ## 0.12.0 (2017-03-10)
 
 **Features**
--- a/cmd/empire/main.go
+++ b/cmd/empire/main.go
@@ -45,9 +45,9 @@ const (
 
 	FlagDB = "db"
 
-	FlagDockerSocket = "docker.socket"
-	FlagDockerCert   = "docker.cert"
-	FlagDockerAuth   = "docker.auth"
+	FlagDockerHost = "docker.socket"
+	FlagDockerCert = "docker.cert"
+	FlagDockerAuth = "docker.auth"
 
 	FlagAWSDebug             = "aws.debug"
 	FlagS3TemplateBucket     = "s3.templatebucket"
@@ -57,6 +57,8 @@ const (
 	FlagECSServiceRole       = "ecs.service.role"
 	FlagECSLogDriver         = "ecs.logdriver"
 	FlagECSLogOpts           = "ecs.logopt"
+	FlagECSAttachedEnabled   = "ecs.attached.enabled"
+	FlagECSDockerCert        = "ecs.docker.cert"
 
 	FlagELBSGPrivate = "elb.sg.private"
 	FlagELBSGPublic  = "elb.sg.public"
@@ -249,7 +251,7 @@ var DBFlags = []cli.Flag{
 
 var EmpireFlags = []cli.Flag{
 	cli.StringFlag{
-		Name:   FlagDockerSocket,
+		Name:   FlagDockerHost,
 		Value:  "unix:///var/run/docker.sock",
 		Usage:  "The location of the docker api",
 		EnvVar: "DOCKER_HOST",
@@ -309,6 +311,17 @@ var EmpireFlags = []cli.Flag{
 		Value:  &cli.StringSlice{},
 		Usage:  "Log driver to options. Maps to the --log-opt docker cli arg",
 		EnvVar: "EMPIRE_ECS_LOG_OPT",
+	},
+	cli.BoolFlag{
+		Name:   FlagECSAttachedEnabled,
+		Usage:  "When enabled, indicates that ECS tasks can be attached to, using `docker attach`. When provided, this will also use ECS to run attached processes. At the moment, this flag should only be set if you're running a patched ECS agent. See http://empire.readthedocs.io/en/latest/configuration/ for more information.",
+		EnvVar: "EMPIRE_ECS_ATTACHED_ENABLED",
+	},
+	cli.StringFlag{
+		Name:   FlagECSDockerCert,
+		Value:  "",
+		Usage:  "A path to the certificates to use when connecting to Docker daemon's on container instances.",
+		EnvVar: "EMPIRE_ECS_DOCKER_CERT_PATH",
 	},
 	cli.StringFlag{
 		Name:   FlagELBSGPrivate,
--- a/scheduler/cloudformation/clients.go
+++ b/scheduler/cloudformation/clients.go
@@ -3,6 +3,7 @@ package cloudformation
 import (
 	"time"
 
+	awswaiter "github.com/aws/aws-sdk-go/private/waiter"
 	"github.com/aws/aws-sdk-go/service/ecs"
 	"github.com/pmylund/go-cache"
 	"github.com/remind101/empire/pkg/arn"
@@ -30,7 +31,7 @@ type cachingECSClient struct {
 }
 
 // ecsWithCaching wraps an ecs.ECS client with caching.
-func ecsWithCaching(ecs *ecs.ECS) *cachingECSClient {
+func ecsWithCaching(ecs *ECS) *cachingECSClient {
 	return &cachingECSClient{
 		ecsClient:       ecs,
 		taskDefinitions: cache.New(defaultExpiration, defaultPurge),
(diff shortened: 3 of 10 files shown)


ejholmes:
To expand a bit on the idea of adding a method to provide the `Tty` and `OpenStdin` flags to `CreateContainer` in the ECS agent, now that Go 1.8 is released and has support for plugins, we could add a feature in the ECS agent for "task plugins", to manipulate the task before starting. I'll open an issue on the amazon-ecs-agent repo to see if there would be any interest in supporting something like that.

phobologic:
Oh that would be awesome if they were down for it.

cmd/empire/main.go
```diff
@@ -57,6 +57,8 @@ const (
 	FlagECSServiceRole       = "ecs.service.role"
 	FlagECSLogDriver         = "ecs.logdriver"
 	FlagECSLogOpts           = "ecs.logopt"
+	FlagECSAttachedEnabled   = "ecs.attached.enabled"
+	FlagECSAttachedPlacement = "ecs.attached.placement"
```
ejholmes: This should probably also allow you to set placement constraints for normal procs and detached procs. If not (and assuming you want to restrict attached processes to a group of hosts), normal tasks could get scheduled onto the hosts that you've designated for attached runs.

scheduler/cloudformation/cloudformation.go
```diff
@@ -911,26 +944,144 @@ func (m *Scheduler) Run(ctx context.Context, app *scheduler.App, process *schedu
...
+	})
+	if err != nil {
+		return fmt.Errorf("error describing ec2 instance (%s): %v", aws.StringValue(containerInstance.Ec2InstanceId), err)
+	}
+
+	ec2Instance := descInstanceResp.Reservations[0].Instances[0]
+
+	// Wait for the task to start running. It will stay in the
+	// PENDING state while the container is being pulled.
+	if err := m.ecs.WaitUntilTasksRunning(&ecs.DescribeTasksInput{
```
ejholmes: This should actually wait until it's RUNNING or STOPPED. If you run something that returns quickly (e.g. `emp run echo hello world`), the task could hit `STOPPED` before this method had a change to actually see it was running. It should be fine to `AttachToContainer` on a stopped task, since the ECS agent waits 3 hours to cleanup containers.

ejholmes:
Ok, @phobologic. If you wanna give this another look, I think this is probably good to go. We've been running this on staging, without any hitches so far.

Primary changes since the last review:

1. Added docs/caveats about the new method, and requirements for patching the ECS agent.
2. Fixed some bugs with attaching to tasks that enter RUNNING for a brief period of time.
3. Stripped out support for placement constraints. This requires a more thoughtful implementation, and isn't necessary for most use cases. I'll follow up on this later.
4. Added an `Attaching to task/<task id>...` message, since we no longer see the Docker pull output (yay!).

The first commit can just be ignored, since it's just an update to the AWS SDK.

ejholmes:
Noticed some zombie procs lying around. This should probably try a `StopTask` after detaching.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * An ALIAS record is now created for `<process>.<app>.<zone>` [#1005](https://github.com/remind101/empire/pull/1005)
 * You can now provide a `-p` flag to the `emp cert-attach` command to attach a certificate to a specific process (instead of just `web`). [#1014](https://github.com/remind101/empire/pull/1014)
 * Empire now supports a SAML authentication backend. [#1017](https://github.com/remind101/empire/pull/1017)
+* Empire supports a new (experimental) feature to enabled attached processes to be ran with ECS. [#1043](https://github.com/remind101/empire/pull/1043)
```
phobologic: `to enabled` -> `to enable`

cmd/empire/main.go
```diff
@@ -310,6 +311,11 @@ var EmpireFlags = []cli.Flag{
 		Usage:  "Log driver to options. Maps to the --log-opt docker cli arg",
 		EnvVar: "EMPIRE_ECS_LOG_OPT",
 	},
+	cli.BoolFlag{
+		Name:   FlagECSAttachedEnabled,
+		Usage:  "When enabled, indicates that ECS tasks can be attached to, using `docker attach`. When provided, this will also use ECS to run attached processes.",
```
phobologic: With this requiring the patched agent, it might be worth pointing at the doc that details what this means.

scheduler/cloudformation/clients.go
```diff
@@ -59,3 +60,43 @@ func (c *cachingECSClient) DescribeTaskDefinition(input *ecs.DescribeTaskDefinit
 
 	return resp, err
 }
+
+// ECS augments the ecs.ECS client with extra waiters.
+type ECS struct {
+	*ecs.ECS
+}
+
+func (c *ECS) WaitUntilTasksNotPending(input *ecs.DescribeTasksInput) error {
```
phobologic: my linter is going to complain about no docstring here :)

scheduler/cloudformation/cloudformation.go
```diff
@@ -911,17 +935,30 @@ func (m *Scheduler) Run(ctx context.Context, app *scheduler.App, process *schedu
...
+	containerDefinition := t.ContainerDefinition(app, process)
+	if attached {
+		if containerDefinition.DockerLabels == nil {
+			containerDefinition.DockerLabels = make(map[string]*string)
+		}
+		// NOTE: Currently, this depends on a patched version of the
+		// Amazon ECS Container Agent, since the official agent doesn't
+		// provide a method to pass these down to the `CreateContainer`
+		// call.
+		containerDefinition.DockerLabels["docker.config.Tty"] = aws.String("true")
```
phobologic: Curious why both?

scheduler/cloudformation/cloudformation.go
```diff
@@ -931,6 +968,101 @@ func (m *Scheduler) Run(ctx context.Context, app *scheduler.App, process *schedu
...
+	})
+	if err != nil {
+		return fmt.Errorf("error describing ec2 instance (%s): %v", aws.StringValue(containerInstance.Ec2InstanceId), err)
+	}
+
+	ec2Instance := descInstanceResp.Reservations[0].Instances[0]
+
+	// Wait for the task to start running. It will stay in the
+	// PENDING state while the container is being pulled.
+	if err := m.ecs.WaitUntilTasksNotPending(&ecs.DescribeTasksInput{
```
phobologic: Wondering if we need some sort of timeout here?

scheduler/cloudformation/cloudformation.go
```diff
@@ -931,6 +968,101 @@ func (m *Scheduler) Run(ctx context.Context, app *scheduler.App, process *schedu
...
+		Tasks:   []*string{task.TaskArn},
+	}); err != nil {
+		return fmt.Errorf("error waiting for %s to transition from PENDING state: %s", aws.StringValue(task.TaskArn), err)
+	}
+
+	// Open a new connection to the Docker daemon on the EC2
+	// instance where the task is running.
+	d, err := m.NewDockerClient(ec2Instance)
+	if err != nil {
+		return fmt.Errorf("error establishing docker client connection to %s", aws.StringValue(ec2Instance.InstanceId))
```
phobologic: Will this bubble up with the task id, etc?  Might want to put some more info in this so it's easier to figure out what happened.

scheduler/cloudformation/cloudformation.go
ejholmes: The `Tty` flag allocates a psuedo TTY for the container, and `OpenStdin` opens the stdin stream.

Without `OpenStdin`, running something like `emp run bash` would just exit as soon as bash hit the prompt. Without `Tty`, I think most commands that expect a TTY/terminal wouldn't work (If I try `docker run -i ubuntu bash` (missing `-t` flag), my terminal just hangs).

CHANGELOG.md
ejholmes: Good catch. I'll update.

cmd/empire/main.go
ejholmes: Good call. I'll update.

scheduler/cloudformation/cloudformation.go
ejholmes: This will timeout after 10 minutes (6 second delay between refreshes, and 100 retries).

scheduler/cloudformation/cloudformation.go
ejholmes: Not atm. I'll add that information to the error.

ejholmes: Actually, it does already show this, since it's output above the error (`Attaching to task/<task id>...`).

ejholmes: Oh, but this is totally missing the `err`.

scheduler/cloudformation/cloudformation.go
phobologic: 👍

scheduler/cloudformation/cloudformation.go
phobologic: 👍

ejholmes:
This has been working out pretty well for us in our staging environment without any major issues/bugs. Gonna go ahead and merge.

PR #1048, opened 2017-03-02
Stop double confirmation messages

### LINKED ISSUE #1047: `emp destroy` does the 'type app name' check twice
I don't know if this is intentional, but when you destroy an app with `emp destroy`, you have to type the app name twice (once b4 the commit message, once after the commit message). 

example with fake app foo: 
```
Calebs-MacBook-Pro:r101-api caleb$ emp-prod destroy foo
warning: This will destroy foo and its add-ons. Please type "foo" to continue:
> foo
A commit message is required, enter one below:
> destroying foo app
warning: This will destroy foo and its add-ons. Please type "foo" to continue:
> foo
Destroyed foo.
```

DavidLu1997:
Fixes #1047 

Changes:
- Added a function similar to `maybeMessage` that only checks for confirmation once
- Removed confirmation checking from the destroy command

### DIFF
--- a/cmd/emp/destroy.go
+++ b/cmd/emp/destroy.go
@@ -7,7 +7,7 @@ import (
 )
 
 var cmdDestroy = &Command{
-	Run:             maybeMessage(runDestroy),
+	Run:             mustConfirmAndMessageRequired(runDestroy, warningMessage),
 	Usage:           "destroy <name>",
 	OptionalMessage: true,
 	Category:        "app",
@@ -34,9 +34,7 @@ func runDestroy(cmd *Command, args []string) {
 	appname := args[0]
 	message := getMessage()
 
-	warning := fmt.Sprintf("This will destroy %s and its add-ons. Please type %q to continue:", appname, appname)
-	mustConfirm(warning, appname)
-
+	appname := args[0]
 	must(client.AppDelete(appname, message))
 	log.Printf("Destroyed %s.", appname)
 	remotes, _ := gitRemotes()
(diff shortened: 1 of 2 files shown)


cmd/emp/destroy.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 var cmdDestroy = &Command{
-	Run:             maybeMessage(runDestroy),
+	Run:             mustConfirmAndMessageRequired(runDestroy, warningMessage),
```
ejholmes: I think we can probably just compose the existing `maybeMessage` function, and prefix this chain with the confirmation. So, something like this:

```go
confirmDestroy(maybeMessage(runDestroy))
```

And then the confirmDestroy func

```go
func confirmDestroy(action func(cmd *Command, args []string)) func(cmd *Command, args []string) {
        cmd.CheckNumArgs(args)
	appname := args[0]
	warning := fmt.Sprintf("This will destroy %s and its add-ons. Please type %q to continue:", appname, appname)
	mustConfirm(warning, appname)
	action(cmd, args)
}
```

Also, note, I added the `cmd.CheckNumArgs` check, since `emp destroy` would panic of the app name wasn't provided.

cmd/emp/destroy.go
```diff
@@ -49,3 +46,10 @@ func runDestroy(cmd *Command, args []string) {
 		}
 	}
 }
+
+func warningMessage(args []string) (warning, desired string) {
+	appname := args[0]
```
ejholmes: I addressed this in the other comment, but this would panic if you don't provide an app name, since the `len(args)` check happens within runDestroy.

cmd/emp/destroy.go
```diff
@@ -29,16 +29,23 @@ Example:
 `,
 }
 
-func runDestroy(cmd *Command, args []string) {
-	if len(args) != 1 {
-		cmd.PrintUsage()
-		os.Exit(2)
+func confirmDestroy(action func(cmd *Command, args []string)) func(cmd *Command, args []string) {
+	return func(cmd *Command, args []string) {
+		if len(args) != 1 {
```
ejholmes: Mind rebasing and changing this to `cmd.CheckNumArgs`?

DavidLu1997:
@ejholmes changes made

PR #1049, opened 2017-03-02
Refactor out argument count check

DavidLu1997:
I was reading through the codebase when I saw:
```
if len(args) != 2 {
	cmd.PrintUsage()
	os.Exit(2)
}
```
repeated a lot.

Therefore, I decided to refactor that out into an optional parameter on the `Command` struct with its own validation function.

I did not touch commands with a variable argument count.

### DIFF
--- a/cmd/emp/certs.go
+++ b/cmd/emp/certs.go
@@ -34,7 +34,7 @@ func init() {
 }
 
 func runCertAttach(cmd *Command, args []string) {
-	cmd.CheckNumArgs(args)
+	cmd.AssertNumArgsCorrect(args)
 
 	cert := args[0]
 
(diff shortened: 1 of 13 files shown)


cmd/emp/certs.go
```diff
@@ -37,10 +34,7 @@ func init() {
 }
 
 func runCertAttach(cmd *Command, args []string) {
-	if len(args) == 0 {
```
ejholmes: Looks like this one breaks previous behavior (note `==` rather than `!=` like the rest). I think `NumArgs` should just be 1, instead of 0.

cmd/emp/main.go
```diff
@@ -103,6 +104,13 @@ func (c *Command) ShortExtra() string {
 	return c.Short[:len(c.Short)-len(extra)]
 }
 
+func (c *Command) CheckNumArgs(args []string) {
+	if len(args) != c.NumArgs {
```
ejholmes: Instead of having to explicitly set `NumArgs`, we could parse this out of `Usage`, since it follows the convention `subcommand <arg1> <arg2> <etc>`.

DavidLu1997:
@ejholmes I fixed `certs.go`, should I still parse args out of usage?

PR #1059, opened 2017-03-14
ECS placement constraints/strategy

ejholmes:
This adds support to the procfile for specifying ECS [placement constraints](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-placement-strategies.html) and [placement strategies](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-placement-strategies.html).

I cherry-picked https://github.com/remind101/empire/pull/941 into this, since it makes it easier, so may be better to only look at the second commit.

### DIFF
--- a/cmd/empire/factories.go
+++ b/cmd/empire/factories.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"html/template"
 	"log"
@@ -27,9 +28,11 @@ import (
 	"github.com/remind101/empire/pkg/dockerauth"
 	"github.com/remind101/empire/pkg/dockerutil"
 	"github.com/remind101/empire/pkg/troposphere"
+	"github.com/remind101/empire/procfile"
 	"github.com/remind101/empire/scheduler/cloudformation"
 	"github.com/remind101/empire/scheduler/docker"
 	"github.com/remind101/empire/stats"
+	"github.com/remind101/empire/twelvefactor"
 	"github.com/remind101/pkg/reporter"
 	"github.com/remind101/pkg/reporter/config"
 )
@@ -136,7 +139,7 @@ func newScheduler(db *empire.DB, c *Context) (empire.Scheduler, error) {
 	return a, nil
 }
 
-func newCloudFormationScheduler(db *empire.DB, c *Context) (*cloudformation.Scheduler, error) {
+func newCloudFormationScheduler(db *empire.DB, c *Context) (twelvefactor.Scheduler, error) {
 	logDriver := c.String(FlagECSLogDriver)
 	logOpts := c.StringSlice(FlagECSLogOpts)
 	logConfiguration := newLogConfiguration(logDriver, logOpts)
@@ -209,7 +212,31 @@ func newCloudFormationScheduler(db *empire.DB, c *Context) (*cloudformation.Sche
 	log.Println(fmt.Sprintf("  ZoneID: %v", zoneID))
 	log.Println(fmt.Sprintf("  LogConfiguration: %v", t.LogConfiguration))
 
+	if v := c.String(FlagECSPlacementConstraintsDefault); v != "" {
+		var placementConstraints []*ecs.PlacementConstraint
+		if err := json.Unmarshal([]byte(v), &placementConstraints); err != nil {
+			return nil, fmt.Errorf("unable to unmarshal placement constraints: %v", err)
+		}
+		log.Println(fmt.Sprintf("  DefaultPlacementConstraints: %v", placementConstraints))
+		return twelvefactor.Transform(s, setDefaultPlacementConstraints(placementConstraints)), nil
+	}
+
 	return s, nil
+}
+
+func setDefaultPlacementConstraints(placementConstraints []*ecs.PlacementConstraint) func(*twelvefactor.Manifest) *twelvefactor.Manifest {
+	return func(m *twelvefactor.Manifest) *twelvefactor.Manifest {
+		for _, p := range m.Processes {
+			if p.ECS == nil {
+				p.ECS = &procfile.ECS{}
+			}
+
+			if p.ECS.PlacementConstraints == nil {
+				p.ECS.PlacementConstraints = placementConstraints
+			}
+		}
+		return m
+	}
 }
 
 func newLogConfiguration(logDriver string, logOpts []string) *ecs.LogConfiguration {
--- a/cmd/empire/main.go
+++ b/cmd/empire/main.go
@@ -53,16 +53,17 @@ const (
 	FlagDockerCert = "docker.cert"
 	FlagDockerAuth = "docker.auth"
 
-	FlagAWSDebug             = "aws.debug"
-	FlagS3TemplateBucket     = "s3.templatebucket"
-	FlagCustomResourcesTopic = "customresources.topic"
-	FlagCustomResourcesQueue = "customresources.queue"
-	FlagECSCluster           = "ecs.cluster"
-	FlagECSServiceRole       = "ecs.service.role"
-	FlagECSLogDriver         = "ecs.logdriver"
-	FlagECSLogOpts           = "ecs.logopt"
-	FlagECSAttachedEnabled   = "ecs.attached.enabled"
-	FlagECSDockerCert        = "ecs.docker.cert"
+	FlagAWSDebug                       = "aws.debug"
+	FlagS3TemplateBucket               = "s3.templatebucket"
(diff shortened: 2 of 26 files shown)


ejholmes:
One thing worth mentioning is that, I can see cases where you might want these to be runtime configurable per environment. For example, you might have a CPU intensive task that you want to run on a `c4.8xlarge` in production, but then only run on a `t2.micro` in staging.

I think, in these cases, the best thing to do is setup "named profiles" using container instance attributes, that map to their production/staging equivalents. For example, in the scenario above, I might add a `profile=cpu` attribute to container instances, where in production it's backed by `c4.8xlarge`, but only `t2.micro` in staging. Then I can use that profile in the Procfile:

```yaml
web:
  ecs:
    placement:
      constraints:
        - { type: memberOf, expression: "attribute:profile = cpu" }
```

server/cloudformation/ports.go
```diff
@@ -14,36 +14,34 @@ type portAllocator interface {
...
-		if err2 != nil {
-			err = fmt.Errorf("physical resource id should have been a port number: %v", err2)
-			return
-		}
-		id = req.PhysicalResourceId
-		err = p.ports.Put(int64(port))
-	default:
-		err = fmt.Errorf("%s is not supported", req.RequestType)
-	}
+func (p *InstancePortsResource) Create(_ context.Context, req customresources.Request) (string, interface{}, error) {
```
phobologic: Man - does your linter not freak out at you for not having comments describing exported methods?  It gets annoying - I'm wondering if I should just find a way to disable it, or if we should try and get better about listening to it.

ejholmes: Yeah, I disabled that a long time ago because you just end up with a bunch of `Thing does thing` doc comments to appease the linter.

ejholmes:
Good call. I'll update the procfile doc with information on that. The Procfile isn't documented in docs/ but there's a readme in [./procfile](https://github.com/remind101/empire/tree/master/procfile) where all the options are documented, which the docs point to.

ejholmes:
Another thing to think about here is that, if you're including a bunch of different instance types in a single ECS cluster, you probably want services that don't specify any placement constraints to default to some constraints, so they don't go bouncing around a bunch of different instance types.

Some kind of first class "profile" support, that maps to some ECS placement constraints might be a better way to go initially. So, when booting Empire, you'd give it a number of named "profiles", which map to ECS placement constraints, and then you specify a default profile. For example:

```json
{
  "default": {
    "placement_constraints": [
      { "type": "memberOf", "expression": "attributes:ecs.instance-type = m3.*" }
    ]
  },
  "memory": {
    "placement_constraints": [
      { "type": "memberOf", "expression": "attributes:ecs.instance-type = x1.*" }
    ]
  }
}
```

Users would be able to define what "profile" to use for a process in the Procfile:

```yaml
worker:
  profile: memory
```

Or they could change this at runtime:

```console
$ emp scale worker --profile memory -a <app> 
```

Or provide it via `emp run`:

```console
$ emp run ./bin/job --profile memory -a <app>
```

Having a simple "named profile" provides users with a lot of simplicity, but still gives operators enough power to define as many profiles as needed, and document those within their org.

Food for thought 🤔

ejholmes:
Alright, I'm feeling pretty good about this one now. I'll start doing some more testing, and giving it a whirl on our staging env.

runner.go
```diff
@@ -21,43 +26,57 @@ func (r *runnerService) Run(ctx context.Context, opts RunOpts) error {
 	}
 
 	procName := opts.Command[0]
-	proc := Process{
-		Quantity: 1,
-	}
+	var proc Process
```
phobologic: same question about this method - it's getting a little long, worth breaking it up?

scheduler/docker/docker.go
```diff
@@ -146,73 +146,75 @@ func NewScheduler(client *dockerutil.Client) *Scheduler {
...
-
-	if err := s.docker.PullImage(ctx, docker.PullImageOptions{
-		Registry:     p.Image.Registry,
-		Repository:   p.Image.Repository,
-		Tag:          p.Image.Tag,
-		OutputStream: replaceNL(out),
-	}); err != nil {
-		return fmt.Errorf("error pulling image: %v", err)
-	}
+	for _, p := range app.Processes {
```
phobologic: This seems like a pretty big change - there will only ever be a single process, right?  Is this just short circuiting needing the `p` since that's already represented as an array in `app.Processes` and it only has a single value when `Run` is called?

ejholmes:
@phobologic I just rebased this into smaller logical commits if it makes reviewing a little easier. The actual change to implement ECS placement constraints and strategies is in the last commit, which is relatively small. Commits 1-4 probably don't need too much attention, since they're mostly internal changes to make the last commit easier.

* The first commit is just https://github.com/remind101/empire/pull/941, which has already been approved/tested, so can probably be ignored.
* The second commit should probably get some attention. It simplifies the `scheduler.Run` method to only take a `Manifest` instead of both a `Manifest` and a `Process`, since the `Manifest` already contains the list of processes to run. It also fixes a "bug" so that `emp run` takes stored process configuration into account when running named procs.
* Commits 3 just cleans up some legacy issues with `scheduler.Restart`. I probably shouldn't include it in this PR.

I'll follow up on the rest of the review comments once we've had a chance to test this in our staging env.

cmd/empire/factories.go
```diff
@@ -209,9 +212,33 @@ func newCloudFormationScheduler(db *empire.DB, c *Context) (*cloudformation.Sche
 	log.Println(fmt.Sprintf("  ZoneID: %v", zoneID))
 	log.Println(fmt.Sprintf("  LogConfiguration: %v", t.LogConfiguration))
 
+	if v := c.String(FlagECSPlacementConstraintsDefault); v != "" {
+		var placementConstraints []*ecs.PlacementConstraint
+		if err := json.Unmarshal([]byte(v), &placementConstraints); err != nil {
+			return nil, fmt.Errorf("unable to unmarshal placement constraints: %v", err)
+		}
+		log.Println(fmt.Sprintf("  DefaultPlacementConstraints: %v", placementConstraints))
+		return twelvefactor.Transform(s, setDefaultPlacementConstraints(placementConstraints)), nil
```
ejholmes: Yeah, I can see how that could make reviewing difficult. I tend to just follow the language communities preference for things like this. In the case of Go: https://github.com/golang/go/wiki/CodeReviewComments#receiver-names.

In regards to breaking it up, there's not much within this that's re-usable, so breaking it up might make it hard to follow.

cmd/empire/main.go
```diff
@@ -334,6 +335,12 @@ var EmpireFlags = []cli.Flag{
 		EnvVar: "EMPIRE_ECS_DOCKER_CERT_PATH",
 	},
 	cli.StringFlag{
+		Name:   FlagECSPlacementConstraintsDefault,
+		Value:  "",
+		Usage:  "ECS placement constraints to set when a process does not set any.",
```
ejholmes: Yep. I'll expand on this before it gets merged.

scheduler/docker/docker.go
ejholmes: I kinda addressed this below, but the overall change is basically a noop, since Empire core only ever passes 1 process, so the end result is the same. It does mean that Empire core _could_ pass multiple processes to run them at once, but primarily this was done just to simplify the method signature since the manifest already contains all the process information.

cmd/empire/factories.go
phobologic: Yeah, I don't get their dislike of descriptive names.  Since when did we run out of bytes for descriptive code?! :)

scheduler/cloudformation/template.go
```diff
@@ -690,6 +690,28 @@ func (t *EmpireTemplate) addService(tmpl *troposphere.Template, app *twelvefacto
...
+	if v := p.ECS; v != nil {
+		if len(v.PlacementConstraints) > 0 {
+			var placementConstraints []interface{}
+			for _, c := range v.PlacementConstraints {
+				placementConstraints = append(placementConstraints, map[string]interface{}{
+					"Type":       c.Type,
+					"Expression": c.Expression,
+				})
+			}
+			serviceProperties["PlacementConstraints"] = placementConstraints
```
ejholmes: This should actually be added to the task definition, rather than the service. Reason being, this isn't updatable on a service, which would require a replacement of the service. If we do it on the task definition, then no replacement required. Placement strategies can only be applied to the service, so that would need to remain.

ejholmes:
This is working pretty nicely in our staging environment. I'll go ahead and pull this in and fix any issues the crop up before the next release.

Issue #1042, opened 2017-02-24
Use negative scale to remove resources


ejholmes:
Right now, resources for a scaled down process are always created (except for [scheduled procs](https://github.com/remind101/empire/blob/master/scheduler/cloudformation/template.go#L256)).

I think the default behavior should be to create the resources (for example, you wouldn't want the load balancer to be destroyed because you scaled the proc down to 0), when the process is scaled down, and use `emp scale <proc>=-1` to indicate that resources should not be created. This would be a backwards incompatible change for scheduled procs, but I think makes things more consistent.

ejholmes:
The currently behavior also enables an annoying bug if you try to deploy an application that _only_ has scheduled processes in it, which are all scaled down. CloudFormation will error out with:

```
Template format error: At least one Resources member must be defined. status code: 400
```

Since there are no resources in the template.

PR #1064, opened 2017-03-24
Allow scaling process to negative values to remove resources

ejholmes:
Closes https://github.com/remind101/empire/issues/1042

There's two changes here:

1. Allow processes to be scaled to a negative value, to remove the AWS resources.
2. Removes the condition where resources for scheduled jobs wouldn't be added to the template if the scale was 0. Now they're added, but the CloudWatch Events trigger is disabled if the process is scaled down.

### DIFF
--- a/cmd/emp/scale.go
+++ b/cmd/emp/scale.go
@@ -97,8 +97,8 @@ func runScale(cmd *Command, args []string) {
 		types[pstype] = true
 
 		opt := heroku.FormationBatchUpdateOpts{Process: pstype}
-		if qty != -1 {
-			opt.Quantity = &qty
+		if qty != nil {
+			opt.Quantity = qty
 		}
 		if size != "" {
 			opt.Size = &size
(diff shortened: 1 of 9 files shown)


cmd/emp/scale.go
```diff
@@ -154,25 +153,28 @@ func parseScaleArg(arg string) (pstype string, qty int, size string, err error)
...
 	} else {
 		if iColon > 0 {
-			qty, err = strconv.Atoi(rem[:iColon])
+			v, err := strconv.Atoi(rem[:iColon])
 			if err != nil {
-				return pstype, -1, "", errInvalidScaleArg
+				return pstype, nil, "", errInvalidScaleArg
 			}
+			qty = &v
+		} else {
```
phobologic: I'm guessing this blank else block isn't necessary?

ejholmes: Argh, no idea why I added this block. I'll remove it. Not necessary.

PR #1067, opened 2017-03-28
Differentiate between added/changed/removed env vars

ejholmes:
This makes it so that the SetEvent, and the release description have more information about what environment variables were added/changed/unset during `emp set/unset`. Previously, the SetEvent only said things were "changed", and the release description was buggy in that sometimes it would say "Set" and sometimes it would say "Unset". This made it hard to see what actually happened from `emp set`.

### DIFF
--- a/configs.go
+++ b/configs.go
@@ -124,17 +124,17 @@ type configsService struct {
 	*Empire
 }
 
-func (s *configsService) Set(ctx context.Context, db *gorm.DB, opts SetOpts) (*Config, error) {
+func (s *configsService) Set(ctx context.Context, db *gorm.DB, opts SetOpts) (*Config, *Config, error) {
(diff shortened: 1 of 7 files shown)


configs.go
```diff
@@ -204,24 +204,79 @@ func mergeVars(old, new Vars) Vars {
...
+	}
+
+	return fmt.Sprintf("%s config var%s", strings.Join(parts, " "), plural)
+}
+
+// DiffVars generates a diff between two Vars objects, treating `a` as the newer
+// version. It can tell you what was added, changed or removed between the two
+// objects.
+func DiffVars(a, b Vars) *VarsDiff {
+	var added, changed, removed sort.StringSlice
```
phobologic: Man, this stuff makes me miss sets from python (and other languages) :)

PR #1069, opened 2017-03-31
Add message to emp help run to prompt user to press enter

DavidLu1997:
This isn't immediately obvious and can confuse people (me). A simple message would help.

### DIFF
--- a/cmd/emp/run.go
+++ b/cmd/emp/run.go
@@ -33,7 +33,9 @@ Run a process on Heroku. Flags such as` + " `-a` " + `may be parsed out of
 the command unless the command is quoted or provided after a
 double-dash (--).
 
-You may need to press enter after this command to continue.
+When running an attached process that reads from stdin (like bash) you may experience a "hang".
+Usually, pressing a key like "enter" will flush the output to your terminal.
+See https://github.com/remind101/empire/issues/609
 
 Options:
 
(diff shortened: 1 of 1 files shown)


ejholmes:
This bug is definitely annoying, but I think if we're going to make changes here, a direct fix for the bug itself would be better. The problem is that, the bug is entirely contextual, and only shows up under certain conditions, so displaying this message when the bug _doesn't_ occur just adds more confusion.

I'd prefer we fix the issue itself, instead of working around it.

DavidLu1997:
I believe that in cases where the enter isn't needed, having an extra enter will not hurt anything (e.g. `run console`). Even if we do not add this, I think this should be at least documented somewhere.

ejholmes:
@DavidLu1997 yeah, I agree documenting the bug would be good. Do you mind updating this to just include documentation changes (e.g. `emp help run`)? When we fix the bug for real, we can take that out.

DavidLu1997:
@ejholmes done, in `emp help run`

cmd/emp/run.go
```diff
@@ -33,6 +33,8 @@ Run a process on Heroku. Flags such as` + " `-a` " + `may be parsed out of
 the command unless the command is quoted or provided after a
 double-dash (--).
 
+You may need to press enter after this command to continue.
```
ejholmes: Can we add a little more information to this? Right now, it's not clear _why_ you would need to do this. Maybe something like:

```
When running an attached process that reads from stdin (like bash) you may experience a "hang". Usually, pressing a key like "enter" will flush the output to your terminal. See https://github.com/remind101/empire/issues/609
```

DavidLu1997:
@ejholmes changes made

PR #1075, opened 2017-05-02
add support for reporting to rollbar

sumeet:
TODO:

- [ ] add rollbar access token from https://rollbar.com/Remind/empire/


cmd/empire/main.go
```diff
@@ -9,6 +10,9 @@ import (
 	"github.com/remind101/empire/server/github"
 )
 
+const hbExampleURL = "hb://api.honeybadger.io?key=<key>&environment=<environment>"
+const rollbarExampleURL = "rb://api.rollbar.com?key=<key>&environment=<environment>"
```
ejholmes: https://github.com/remind101/stacks/pull/683#discussion_r114231716.

sumeet: ah shit, fixed!

ejholmes:
Awesome. Looks good! Can you also add a line to [CHANGELOG.md](https://github.com/remind101/empire/blob/master/CHANGELOG.md#head)?

We're running this internally in our staging environment, I think we should give this a couple days and assuming no issues, we can merge.

sumeet:
@ejholmes sweet, sounds good. all set!

ejholmes:
Just adding some comments on my findings here:

It seems like a lot of errors end up getting grouped incorrectly in Rollbar. Originally thought that this may have been a bug in remind101/pkg/reporter/rollbar, and tried to test with https://github.com/remind101/empire/commit/293bd7a647b2a5c1eeedee5374fa50330759069c, however, errors still got clumped together.

Digging a little deeper, I noticed that all of the errors that got clumped together shared the same stack. In [stvp/rollbar](https://github.com/stvp/rollbar/blob/master/stack.go#L49) the fingerprint that gets generated for the error only takes the stack into account:

```golang
func (s Stack) Fingerprint() string {
	hash := crc32.NewIEEE()
	for _, frame := range s {
		fmt.Fprintf(hash, "%s%s%d", frame.Filename, frame.Method, frame.Line)
	}
	return fmt.Sprintf("%x", hash.Sum32())
}
```

I think this is ok, but ideally the fingerprint would be a combination of stack and error class, but it looks like there's no way to easily inject a custom fingerprint into stvp/rollbar.

PR #1081, opened 2017-05-05
docs improvements

iserko:
* Fix the link to the SAML markdown file
* Mention Production Best Practices in the Configuration document

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -2,6 +2,8 @@
 
 The following documents the various configuration parameters that you can use to tailor your Empire environment to your needs.
 
+If you're configuring for production, make sure to read through [production best practices](./production_best_practices.md).
+
 ### GitHub Authentication
 
 1. Create new OAuth application in Github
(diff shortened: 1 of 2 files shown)


docs/configuration.md
```diff
@@ -15,7 +17,7 @@ It's recommended that you also set either `EMPIRE_GITHUB_ORGANIZATION`, or `EMPI
 
 ### SAML Authentication
 
-Refer to the [docs](./saml) on configuring the SAML authentication backend.
+Refer to the [docs](./saml.md) on configuring the SAML authentication backend.
```
ejholmes: To make this work with readthedocs, I thing this needs to remain `./saml`, but a section needs to be added to https://github.com/remind101/empire/blob/master/mkdocs.yml#L10

iserko: I've added SAML to mkdocs.yml now. And from what I can see in https://github.com/remind101/empire/blob/master/docs/quickstart_using.md#empire--using ... it looks like you still need an `.md` (looking at the installing link)

PR #1082, opened 2017-05-05
Set Load Balancer internal/external per process instead of per application

### LINKED ISSUE #1076: Ability to set "exposure" by process
Currently "exposure" as it is phrased in the code is controlled by running:
- `emp domain-add <domain>` which makes ELB/ALB `internet-facing`
- `emp domain-remove <domain>` which makes ELB/ALB `internal` (or if you never use domain-add for an app

That setting is applied across the whole application and has no ability to set it per web process.

What I would like to see is to be able to control it via the Procfile using `EMPIRE_X_EXPOSURE` if it existed:
```
# internal
qint:
  command: ./bin/web
  ports:
    - "80:8080"
  environment:
    EMPIRE_X_LOAD_BALANCER_TYPE: "alb"
    EMPIRE_X_EXPOSURE: "internal"

# external
qext:
  command: ./bin/web
  ports:
    - "80:8080"
  environment:
    EMPIRE_X_LOAD_BALANCER_TYPE: "alb"
    EMPIRE_X_EXPOSURE: "internet-facing"  
```

iserko:
Introduces EMPIRE_X_EXPOSURE which needs to be `public`.

If that environment variable is not set, we fallback to the default selection via application (where you use `domain-add`).

Added this explanation in the Procfile doc.

Fixes #1076

Things working and tested:
- [x] Creating a new application using Procfile work as specified (one internal and one external ALB) https://github.com/iserko/flask_test/blob/f369c115fd41a0086d364f38b693a353963c041d/Procfile :white_check_mark:
- [x] Modifying a running application by setting this environment variable and deploying ✅ (new ALB/ELB gets created and the old one is **deleted**)
- [x] When using old Procfile format doing `env set EMPIRE_X_EXPOSURE=public` creates a public ALB/ELB
- [x] Migrating from old Procfile with an internal ALB/ELB to new Procfile format (LB is not recreated)
- [x] Migrating from old Procfile with an external ALB/ELB to new Procfile format (LB is not recreated)

The biggest thing is that this setting shouldn't be considered as harmless. If you change it ... your load balancer WILL get recreated.

@ejholmes Thoughts?

### DIFF
--- a/procfile/README.md
+++ b/procfile/README.md
@@ -85,5 +85,6 @@ inside the `Procfile` for a specific process.
 Name | Default value | Available options | Description
 -----|---------------|-------------------|------------
 `EMPIRE_X_LOAD_BALANCER_TYPE` | `elb` | `alb`, `elb`| Determines whether you will use an ALB or ELB
+`EMPIRE_X_EXPOSURE` | `private` | `private`, `public` | Sets whether your ALB or ELB will be public (internet-facing) or private (internal), the default is private, however if you have used the deprecated `domain-add` command then the load balancer will become public. **If you change this setting, the load balancer will be recreated as soon as you deploy**
 `EMPIRE_X_TASK_DEFINITION_TYPE` | not set | `custom` | Determines whether we use the Custom::ECSTaskDefinition (better explanation needed)
 `EMPIRE_X_TASK_ROLE_ARN` | not set | any IAM role ARN | Sets the IAM role for that app/process. **Your ECS cluster MUST have Task Role support enabled before this can work!**
--- a/releases.go
+++ b/releases.go
@@ -306,6 +306,11 @@ func releasesCreate(db *gorm.DB, release *Release) (*Release, error) {
 func newSchedulerApp(release *Release) (*twelvefactor.Manifest, error) {
 	var processes []*twelvefactor.Process
 
+	env := environment(release.Config.Vars)
+	env["EMPIRE_APPID"] = release.App.ID
+	env["EMPIRE_APPNAME"] = release.App.Name
+	env["EMPIRE_RELEASE"] = fmt.Sprintf("v%d", release.Version)
+
 	for name, p := range release.Formation {
 		if p.NoService {
 			// If the entry is marked as "NoService", don't send it
@@ -319,17 +324,12 @@ func newSchedulerApp(release *Release) (*twelvefactor.Manifest, error) {
 			continue
 		}
 
-		process, err := newSchedulerProcess(release, name, p)
+		process, err := newSchedulerProcess(release, name, p, env)
 		if err != nil {
 			return nil, err
 		}
 		processes = append(processes, process)
 	}
-
-	env := environment(release.Config.Vars)
-	env["EMPIRE_APPID"] = release.App.ID
-	env["EMPIRE_APPNAME"] = release.App.Name
-	env["EMPIRE_RELEASE"] = fmt.Sprintf("v%d", release.Version)
 
 	labels := map[string]string{
 		"empire.app.id":      release.App.ID,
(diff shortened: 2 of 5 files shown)


scheduler/cloudformation/template.go
```diff
@@ -54,6 +69,20 @@ func loadBalancerType(app *twelvefactor.Manifest, process *twelvefactor.Process)
 	return classicLoadBalancer
 }
 
+func schemeExposure(app *twelvefactor.Manifest, process *twelvefactor.Process) string {
+	env := twelvefactor.Env(app, process)
+	if v, ok := env["EMPIRE_X_EXPOSURE"]; ok {
```
ejholmes: I think we should move this check into https://github.com/remind101/empire/blob/master/releases.go#L456, where it can just set the appropriate `Exposure` on the `twelvefactor.Process`.

I've been intending on moving all of the `EMPIRE_X_*` checks from this file and into [releases.go](https://github.com/remind101/empire/blob/master/releases.go) so this would be a good start.

iserko:
@ejholmes I've just rebased and pushed additional changes.

The TargetGroup prefix was really easy to put it, so that wasn't the problem.
The biggest problem was moving the `EMPIRE_X_EXPOSURE` var into releases.go
At that point the `app` and `process` Environments aren't merged yet (it happens in `twelvefactor.go`). So I exposed the `merge` function by capitalizing it and using it in `releases.go`.

Please let me know if you have comments either via Github or if you have time in the next few hours I'll be online via Slack

procfile/README.md
```diff
@@ -78,3 +78,13 @@ This allows you to set process specific environment variables. If these are set
...
   EMPIRE_X_LOAD_BALANCER_TYPE: "alb"
 ```
+
+Supported environment variables that can either be set via `emp set` for the whole application or
+inside the `Procfile` for a specific process.
+
+Name | Default value | Available options | Description
+-----|---------------|-------------------|------------
+`EMPIRE_X_LOAD_BALANCER_TYPE` | `elb` | `alb`, `elb`| Determines whether you will use an ALB or ELB
+`EMPIRE_X_EXPOSURE` | `private` | `private`, `public` | Sets whether your ALB or ELB will be public (internet-facing) or private (internal), the default is private, however if you have used the deprecated `domain-add` command then the load balancer will become public. **If you change this setting, the load balancer will be recreated as soon as you deploy**
```
ejholmes: One thing that we've tried to do with the extended Procfile is maintain some level of "parity" with `docker-compose.yml`, so that concepts just kinda translate over (e.g. ports).

In docker-compose.yaml, [you can attach a container to a "network"](https://docs.docker.com/compose/compose-file/#networks) (like `external` or `internal`), which I think maps pretty nicely to how we use internal/external ELB's. I wonder if we should just use a `networks` key to determine whether the process should be exposed external/internal.

For example:

```yaml
web:
  command: ./bin/web
  networks:
    - external # Expose this process with an "external" ELB.
admin:
  command: ./bin/web
  networks:
    - internal # Expose this process with an "internal" ELB
```

Just a thought, totally fine with starting with some `EMPIRE_X` env vars.

releases.go
```diff
@@ -365,14 +365,25 @@ func newSchedulerProcess(release *Release, name string, p Process) (*twelvefacto
 		exposure *twelvefactor.Exposure
 		err      error
 	)
+	mergedEnv := twelvefactor.Merge(appEnv, env)
+
+	isExternal := release.App.Exposure == exposePublic
+	if v, ok := mergedEnv["EMPIRE_X_EXPOSURE"]; ok {
```
ejholmes: Can we should add some tests similar to https://github.com/remind101/empire/blob/master/tests/empire/empire_test.go#L77-L180 that exercises this?

procfile/README.md
```diff
@@ -78,3 +78,13 @@ This allows you to set process specific environment variables. If these are set
 environment:
   EMPIRE_X_LOAD_BALANCER_TYPE: "alb"
 ```
+
+Supported environment variables that can either be set via `emp set` for the whole application or
```
ejholmes: Thanks for documenting these!

releases.go
iserko: yep definitely makes sense. Let me have a stab at it

procfile/README.md
iserko: I like that idea too

PR #1086, opened 2017-06-01
Maintenance mode

ejholmes:
This adds support for putting applications into "maintenance mode", which can be useful for doing things that might require complete downtime, like DB upgrades.

You can put an application into maintenance mode using the `emp maintenance-enable` command, which masks the scale of all processes to 0. You can then disable maintenance mode with `emp maintenance-disable`, which unmasks the scale and returns everything back to normal.


tests/cli/maintenance_test.go
```diff
@@ -0,0 +1,53 @@
...
+			"maintenance-enable -a acme-inc",
+			"Enabled maintenance mode on acme-inc.",
+		},
+		{
+			"ps -a acme-inc",
+			``,
+		},
+		{
+			"scale -l -a acme-inc",
+			"rake=0:1X scheduled=0:1X web=2:1X worker=0:1X",
```
phobologic: Should we have some indication here that the stack is in maintenance mode so people aren't confused?

ejholmes: I kinda waffled on that a bit. My current feeling is that this shouldn't change, since it would be a breaking change for scripts. We could return something like `rake=0:1X scheduler=0:1X web=0:1X worker=0:1X`, but then that's not _entirely_ true, because web is still scaled to 2 in the release.

PR #1094, opened 2017-07-28
Add flag for stack name template

ejholmes:
Previously, the `--environment` flag was used to generate a template for determining the CloudFormation stack name for an application. This can be a little constraining and overloaded the responsibilities of `--environment`.

With this change, you can now directly specify a stack name template like this:

```console
EMPIRE_CLOUDFORMATION_STACK_NAME_TEMPLATE="my-custom-prefix-{{.Name}}"
```

Empire will still fallback to including `--environment` in the prefix if `--cloudformation.stack-name-template` isn't provided.

### DIFF
--- a/cmd/empire/main.go
+++ b/cmd/empire/main.go
@@ -73,6 +73,8 @@ const (
 
 	FlagRoute53InternalZoneID = "route53.zoneid.internal"
 
+	FlagCloudFormationStackNameTemplate = "cloudformation.stack-name-template"
+
 	FlagSNSTopic           = "sns.topic"
 	FlagCloudWatchLogGroup = "cloudwatch.loggroup"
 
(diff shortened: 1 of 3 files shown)


cmd/empire/main.go
```diff
@@ -376,6 +378,12 @@ var EmpireFlags = []cli.Flag{
 		EnvVar: "EMPIRE_ROUTE53_INTERNAL_ZONE_ID",
 	},
 	cli.StringFlag{
+		Name:   FlagCloudFormationStackNameTemplate,
+		Value:  "",
+		Usage:  "If provided, this should be a Go text/template that will be used to generate a CloudFormation stack name for an application.",
```
phobologic: maybe a comment about what happens if not specified in the usage?

PR #1096, opened 2017-07-29
Configurable instance port pool

ejholmes:
This adds options for configuring the range of EC2 instance ports that Empire manages and allocates to applications that use legacy ELB's to prevent port conflicts.

This makes it easier to run multiple Empire daemons that schedule work into a single ECS cluster. Each Empire daemon can now specify it's own range of instance ports to allocate from to prevent overlap.

Note that, once configured, changing the flags will have no effect, so this only applies to new installations of Empire. Previous installations will continue to use the 9000-10000 range.

### DIFF
--- a/cmd/empire/factories.go
+++ b/cmd/empire/factories.go
@@ -37,7 +37,19 @@ import (
 // DB ===================================
 
 func newDB(c *Context) (*empire.DB, error) {
-	return empire.OpenDB(c.String(FlagDB))
+	db, err := empire.OpenDB(c.String(FlagDB))
+	if err != nil {
+		return nil, err
+	}
+
+	db.Schema = &empire.Schema{
+		InstancePortPool: &empire.InstancePortPool{
+			Start: uint(c.Int(FlagInstancePortPoolStart)),
+			End:   uint(c.Int(FlagInstancePortPoolEnd)),
+		},
+	}
+
+	return db, nil
 }
 
 // Empire ===============================
--- a/migrations.go
+++ b/migrations.go
@@ -4,12 +4,77 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
+	"sort"
 
 	"github.com/lib/pq/hstore"
 	"github.com/remind101/empire/internal/migrate"
 	"github.com/remind101/empire/pkg/constraints"
 	"github.com/remind101/empire/procfile"
 )
+
+const (
+	DefaultInstancePortPoolStart = 9000
+	DefaultInstancePortPoolEnd   = 10000
+)
+
+// DefaultSchema is the default Schema that will be used to migrate the database
+// if none is provided.
+var DefaultSchema = &Schema{
+	InstancePortPool: &InstancePortPool{
+		Start: DefaultInstancePortPoolStart,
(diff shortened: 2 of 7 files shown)


migrations.go
```diff
@@ -4,13 +4,78 @@ import (
...
+}
+
+type InstancePortPool struct {
+	Start, End uint
+}
+
+type Schema struct {
+	// For legacy ELB's (not ALB) Empire manages a pool of host ports to
+	// ensure that all applications have a unique port on the EC2 instance.
+	// This option specifies the beginning of that range.
```
phobologic: not just the beginning

migrations.go
```diff
@@ -4,13 +4,78 @@ import (
...
+}
+
+func (s *Schema) migrations() []migrate.Migration {
+	ports := migrate.Migration{
+		ID: 4,
+		Up: func(tx *sql.Tx) error {
+			const table = `CREATE TABLE ports (
+  id uuid NOT NULL DEFAULT uuid_generate_v4() primary key,
+  port integer,
+  app_id uuid references apps(id) ON DELETE SET NULL
```
phobologic: should it also map to a process as well?

migrations.go
```diff
@@ -4,13 +4,78 @@ import (
...
+// latestSchema returns the schema version that this version of Empire should be
+// using.
+func (s *Schema) latestSchema() int {
+	migrations := s.migrations()
+	return migrations[len(migrations)-1].ID
+}
+
+func (s *Schema) migrations() []migrate.Migration {
+	ports := migrate.Migration{
+		ID: 4,
```
phobologic: Is this changing an older migration?  Is that going to cause any issues?

cmd/empire/factories.go
```diff
@@ -37,7 +37,19 @@ import (
...
-	return empire.OpenDB(c.String(FlagDB))
+	db, err := empire.OpenDB(c.String(FlagDB))
+	if err != nil {
+		return nil, err
+	}
+
+	db.Schema = &empire.Schema{
+		InstancePortPool: &empire.InstancePortPool{
+			Start: uint(c.Int(FlagInstancePortPoolStart)),
+			End:   uint(c.Int(FlagInstancePortPoolEnd)),
```
phobologic: Is this where we'd catch it if one of the two vars isn't defined?  Might be useful to have some helpful warning/error messages in that case.

migrations.go
ejholmes: It's explained in the PR description, but, for existing installations of Empire, this doesn't change anything. For new installations of Empire, you can set these options to pre-configure the pool when the migration runs.

migrations.go
ejholmes: This actually gets updated in a later migration: https://github.com/remind101/empire/blob/master/migrations.go#L250-L251

phobologic: Ahhh, ok - makes sense.

cmd/empire/factories.go
ejholmes: In any case where you provided invalid values here, the migration would fail, and you'd see an error in logs.

migrations.go
ejholmes: Good catch. Updated.

PR #1100, opened 2017-08-02
moved Procfile environment variables doc

iserko:
Follow up from #1099 moving these into docs/

Also added a note about them being experimental features and may change in the future.

@ejholmes


ejholmes:
Looks great. Thanks @iserko!

Issue #247, opened 2015-03-20
Move to immutable identifiers


ejholmes:
This is a ways off, but something to keep on radar. Docker 1.6 will support immutable identifiers. See https://github.com/docker/docker/pull/11109

ejholmes:
Gonna close this as a wontfix for now, mainly because `@sha256` isn't supported in ECS and this hasn't felt like a real problem.

ejholmes:
I'm re-opening this, since it's now supported in ECS as of ~May: https://github.com/aws/amazon-ecs-agent/issues/104#issuecomment-301203089

I think this is pretty important from a security perspective, since digests solve various attack vectors like mitm attacks, or credential exposure of your registry.

PR #1104, opened 2017-09-02
Automatically resolve Docker images to immutable digests

ejholmes:
Closes https://github.com/remind101/empire/issues/247

With this change, when you deploy an image using a mutable tag (e.g. `emp deploy remind101/acme-inc:latest`), Empire will resolve the image reference to an immutable reference (e.g. `remind101/acme-inc@sha256:c6f77d2098bc0e32aef3102e71b51831a9083dd9356a0ccadca860596a1e9007`) before saving the slug. This has a number of benefits, both for security and for stability.

This was always the intention, but couldn't be done until ECS supported digests (which it now does). 

In this PR, the first commit is a small refactor to change the `ProcfileExtractor` interface to an `ImageRegistry` interface. And then the second commit is a small change that implements digest resolution.

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -23,6 +23,9 @@ import (
 type dockerDaemon struct {
 	docker    *dockerutil.Client
 	extractor empire.ProcfileExtractor
+
+	// can be set to false to disable Pull-before-Resolve.
+	noPull bool
 }
 
 // DockerDaemon returns an empire.ImageRegistry that uses a local Docker Daemon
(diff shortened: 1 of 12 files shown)


tests/empire/empire_test.go
```diff
@@ -207,72 +204,6 @@ func TestEmpire_Deploy_ImageNotFound(t *testing.T) {
 	s.AssertExpectations(t)
 }
 
-func TestEmpire_Deploy_Concurrent(t *testing.T) {
```
ejholmes: I'm removing this test, mainly because it's been buggy and fails randomly (pretty sure it's the test itself, and not what it's actually testing). I'll try to follow up in the future with a better version of this test.

registry/registry.go
```diff
@@ -16,28 +17,68 @@ import (
...
 
-		return e.Extract(ctx, img, w)
-	})
+	i, err := r.docker.InspectImage(img.String())
+	if err != nil {
+		return img, err
+	}
+
+	if len(i.RepoDigests) <= 0 {
+		return img, &DigestError{Image: img.String()}
```
ejholmes: Turns out RepoDigests is only populated from a `docker pull` in Docker 1.12+. This thread https://github.com/moby/moby/issues/15508 explains the details of that.

We've been locked to 1.11.x for some time now, just for stability, but should probably consider upgrading to a later version.

Instead of returning an error here, this could just fallback to the existing behavior instead.

phobologic: I was just wondering about this - yeah, seems like falling back to existing behavior - or possibly having it be able to configure that you require digests - would be good.

phobologic:
Also could use some docs/etc.

ejholmes:
Ok, really happy with this now. It now falls back to the older behavior if no digests are available, and prints info in the deployment stream. If the underlying Docker Daemon supports digests, you'll see:

```console
...
Status: Resolved remind101/acme-inc:master to remind101/acme-inc@sha256:c6f77d2098bc0e32aef3102e71b51831a9083dd9356a0ccadca860596a1e9007
```

If there are no digests:


```console
...
Status: Image has no repository digests. Using remind101/acme-inc:master as image identifier
```

Also updated the relevant docs.

phobologic:
Should there be a way to configure Empire so it only accepts images with a digest? Not sure what the implications are in the fall back mode, or if it exposes us to any potential security vulnerabilities.

ejholmes:
> Should there be a way to configure Empire so it only accepts images with a digest

That'd probably be good. I agree the fallback mode is a little fishy, since it could potentially be used as an attack vector.

For the fallback, I think either a flag to enforce digests, or change the logic to use the Docker API version (or a combination of the both).

ejholmes:
Alrighty, updated this with a `--docker.digests` flag (`DOCKER_DIGESTS` env var) to control the behavior. Default is to "prefer" digests if the image has one, but can be changed to a more secure "enforce", or disabled entirely with "disable".

I'll run this in our staging env for a couple days and merge if no issues come up.

ejholmes:
This is working well in staging. Gonna merge.

PR #1109, opened 2017-10-04
Log/Instrument authentication failures

ejholmes:
This can be helpful for audit logging, and generating alerts on authentication failures. If a request fails to authenticate, you'll see a log line like:

```console
msg=authentication.failure request_id=7caf77ca-fdc3-4ddc-8c75-6ea0b2f319be username=ejholmes remote_ip=172.19.0.1 err="auth: forbidden"
```

Additionally, it extends the `request.start` log line with some additional fields that can be useful for debugging/auditing (user agent, remote ip):

```console
lvl=info msg=request.start request_id=67803998-4a38-42a4-9fd8-c56fb292e2e7 method=GET path=/apps user_agent="hk/0.12.0 (darwin; amd64) heroku-go/0.10.2 (darwin; amd64)" remote_ip=172.19.0.1
```

### DIFF
--- a/internal/realip/realip_test.go
+++ b/internal/realip/realip_test.go
@@ -1,0 +1,86 @@
+package realip
+
+import (
+	"net/http"
+	"strings"
+	"testing"
+)
+
+func TestIsLocalAddr(t *testing.T) {
+	testData := map[string]bool{
+		"127.0.0.0":   true,
+		"10.0.0.0":    true,
+		"169.254.0.0": true,
+		"192.168.0.0": true,
+		"::1":         true,
+		"fc00::":      true,
+
+		"172.15.0.0": false,
+		"172.16.0.0": true,
+		"172.31.0.0": true,
+		"172.32.0.0": false,
+
+		"147.12.56.11": false,
+	}
+
+	for addr, isLocal := range testData {
+		if isLocalAddress(addr) != isLocal {
+			format := "%s should "
+			if !isLocal {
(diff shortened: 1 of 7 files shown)


internal/realip/realip_test.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	for remoteAddr, expectedAddr := range testData {
+		if actualAddr := ipAddrFromRemoteAddr(remoteAddr); actualAddr != expectedAddr {
+			t.Errorf("ipAddrFromRemoteAddr of %s should be %s but get %s", remoteAddr, expectedAddr, actualAddr)
+		}
+	}
+}
+
+func TestRealIP(t *testing.T) {
```
ejholmes: I pulled this implementation from https://github.com/tomasen/realip, but it's a pretty poor implementation after reviewing:

1. X-Real-Ip is easily spoofed when on ELB.
2. X-Forwarded-For is implemented incorrectly, and checks the chain in the wrong order, which makes it spoofable on ELB.

It's a pretty small issue, given that this is only being used for logs, but could be better. Really, there's no good way to do it without having a trusted proxy sign the request headers.

ejholmes: Ok, last commit adds some security enhancements. Primarily:

1. Headers are untrusted by default, and need to be whitelisted.
2. X-Forwarded-For is no longer spoofable.

PR #1129, opened 2018-08-10
Support soft deletions of apps

ejholmes:
Before this PR, deleting an app would permanently delete all of the history. For something like Empire, it's much better to never delete anything, so there's always an audit trail and complete history of changes.


russellballestrini:
Basically instead of deleting records we are adding a new column `deleted_at` which is null by default. If null the app is not deleted, if it has a timestamp it is considered deleted but all data, history, and references still exist.

PR #1134, opened 2019-02-21
Add opts.Env to proc.Environment for task runner.

russellballestrini:
This should fix the issue we have with the terminal width environment
var not being forwarded / set.

tested in staging

### DIFF
--- a/runner.go
+++ b/runner.go
@@ -61,6 +61,16 @@ func (r *runnerService) Run(ctx context.Context, opts RunOpts) error {
 		proc.SetConstraints(*opts.Constraints)
 	}
 
+	// initialize the proc environment map with string keys and string values.
+	if proc.Environment == nil {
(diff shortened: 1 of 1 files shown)


runner.go
```diff
@@ -61,6 +61,11 @@ func (r *runnerService) Run(ctx context.Context, opts RunOpts) error {
 		proc.SetConstraints(*opts.Constraints)
 	}
 
+	// Add additional environment variables to the process.
+	for k, v := range opts.Env {
```
hamiltop: where/how are these defined? in the procfile?

russellballestrini:
@hamiltop this PR crashes, could you help me form a valid fix? Here are some references.

* https://github.com/remind101/empire/blob/internal-v2/runner.go#L22-L67
* https://github.com/remind101/empire/blob/master/runner.go#L72-L82

PR #1135, opened 2019-02-28
Add rollback support

hamiltop:
Implements Rollback support for our internal branch.

**Russell's Update**

This implementation is _slow_ because it does a _lot_ of github API querying. The rollback command takes about 30-45 seconds to complete in my testing environment , but it might take even long in our real environment.

I suppose this is more of a proof of concept but we would likely need a better way of resolving version numbers to commits.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,6 +59,9 @@ jobs:
 
     steps:
       - checkout
+      - run:
+          name: Install System Dependencies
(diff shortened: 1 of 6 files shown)


russellballestrini:
Ok I was able to test this branch locally and I got the following error:

```
laptop-195:empire russellballestrini$ emp-local rollback -a acme-inc v7
error: ReleasesFind not implemented
```

russellballestrini:
```
$ emp-local rollback -a acme-inc v28 -m "asdf"
Rolled back acme-inc to v28 as v31.
```

PR #1137, opened 2019-03-12
Produce statically-linked emp and empire binaries

### LINKED ISSUE #1068: emp binary doesn't run on Alpine Linux based containers
Alpine Linux is the preferred base image for docker containers, and a lot of projects are migrating. Alpine always supported golang binaries. 

However, looks like you are doing something strange and incompatible.

Tested :
( alpine image with wget and openssl installed ) 
```bash
wget https://github.com/remind101/empire/releases/download/v0.12.0/emp-Linux-x86_64
chmod +x emp-Linux-x86_64
./emp-Linux-x86_64 
bash: ./emp-Linux-x86_64: No such file or directory
```
After some investigation, looks like your binary is not statically linked and strings are showing this:
```bash
strings emp-Linux-x86_64 | grep -e '^/' | grep lib
/lib64/ld-linux-x86-64.so.2

ls /lib64/ld-linux-x86-64.so.2
ls: /lib64/ld-linux-x86-64.so.2: No such file or directory
```
When symlinking that path to a ld-mucl, it starts acting normally
```bash
mkdir /lib64 ; ln -s /lib/ld-musl-x86_64.so.1 /lib64/ld-linux-x86-64.so.2
./emp-Linux-x86_64 
Usage: emp <command> [-a <app or remote>] [options] [arguments]


Commands:

    create             create an app
    apps         ...
```

aengelas:
This change sets flags on the linker to ensure that all libraries are
statically linked. It also forces the compiler to use the Go
implementations of networking libraries (the default).

Making these changes allows the binaries to be run without libc in a
scratch Docker container, or in distributions like Alpine that don't
include libc.

Fixes #1068

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 **Improvements**
 
+* [cmd/empire] Produce statically-linked binaries
+* [cmd/emp] Produce statically-linked binaries
 * [cmd/empire] The internal upper bound constraint for CPU shares was removed. [#1124](https://github.com/remind101/empire/pull/1124)
 
 ## 0.13.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,6 @@
 
 **Improvements**
 
-* [cmd/empire] Produce statically-linked binaries
 * [cmd/emp] Produce statically-linked binaries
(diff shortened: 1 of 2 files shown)


aengelas:
I've verified that this produces statically-linked binaries that will execute on Alpine, but I'm not that familiar with the codebase, and I'm not sure what implications this may have.

aengelas:
Okay, it looks like this doesn't work for the empire binary, but tests pass for the emp command.

PR #1141, opened 2019-03-27
Upgrade base docker image

russellballestrini:


server/auth/github/client_test.go
```diff
@@ -1,3 +1,5 @@
+// +build !go1.10
```
russellballestrini: @ejholmes any clue what this does? I copied it from `internal-v2` and now the tests pass?!

PR #1153, opened 2020-08-28
Allow passing DeploymentConfiguration to ECS tasks

aengelas:
This should read the DeploymentConfiguration from the payload of the
custom resource.


aengelas:
I'm not sure if this is enough to get things working. There will (hopefully) be some test failures to fix, but it's not playing nicely with my go install right now.

aengelas:
I got tests working locally, and they pass, but currently the circle config requires docker hub access, which isn't available from my fork.

PR #1154, opened 2020-08-29
Added request logging to ecs custom resource.

russellballestrini:
also ran `go fmt`

	modified:   server/cloudformation/ecs.go

### DIFF
--- a/server/cloudformation/ecs.go
+++ b/server/cloudformation/ecs.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"log"
 	"strings"
 
 	"golang.org/x/net/context"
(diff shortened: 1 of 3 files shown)


server/cloudformation/ecs.go
```diff
@@ -95,6 +95,9 @@ func newECSServiceProvisioner(resource *ECSServiceResource) *provisioner {
 }
 
 func (p *ECSServiceResource) Create(ctx context.Context, req customresources.Request) (string, interface{}, error) {
+	// TODO: how to make this a debug log level?
+	fmt.Printf("%+v\n", req)
```
aengelas, line 99:
```suggestion
	fmt.Printf("%+v\n", *req.PropagateTags)
```

I think you need to do that. That'll print the actual string referenced by req.PropagateTags.

aengelas:
This looks like it works, but I think that there's an edge case where you pass a customresources reference but the struct is missing one or more fields. I think that'll be a nil pointer error. I'll take a look at that in the morning. I _think_ this should unblock you testing, @russellballestrini.

PR #1155, opened 2020-09-01
DeploymentConfiguration to ECS Services Custom Resource

russellballestrini:
parse DeploymentConfiguration and pass it to the create ECS Service API call.

### DIFF
--- a/server/cloudformation/ecs.go
+++ b/server/cloudformation/ecs.go
@@ -52,15 +52,16 @@ type LoadBalancer struct {
 // ECSServiceProperties represents the properties for the Custom::ECSService
 // resource.
 type ECSServiceProperties struct {
-	ServiceName          *string
-	Cluster              *string
-	DesiredCount         *customresources.IntValue `hash:"ignore"`
-	LoadBalancers        []LoadBalancer
-	Role                 *string
-	TaskDefinition       *string `hash:"ignore"`
-	PlacementConstraints []ECSPlacementConstraint
-	PlacementStrategy    []ECSPlacementStrategy
-	PropagateTags        *string
+	Cluster                 *string
+	DeploymentConfiguration *DeploymentConfiguration  `hash:"ignore"`
+	DesiredCount            *customresources.IntValue `hash:"ignore"`
+	LoadBalancers           []LoadBalancer
+	PlacementConstraints    []ECSPlacementConstraint
+	PlacementStrategy       []ECSPlacementStrategy
+	PropagateTags           *string
+	Role                    *string
+	ServiceName             *string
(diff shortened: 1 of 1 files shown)


aengelas:
I can make the additional required changes. It was my changes that caused the issue.

russellballestrini:
https://github.com/remind101/empire/pull/1158#discussion_r483734775 thanks for calling out that I didn't revert the problem PR @aengelas - preparing a PR to do so now.

aengelas:
I don't want to review this, since I wrote a bunch off it.

aengelas:
Related AWS support thread about strings in cf custom resources: https://remind.slack.com/archives/CDNSXHEEL/p1598398372094500

june-remind101 (approved):
LGTM, maybe a test or two around handling errors

aengelas:
@hamiltop @bitglue  Do we want to continue with this and get it deployed? Seems like betterscaling might make this unnecessary?

hamiltop:
We've had r101-api in prod with 150% max percent and I'm not sure how much it's helped or hurt.

bitglue:
Since betterscaling rapidly terminates capacity after api deploys, the importance of reducing that "bump" from an api deploy is lower. But it would still have some benefit: currently there's about 18 minutes where the cluster is significantly overprovisioned to accommodate the deploy. There's also significant shuffling of containers around at the end of the deploy when the excess compute instances are terminated.

The other concern though is if we'd even want to make all API deployments slower just to reduce that bump.

If we did merge this functionality, I might use it to deploy betterscaling with an at-most-1 policy, rather than at-least-1. That's a pretty minor concern though: 2 betterscalings running just thrashes the ASG a bit, sometimes.

If we adopt BANDAID we may end up needing something like this since the max deployment percentage can't go over 100% for a daemon service. But, BANDAID may work with an entirely different deployment system.

I guess my feeling is there isn't a compelling use case for this functionality, but if it's implemented and works there's not much downside to merging it.

bitglue:
> We've had r101-api in prod with 150% max percent and I'm not sure how much it's helped or hurt.

From my casual observation it does not seem to have made much of a difference either way. I suspect the rate limiting you've been investigating and/or the termination behavior of API is (still) the limiting factor in deployment speed, so we wouldn't see any change until the numbers are reduced more.

russellballestrini:
If I remember correctly that 150% set in stacks isn't applied, we might
have applied it outside of stacks but this would need to be merged if we
want stacks to manage and maintain this value.

june-remind101:
@remind101/ops-engineering Safe to close this at this point I assume?

bitglue:
yes

PR #1157, opened 2020-09-03
Allow empire to redirect oauth callbacks to another host

jherico:
Our OAuth setup uses the same token for Empire V1 and Empire V2.  However, the token is associated with a callback URL that can only target one server, V1 _or_ V2, and it currently points at V1.  This new setting will allow the V1 server to redirect the OAuth callbacks to the V2 server so it can handle them.  

Includes changes from #1158 so that the build doesn't fail.  That PR should be merged first.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -5,6 +5,7 @@ package server
 
 import (
 	"io"
+	"log"
 	"net/http"
 	"net/url"
 
@@ -57,9 +58,10 @@ func New(e *empire.Empire, options Options) *Server {
 
 	if options.OauthRedirectURL != "" {
 		parsedUrl, err := url.Parse(options.OauthRedirectURL)
-		if err == nil {
-			s.OauthRedirectURL = parsedUrl
+		if err != nil {
+			log.Fatal(err)
 		}
+		s.OauthRedirectURL = parsedUrl
 	}
 	if options.GitHub.Webhooks.Secret != "" {
 		// Mount GitHub webhooks
@@ -109,7 +111,7 @@ func (s *Server) handler(r *http.Request) http.Handler {
 		return s.Heroku
 	}
 
-	if s.OauthRedirectURL != nil && r.URL.Path =="/oauth/exchange" {
+	if r.URL.Path =="/oauth/exchange" && r.FormValue("code") != "" && s.OauthRedirectURL != nil   {
 		return http.HandlerFunc(s.redirectOauth)
 	}
 
--- a/server/server.go
+++ b/server/server.go
@@ -19,6 +19,7 @@ var (
 )
 
 type Options struct {
+	OauthRedirectURL string
 	GitHub struct {
 		// Deployments
 		Webhooks struct {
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -75,6 +84,21 @@ func (s *Server) Handler(r *http.Request) http.Handler {
 	return h
 }
 
+func (s *Server) redirectOauth(w http.ResponseWriter, req *http.Request) {
+	// Shallow copy the existing URL
+	newDest := req.URL
+	// Replace the hostname with the provided hostname
+	newDest.Host = s.OauthRedirectURL.Host
```
june-remind101, line 91: For completeness you may want to also include the port from `OauthRedirectUR` on `newDest`

jherico: I don't follow.  The url.URL.Host field is both the hostname and the port, if any exists, so there's no extra field to copy.

server/server.go
```diff
@@ -47,11 +48,19 @@ type Server struct {
...
 
 func New(e *empire.Empire, options Options) *Server {
 	s := &Server{}
 
+	if options.OauthRedirectURL != "" {
+		parsedUrl, err := url.Parse(options.OauthRedirectURL)
+		if err == nil {
+			s.OauthRedirectURL = parsedUrl
+		}
+	}
```
aengelas, line 63: I think we should handle the error case here. This could result in very frustrating behavior where the url is set but invalid and emp v1 just ignores it. I'd do something like 

```suggestion
		parsedUrl, err := url.Parse(options.OauthRedirectURL)
		if err != nil {
		        // return some sort of appropriate error
		}
			
		s.OauthRedirectURL = parsedUrl
	}
```

aengelas: I think that'd be more idiomatic for go, following the `if err != nil` pattern.

server/server.go
```diff
@@ -85,6 +109,10 @@ func (s *Server) handler(r *http.Request) http.Handler {
 		return s.Heroku
 	}
 
+	if s.OauthRedirectURL != nil && r.URL.Path =="/oauth/exchange" {
```
aengelas, line 112: Is the path consistent/reliable enough to use for checking here? Is there a header or something that indicates the request is part of an oauth exchange? It also might be more efficient to reverse the order of these checks, since presumably we'll usually have the redirect URL set in deployed environments?

jherico: The incoming request should have a GET verb and the request should contain a `code` parameter.  I'm not sure how to extract the request type from the request though.

aengelas: that seems more fragile than what you're doing here, so if this consistently works, probably best to just leave it. If it's coming from github though, I'm surprised they haven't set some X- custom headers

server/server.go
jherico: fixed

aengelas (approved):

server/server.go
jherico: I've reordered and added the check for the code parameter.  While it may have some custom headers as well, if that's not well documented behavior [here](https://docs.github.com/en/developers/apps/authorizing-oauth-apps#2-users-are-redirected-back-to-your-site-by-github) then I'd rather not rely on it, since it could technically change at any time.

server/server.go
june-remind101: Ah, I always use `Hostname()` Never mind me.

PR #1158, opened 2020-09-04
Fix CI build failures

jherico:
There are some test failures that are happening on all builds, even master.  They appear to be the result of bitrot with the Dockerfiles becoming outdated and not working with updated images.  

This attempts to address some of them by adding some missing precision in the postgres source image, adding the postgresql-client package to the test library, since it depends on `pg_dump` and fixing an outright bug in code that launches and external program when the SHELL environment variable isn't present.

### DIFF
--- a/cmd/emp/hkclient/creds.go
+++ b/cmd/emp/hkclient/creds.go
@@ -11,7 +11,7 @@ import (
 )
 
 type NetRc struct {
-	netrc.Netrc
+	*netrc.Netrc
 }
 
 func netRcPath() string {
@@ -26,12 +26,12 @@ func LoadNetRc() (nrc *NetRc, err error) {
 	onrc, err := netrc.ParseFile(netRcPath())
 	if err != nil {
 		if os.IsNotExist(err) {
-			return &NetRc{}, nil
+			return &NetRc{&netrc.Netrc{}}, nil
 		}
 		return nil, err
 	}
 
-	return &NetRc{*onrc}, nil
+	return &NetRc{onrc}, nil
 }
 
 func (nrc *NetRc) GetCreds(apiURL *url.URL) (user, pass string, err error) {
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -14,4 +14,7 @@ services:
     environment:
       TEST_DATABASE_URL: "postgres://postgres:postgres@db/postgres?sslmode=disable"
   db:
-    image: postgres
+    image: postgres:9
+    environment:
+      POSTGRES_HOST_AUTH_METHOD: trust
+
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -14,7 +14,8 @@ services:
     environment:
       TEST_DATABASE_URL: "postgres://postgres:postgres@db/postgres?sslmode=disable"
   db:
-    image: postgres:9
+    # Postgres 9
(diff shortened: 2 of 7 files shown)


russellballestrini:
Hey @jherico this seems to superseed this PR as well? https://github.com/remind101/empire/pull/1155

cmd/emp/hkclient/creds.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 type NetRc struct {
-	netrc.Netrc
```
aengelas, line 14: Why are we wrapping Netrc in our own struct?

docker-compose.test.yml
```diff
@@ -14,4 +14,7 @@ services:
     environment:
       TEST_DATABASE_URL: "postgres://postgres:postgres@db/postgres?sslmode=disable"
   db:
-    image: postgres
+    image: postgres:9
```
aengelas, line 17: Would be even more reliable to use the sha256 pinning here; but then we wouldn't get patches. If this is for testing, might be better to keep it consistent.

server/cloudformation/ecs_test.go
```diff
@@ -59,6 +59,10 @@ func TestECSServiceResource_Create(t *testing.T) {
 		ServiceName:  aws.String("acme-inc-web-dxRU5tYsnzt"),
 		Cluster:      aws.String("cluster"),
 		DesiredCount: aws.Int64(1),
+		DeploymentConfiguration: &ecs.DeploymentConfiguration{
+			MaximumPercent: aws.Int64(100),
+			MinimumHealthyPercent: aws.Int64(80),
+		},
```
aengelas, line 65: How come you added this here? We don't currently use DeploymentConfiguration (afaik) in empire. The PR Russell mentioned is working on adding it.

cmd/emp/hkclient/creds.go
jherico: I have no idea.  I only modified this to quiet the warning, I don't want to visit it as an architectural decision.

aengelas: I don't either; I was just hoping you had some knowledge I could absorb...

server/cloudformation/ecs_test.go
jherico: The code [here](https://github.com/remind101/empire/blob/master/server/cloudformation/ecs.go#L187) means that unless a `DeploymentConfiguration` with a non-nil `MaximumPercent` and `MinimumHealthyPercent` is present, the ECSServiceResource Update method will panic on accessing the `MaximumPercent.Value()` of a nil `DeploymentConfiguration`

I could have added more checking in the `Update` method, but because I'm not super familiar with either Go or the Empire guts, I opted to try to fix it by only modifying the test, even though that was more extensive.

aengelas: @russellballestrini did you never revert your PR that caused those problems? Sorry Brad, I'd assumed that code had been removed. If that's still in there, master is quite broken ATM.

docker-compose.test.yml
jherico: fixed

aengelas (approved):

server/cloudformation/ecs_test.go
russellballestrini: I didn't revert it, I assumed we wanted to fail forward since you had a PR in draft which didn't have the needed changes from the original bad commit. @aengelas what do you prefer we do in this case?

PR #1163, opened 2020-10-01
Create codeql-analysis.yml

aengelas:
Try adding codeql github workflow, as a potentially free alternative to sonarcloud.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,71 +1,51 @@
-# For most projects, this workflow file will not need changing; you simply need
-# to commit it to your repository.
-#
-# You may wish to alter this file to override the set of languages analyzed,
-# or to provide custom queries or build logic.
-name: "CodeQL"
+name: "Code Scanning - Action"
 
 on:
   push:
-    branches: [master, internal-v2, stable-0.10, stable-0.11, stable-0.12, stable-0.13]
   pull_request:
-    # The branches below must be a subset of the branches above
-    branches: [master]
   schedule:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+      uses: actions/checkout@v2
+      with:
+        # We must fetch at least the immediate parents so that if this is
+        # a pull request then we can checkout the head.
+        fetch-depth: 2
+
+    # If this run was triggered by a pull request event, then checkout
+    # the head of the pull request instead of the merge commit.
+    - run: git checkout HEAD^2
+      if: ${{ github.event_name == 'pull_request' }}
```
isobit, line 42: This seems odd, wouldn't we want to test the result of the intended merge rather than the pre-merge?

aengelas: yeah, it does. the comment also seems to be saying something other than the action. this is just the default config i dropped in

isobit: Where did you get this config? The one in https://github.com/github/codeql-action#usage doesn't do the `git checkout HEAD^2`

isobit (approved):

.github/workflows/codeql-analysis.yml
isobit: Approved because it should work fine either way, but I just noticed this was from Oct 2020 - I'm guessing it's just been removed from their example since then?

Lowercases (approved):

isobit (approved):

Issue #1156, opened 2020-09-03
Upcoming AWS change in ARN format breaks `emp ps`


tyrken:
AWS have long had an option to use a new format for ECS arns for services, tasks and container-instances, see https://aws.amazon.com/blogs/compute/migrating-your-amazon-ecs-deployment-to-the-new-arn-and-resource-id-format-2/ 
 What's new is that from October 2020 these options default to on for new accounts, and April 2021 they become mandatory for all accounts.  So from that latter date our existing empire controllers will see a mix of old & newer formats.

I've tested this by making a new account with the options turned on and it _mostly_ works.  The only fault I've found so far is that `emp ps` breaks, giving a "invalid ARN resource" error, likely from https://github.com/remind101/empire/blob/7ab6b866c6f4d884abf5900915c33db878c58aa2/pkg/arn/arn.go#L63-L77

Anybody else come across this?

PR #1165, opened 2020-10-22
Workaround new long ARNs

tyrken:
This patch should work around the new long ARNs that AWS are creating by default now (not next year as I had thought in https://github.com/remind101/empire/issues/1156).  It seems if you don't explicitly opt-out of the long ARNs you get them now, but you can opt out again manually.

We're actually using this server-side fix rebased to an older empire version, but it should also work in latest master I hope...

### DIFF
--- a/pkg/arn/arn.go
+++ b/pkg/arn/arn.go
@@ -63,7 +63,7 @@ func (a *ARN) String() string {
 // SplitResource splits the Resource section of an ARN into its type and id
 // components.
 func SplitResource(r string) (resource, id string, err error) {
-	p := strings.Split(r, "/")
+	p := strings.SplitN(r, "/", 2)
 
 	if len(p) != 2 {
 		err = ErrInvalidResource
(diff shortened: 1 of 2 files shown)


Lowercases (approved):
We've a similar patch on our ecs library, it works with the new arns.

oliviagunton:
Hello all (hi @aengelas, nice to see a familiar face here!)

Any chance we could get this approved and merged? AWS is removing the ability to opt out of the long ARN format [at the end of this month](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-account-settings.html#ecs-resource-arn-timeline).

aengelas:
Hey Olivia :-) Unfortunately, CircleCI won't build this because no one was following it it :-/ I fixed that; @tyrken any chance you can rebase and re-push this PR to trigger a new CircleCI build?

If we don't get a response, I can make a new a new PR with the changes.

isobit:
Looks good to me, but I just wanted to point out that this change means empire extracts `<cluster>/<name>` as service/task names from ARNs rather than just `<name>`. Luckily the ECS API accepts that for `serviceName` etc., but as far as I can tell that's not documented.

This isn't strictly necessary but it might also be worth updating the tests in [`cloudformation_tests.go`](https://github.com/remind101/empire/blob/master/server/cloudformation/cloudformation_test.go) to use the long ARN format instead, since that'll be the only format going forward.

PR #1168, opened 2021-03-01
Remove outdated MAINTAINER lines from Dockerfiles

aengelas:
Eric no longer works at Remind, and this field is deprecated. Given that
this repository is maintained by the remind101 org, I'm just removing
the line completely.


aengelas:
Tests failing because I opened this from my fork, and for some reason we require forks to auth with docker hub...

rex-remind101 (approved):
👍

